PR #15, opened 2019-02-17
TLS_Cipher_String_Cheat_Sheet: added markdown tables

valentinxxx:


valentinxxx:
Do you think is it possible to add `<br>` as an `allowed_elements` to markdownlint rule `MD033 - Inline HTML`?

righettod:
The need for usage of `<br>` in your context is legit so you can use this syntax to add `<br>` in your MD table:

```html
<!-- markdownlint-disable MD033 -->
Your MD table here
<!-- markdownlint-enable MD033 -->
```
It allow you to use HTML tags in your MD table, but please limit usage to `<br>` tag.

righettod:
You can also remove link to the image versions of the both tables if your contribution replace them.

valentinxxx:
I've commited the `markdownlint-disable` comments.
There is background colorization in the image versions, which we cannot reproduce in markdown.
Do you think is it OK to replace them?

righettod:
We can go in this way, for me it is not an issue to have the both tables
without the colorization. If needed, we will add legend indicators later.
I will merge the PR soon.
Thank you very much for your contribution.

PR #35, opened 2019-03-12
Fixed two security issues in disable_functions setting

crossjoin:
1) If 'show_source' is disabled, also 'highlight_file' must be disabled. 'show_source' is just an alias name of 'highlight_file'.
2) All disabled function names must be in one line. The previous split into multiple settings/lines resulted in only the last line was active (so all other functions were NOT disabled).


righettod:
Thanks you very much for the PR.

PR #49, opened 2019-03-21
CSRF CS Synchronizer Token Update

greyshell:
Additional consideration for synchronizer token where the CSRF token is added in a custom header for an extra layer of security by relying on the CORS technology to confirm the origin and the allowed headers.


cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -47,6 +47,8 @@ This defense is one the most popular and recommended methods to mitigate CSRF. I
 
 Any state changing operation requires a secure random token (e.g., CSRF token) to prevent CSRF attacks. A CSRF token should be unique per user session, large random value, and also generated by a cryptographically secure random number generator. The CSRF token is added as a hidden field for forms headers/parameters for AJAX calls, and within the URL if the state changing operation occurs via a GET. See "Disclosure of Token in URL" section below. The server rejects the requested action if the CSRF token fails validation.
 
+Inserting the CSRF token in the HTTP request header via JavaScript, is considered more secure than adding in the hidden field form parameter. In this situation, even if the CSRF token is weak, predictable or leaked but still an attacker cannot forge the POST request directly by setting the custom request header through XmlHttpRequest. This type of implementation gives an extra layer of defense.
```
righettod: *In this situation, even if the CSRF token is weak, predictable or leaked but still an attacker cannot forge the POST request directly by setting the custom request header through XmlHttpRequest*

An attacker having access to the JS context of the page (ex: XSS) can create a full custom HTTP POST request including [custom request headers](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/setRequestHeader)...

Do you refering to the case of an HTML form for which a submit event handler is attached in order to add the CSRF token before to send the form?

Perhaps i miss something but can you clarifiy you explanation please?

greyshell: @righettod 
> An attacker having access to the JS context of the page (ex: XSS) can create a full custom HTTP POST request including custom request headers...

Yes. As per my understanding, through XSS you can include the custom header in the HTTP POST Request. But before sending that request, browser will send the pre-flight request and in the response if Access-Control-Allow-Origin , Access-Control-Allow-Headers, Access-Control-Allow-Methods, Access-Control-Allow-Credentials are configured insecurely then only the CSRF attack will possible.

Please let me know your thoughts on this.

righettod: **Let use a case to clarifiy what you want to explain in your paragraph:**
The CSRF token is weak or predicable or leaked and the app expect the token to be passed via an dedicated HTTP header, We assume here that the attacker know the value of the token:
- He can use an iframe to post a form (GET/POST) but it will not be able to set the expected header.
- He can use a full JS well forged request but it require that the target app enable CORS in a unsafe way: *I have a doubt, does the browser let the request go and only block the access to the response OR the browser block the request directly?*

Usage of token in HTTP header prevent a risk of disclosure in URL or app/web server access log if this one is configure to log the body.

Does I have understand what you want to explain?

righettod: *I have a doubt, does the browser let the request go and only block the access to the response OR the browser block the request directly (we are outside of a CORS preflight flow here)?*

After some test using Chrome/Firefox (last version): The browser let the request flow to the target server but do not let the JS code access to the response content if there no CORS headers allowing that.

So, for a CSRF context (attacker want to cause a modification action on server even in blind mode), if an attacker have access to the CSRF token (weak, predictable, leaked), it can forge a valid request in JS that will be received by the target server and so the CSRF attack can be leveraged.

To resume, for me, using a HTTP header instead of a HTTP parameter (body/url) to carry the CSRF token, prevent only a risk of disclosure of the CSRF token in URL or app/web server access log if this one is configured to log the body.

greyshell: @righettod - Thanks joining with me on this discussion.

Let me simplify the topic.

**Scenario A:**
- request method is POST.
- CORS is not configured on the site
- there is no XSS hole in the site
- application passes the anti-csrf token in POST parameter.
- If the anti-csrf token is leaked and somehow attacker gets hold of the token, then can he perform the CSRF attack? 

My understanding is Yes. He can.


**Scenario B:**
- request method is POST.
- CORS is not configured on the site
- there is no XSS hole in the site
- application passes the anti-csrf token in request header. This is a custom header.
- If the anti-csrf token is leaked and somehow attacker gets hold of the token, then can he perform the CSRF attack? 

My understanding is No. He can't. 
Because when attacker sets custom header through XmlHttpRequest, browser will send the OPTIONS (pre-flight) request. 
With that argument, I would say, passing the anti-csrf token in custom header provides additional protection.

**My PoC**
Request 1:
```
GET /evil0wl/web-attacks/csrf/csrf-POST_add_custm_header.html HTTP/1.1
Host: 127.0.0.1
Cache-Control: max-age=0
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.82 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8
If-None-Match: "2d7-58567084d5ebc-gzip"
If-Modified-Since: Sun, 31 Mar 2019 17:14:10 GMT
Connection: close
```

Response 1:
```
HTTP/1.1 200 OK
Date: Sun, 31 Mar 2019 17:18:15 GMT
Server: Apache/2.4.29 (Debian)
Last-Modified: Sun, 31 Mar 2019 17:18:00 GMT
ETag: "2d4-5856715ff244d-gzip"
Accept-Ranges: bytes
Vary: Accept-Encoding
Content-Length: 724
Connection: close
Content-Type: text/html

<html>
<!-- This PoC script will send the request without any user interaction-->

<script type="text/javascript">
	
   // Vulnerable URL and POST parameters
   var url =  "http://demo.testfire.net/admin/admin.jsp";
   var params =  "username=jdoe&accttypes=Savings";
   
   var CSRF = new XMLHttpRequest();
   CSRF.open("POST", url, true);
   CSRF.withCredentials = 'true'; //IMPORTANT to exploit the active session of victim, for this browser will send cookie.
   
   // Provide valid request header
   CSRF.setRequestHeader("Content-type", "application/x-www-form-urlencoded");
   // set the custom header
   CSRF.setRequestHeader("Anti-CSRF", "1234");
   CSRF.send(params);
</script>

</html>
```

Request 2:
```
OPTIONS /admin/admin.jsp HTTP/1.1
Host: demo.testfire.net
Cache-Control: max-age=0
Access-Control-Request-Method: POST
Origin: http://127.0.0.1
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.82 Safari/537.36
Access-Control-Request-Headers: anti-csrf
Accept: */*
Referer: http://127.0.0.1/evil0wl/web-attacks/csrf/csrf-POST_add_custm_header.html
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8
Connection: close
```

Response 2:
```
HTTP/1.1 302 Found
Server: Apache-Coyote/1.1
Set-Cookie: JSESSIONID=CAFE46722F1A74276453BA9FF1CFBECF; Path=/; HttpOnly
Location: /login.jsp
Content-Length: 0
Date: Sun, 31 Mar 2019 17:18:15 GMT
Connection: close
```
csrf

Please let me know your thoughts on this. 

Also kindly educate me / nudge me to the right direction if the attack is still possible for **scenario 2**. 
I tried to set the header through flash as well and as excepted GET request was initiated for crossdomain.xml. 
csrf2

righettod:
Thanks you very much for the exchange. Indeed, adding via js a custom
header in a cross domain request will make fall the request in preflight
condition if i do made a mistake. So yes, we are agree :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
righettod: We are agree.
Can you just add this info between parenthesis in order to indicate to user the context of your explanation and allow the user to directly understand by header is preferred over parameter ?
Thanks you very much in advance 😃

ThunderSon: Thank you for this addition. Kindly add the following [reference to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests) to better clarify the notion of the preflighted requests.

greyshell:
@righettod Thanks you very much :)
I have added more info in my changes. Please review.

manideepkonakandla:
> Moreover, when the attacker tries to spoof the header through flash, the browser initiates the GET request for crossdomain.xml. In both cases, the browser prevents the forged request to be sent.

Hey, really good discussion going on here. Can you help me better understand this line? How does the GET request prevent the forged request to be sent? Can you once read the "use of custom headers" section in the cheatsheet that I wrote. I think the request would still go through (GETs are not pre-flighted). Also, I would not term it as "header spoofed through flash" since we are not doing any spoofing here but an redirection based attack. Does that make sense? Please let me know if you are referring to something else.

In a broad level discussion, in the scenario A you gave, you are assuming that "_CORS is not configured on the site_" but the default SOP would block the request in the first hand. Am I wrong? If CORS are enabled on the destination site for the site where you are trying to attack it from, parameter/header  does not matter as the attack will go through. I am not sure how it is providing additional protection? We are instead adding risk by tokens in server logs etc. as @righettod righettod mentioned (in web frameworks, it is very common practice to log tokens but I have not seen any frameworks doing body logging). I would love to know if I am missing something in this context.

One other assumption you made is token being leaked. Only ways token can be leaked if its stored on DOM is if
a) XSS - then you dont need to think about CSRF at all
b) HTTP connection - then you dont need to think about CSRF at all since cookies etc. are also stolen

Hence, I still think its better to add tokens in body parameters and not headers. There are definitely cases where people want to include in headers than body parameters (frameworks/libraries doing auto token inclusion etc. since its hard for them to identify the body and include payload as per the body structure, they prefer to do it in headers in which case they can just make sure that that particular header is not logged in their internal servers. I made this change recently in the cheat sheet which is merged)

Please let me know if there is any threat that I missed/mis-understood.

greyshell:
@manideepkonakandla 

Well. 🙂 let's discuss one by one in details.



> Moreover, when the attacker tries to spoof the header through flash, the browser initiates the GET request for crossdomain.xml. In both cases, the browser prevents the forged request to be sent.
> Hey, really good discussion going on here. Can you help me better understand this line? How does the GET request prevent the forged request to be sent? Can you once read the "use of custom headers" section in the cheatsheet that I wrote. I think the request would still go through (GETs are not pre-flighted).

I think here you entirely missed the flash context and the screenshot that I shared before. 
In that PoC, I was trying the set a custom header through flash and observing the behavior. 

If you notice the screenshot carefully then you will find there are total 3 requests were sent.

     request 80: GET request from the victim's browser to fetch the swf file. 
     request 81: POST request to the php redirector. 
     request 82: GET request for the crossdomain.xml


- After receiving the response for the request #80, the browser executes the swf file and that generates the POST request(#81).
- This behavior clearly indicates that I was not forging any GET request to perform the CSRF attack. I was trying to forge a POST request(line 37), with setting JSON content type header(line32) and Anti-CSRF header (line 34).
- After 307 redirections, browser finally initiated the GET request(#82) for `crossdomain.xml.`
- Summary - I did not get success to perform CSRF attack by setting the custom header in flash.

Please find the content of that `csrf.as` file

     package
     {
       import flash.display.Sprite;
       import flash.net.URLLoader;
       import flash.net.URLRequest;
       import flash.net.URLRequestHeader;
       import flash.net.URLRequestMethod;
     public class csrf extends Sprite
       {
         public function csrf()
         {
           super();
           var member1:Object=null;
           var myJson:String=null;
           member1=new Object(); 
         // compile command: mxmlc csrf.as -static-link-runtime-shared-libraries
           member1= 
         {
         "contractId":"7777",
             "usageMeterId": "18cbc422-9af1-4b0d-b5f9-8776e72b7784",
         "label": "blahh",
         "siteId":"7878"
    
           };
           var myData:Object=member1;
           myJson=JSON.stringify(myData);
           myJson=JSON.stringify(myData);
         // php file which has endpoint of the target
           var url:String="http://127.0.0.1/json-csrf.php";
           var request:URLRequest=new URLRequest(url);
         // setting the content type header to json
           request.requestHeaders.push(new URLRequestHeader("Content-Type","application/json"));
           // setting a customer header
         request.requestHeaders.push(new URLRequestHeader("Anti-CSRF","1234"));
    
           request.data=myJson;
           request.method=URLRequestMethod.POST;
           var urlLoader:URLLoader=new URLLoader();
     try
           {
               urlLoader.load(request);
               return;
           }
           catch(e:Error)
           {
               trace(e);
               return;
           }
         }
       }
     }



> In a broad level discussion, in the scenario A you gave, you are assuming that "CORS is not configured on the site" but the default SOP would block the request in the first hand. Am I wrong?

Yes, you are wrong.


> Scenario A:
  - request method is POST.
  - CORS is not configured on the site
  - there is no XSS hole in the site
  - application passes the anti-csrf token in POST parameter.
  - If the anti-csrf token is leaked and somehow attacker gets hold of the token, then can he perform the CSRF attack?

Here SOP would not block the request because I am just adding the form parameter through the XMLHttpRequest / by crafting a HTML file.

Now lets talk about Scenario B.


> Scenario B:
  - request method is POST.
  - CORS is not configured on the site
  - there is no XSS hole in the site
  - application passes the anti-csrf token in request header. This is a custom header.
  - If the anti-csrf token is leaked and somehow attacker gets hold of the token, then can he perform the CSRF attack?

My understanding → when any JavaScript is loaded from a different origin and tries the set any custom header through XMLHttpRequest, the browser will not simply block the request first hand, instead of that it sends the preflight (OPTIONS) request. 

Interesting headers in that request, (please refer the screenshot as well)

     Access-Control-Request-Headers: Anti-CSRF
     Origin: http://127.0.0.1
     Access-Control-Request-Method: POST

Based on the response from the server and with the `presence of specific headers and its values`, the browser determines whether to block/ send the forged request with cookie.


> If CORS is enabled on the destination site for the site where you are trying to attack it from, parameter/header does not matter as the attack will go through. I am not sure how it is providing additional protection?

Again I don’t agree with you.
My understanding → if `CORS is enabled on the destination site` then by default the attack will NOT go through. 

In order to make the attack successful below header values should be set at the minimum.

     Access-Control-Allow-Origin: *
     Access-Control-Allow-Headers: Anti-CSRF
     Access-Control-Allow-Methods: POST
     Access-Control-Allow-Credentials: true


Correct me if I missed any important header here.

Thus insecure_cors configuration is a pre-condition to perform the attack when the csrf token passes in the header. 

Does this pre-condition required for performing the attack when csrf token passes in the body parameter? No

With that argument, can't we say now, passing csrf token in header providing additional protection?


> We are instead adding risk by tokens in server logs etc. as @righettod righettod mentioned (in web frameworks, it is very common practice to log tokens but I have not seen any frameworks doing body logging). I would love to know if I am missing something in this context.
> One other assumption you made is token being leaked. Only ways token can be leaked if its stored on DOM is if
> a) XSS - then you dont need to think about CSRF at all
> b) HTTP connection - then you dont need to think about CSRF at all since cookies etc. are also stolen


I mentioned the `token leakage` term, in order to assess the `exploitability` factor. I was trying to mean if attacker knows the token somehow then what damage he can do with that stolen token.

There could be total three possible assumptions A, B, and C 
Assumption A → adding token in body parameter or header both does not make any difference
Assumption B→ adding a token in body parameter is better than adding in headers. 
Assumption C→ adding token in the header is better than adding in body parameter


> Hence, I still think its better to add tokens in body parameters and not headers.

Basically you are on Assumption B. I strongly disagree with this. This is most easiest way to perform CSRF attack when the token is weak / predictable(ie. 123) and transmits in body parameter.

Assumption A  will be valid ONLY if you are able to perform the CSRF attack when the application passes token in POST request header (custom header).
precondition → predictable token value (i.e 123), no XSS hole, no CORS is configured.

Finally if you can’t perform the attack then Assumption C will be valid. 

@righettod - Are you still agree with me?

@manideepkonakandla - If you know any attack technique that allows you to perform CSRF while passing the token value in header (adhering those pre-conditions) then please let me know and we can continue with this interesting discussion.

manideepkonakandla:
> n
Hey @greyshell  

Thank you so much for your elaborate answer and my sincere apologies for missing a fundamental point that form tags are not subject to SOP (I posted my comment when I was about to leave from office after 10 long hours of work but its still my mistake to miss that 👍 ). There is some terminology different usage between you and me when we say "CORS is enabled". Nevertheless, leaving all that

_a) If you know any attack technique that allows you to perform CSRF while passing the token value in header (adhering those pre-conditions) then please let me know and we can continue with this interesting discussion._
>> Yes, please refer to "[use of custom headers](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#use-of-custom-request-headers)" section in the cheatsheet where I wrote about this and referenced lot of articles around this. 

In simple words, what you are referring about flash/custom header etc. is nothing but this technique which itself is not secure.

So, 
1) using header for passing tokens
a. Yes, it adds additional protection by pre-flighting etc. but as mentioned above that technique is not fully secure and is exploitable (GET request you got for crossdomain.xml will be received only after the request is processed in the exploitation case)
b. We are assuming tokens are weak/stolen etc. (I dont think that assumption is a good one to make. I gave some reasons in my previous correspondence and also we have lot of recommendations where we say not to use weak tokens. Its like people using weak passwords)
c. Has an add-on risk of servers logging headers

2) using body parameter for passing tokens
a. Does not provide the protection offered by pre-flighting but again as mentioned that technique is not fully secure and is exploitable. 
b. We are assuming tokens are weak/stolen etc. (I dont think that assumption is a good one to make. I gave some reasons in my previous correspondence and also we have lot of recommendations where we say not to use weak tokens. Its like people using weak passwords)
c. Has no add-on risk of servers logging since I personally have not seen any servers logging request bodies on a continuous basis

Since b. is common for both, we need to make a decision basing on a) and c) to decide "If depending on a vulnerable technique to offer some protection (pass via headers)  is better than not having the risk of logging headers (via parameter bodies) which is common"?. I think its not.

Does this explanation convince you to prefer request bodies over headers? Again, as mentioned in my previous technique there are some places where you can accept the risk of logging headers and do it in headers but if you are doing it standalone for request, I prefer body parameters.

I hope I was able to convince but if you see that I missed any threat or made any wrong risk assessments, please let me know and I will be glad to learn them. Thanks again for your time :)

righettod:
Just a remark about this CORS configuration topology:
```
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: Anti-CSRF
Access-Control-Allow-Methods: POST
Access-Control-Allow-Credentials: true
```
You cannot allow all origins and allowing credentials, see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials). So, even if a web server is configured like this, the brower supporting CORS will raise an error.

About the initial topic of the PR: Your update provide for the expected missing explanation as well as documentation via the MDN link. 

For flash as it require the presence of the *crossdomain.xml* file on the victim server (same as enabling of CORS headers in COR context), it imply so a specific exploitation context in which the target web server has multiple configuration issues (CORS/crossdomain.xml) in addition of the CSRF issue that allow the exploitation of the CSRF issue. 

So, for me, the main topic of the PR about passing token in body/header in processed with proofs in the PR. Thanks for all this deep study.

I validate the PR.

@ThunderSon @mackowski Do you validate the PR?

manideepkonakandla:
> target web server has multiple configuration issues (CORS/crossdomain.xml)

Can you enlighten me on how does flash redirection based CSRF exploitation attack needs CORS misconfiguration? It works with/without CORS is my learning, please direct me to a resource contradicting it if I am wrong, I will be glad to learn.

Also, how does having flash is an misconfiguration? As per my learning, it is a feature not a mis- configuration which lot of people use. Again, please direct me to a resource contradicting it if I am wrong, I will be glad to learn.

in addition of the CSRF issue that allow the exploitation of the CSRF issue.
>> this is basic thing for entire discussion, how can it be "in addition" to the previous things mentioned by you?

I understand you and tagged members are core team members and I highly respect that, your opinions since you guys are doing way more work here than me but I think in a good conversation we also need to see opinion of the existing people who are already providing their view before making any final decision.

righettod:
There no relation between CORS and Flash, i have just said that in the explained context by @greyshell , to trigger the issue in CSRF you need that the victim server have to be configured with a weak CORS or a weak crossdomains.xml otherwise, like mentioned in the proof of @greyshell, the exploitation did not succeed.

Regarding the PR, the contribution of @greyshell is scoped (like every PR) and he have technically justified all the arguments on his contribution so there no reason to not merge the PR.

Regarding the core team, this team of 4 differents peoples with differents origin/countries/skills/mindset/XP pilot the overall project and decide to accept or not the PR/proposal, for the moment every user/contributor is happy. Here the PR intiator have done a great work and justified all the arguments with proofs for the scope of the PR so the team will finally decide (or not) to merge the PR and accept the contribution.

The core team decide to accept or reject a proposal/PR, it's how the project is managed. Some people can perhaps do not like this way but it's the choosen way to pilot the project so we deal wth it because for the moment our final users as well as OWASP management are happy on how the project is managed.

manideepkonakandla:
> trigger the issue in CSRF you need that the victim server have to be configured with a weak CORS or a weak crossdomains.xml otherwise

No, you are "absolutely" wrong in entire statement you made I clarified this 3 times so far in my above replies. @greyshell did the test on the patched browser and I think he already realized that since it looks like he got it cleared from a previous colleague of mine. Please double read the use of custom headers section and the articles linked in it (If you had read them clearly, you would have not made your previous statement).

You can go ahead and merge PR if you want.

manideepkonakandla:
In many browser versions this is still an issue, we cannot assume that all customers have latest version of browser and provide security leaving all other users with old versions of browsers vulnerable. There is a workaround linked in use of custom headers where it can be exploited even in latest browsers.

righettod:
@manideepkonakandla Can you provide PoC like @greyshell does in order that we see the exploitation scenario and the result on a panel of recent browsers (IE/EDGE/FF/Chrome) ?

manideepkonakandla:
> @manideepkonakandla Can you provide PoC like @greyshell does in order that we see the exploitation scenario and the result on a panel of recent browsers (IE/EDGE/FF/Chrome) ?

PoC is already provided in the section which I referenced above ("use of custom headers"). Anyway, I am providing it here for your easy reference [1]. They tested it with Firefox 50.0 with Shockwave Flash installed. You can try to replicate it. I dont think we need to show that its vulnerable on all browsers. As long as its shown on one, I think it is enough to eliminate that option.

[1] https://blog.appsecco.com/exploiting-csrf-on-json-endpoints-with-flash-and-redirects-681d4ad6b31b

On my request, [Appsecco](https://appsecco.com) team conducted an extensive work on it and provided me with results where they mentioned me all other vulnerable browser details etc. If you are interested in knowing all of them besides the one mentioned above, please drop me an email and I can send it to you.

manideepkonakandla:
and FWIW there is an entire email thread where I explained Dave Wichers, Jim Manico about this issue and both of them were convinced. You were on one of those email threads as well @righettod  :)

mackowski:
I will review all of this on Monday morning. If you can share all details (like old emails with detailed explanation) it will be awesome.
For sure we have to refactor this CS because as we can see in comments some details are not clear ;)
Lets do not rush with it, the aim is to create good dokument that everybody can depend on.

manideepkonakandla:
@mackowski  Thanks for making an open and healthy conversation. I will extract all of those emails and upload as PDFs/links to those PDFs sometime over this weekend.

greyshell:
First of all, I would like to thank all of you for providing your valuable insights and efforts regarding this discussion.

In the OWASP cheatsheet → Section: Use of Custom Request Headers

> Riyaz Walikar from Appsecco Labs was able to exploit this even in early 2019 using older versions of Firefox and also in some recent versions of Chrome (with a bit of user involvement).

 In the document, it is mentioned →  `recent version of Chrome`.
 
- I found the custom header issue was fixed long before. Reference → https://bugs.chromium.org/p/chromium/issues/detail?id=332023
- To verify that OWASP statement, I followed the same blog link → https://blog.appsecco.com/exploiting-csrf-on-json-endpoints-with-flash-and-redirects-681d4ad6b31b.
- In that PoC, the author only set the `Content-Type` header to `application/json` and that worked on my tested browser as well. But the interesting part to note, he did not actually add any custom header in the request.
- I tested on Chrome Version `52.0.2743.82` (64-bit), `flex sdk 4.6` and found the exploitation did not succeed when I tried to add a custom header like `Anti-CSRF`. 
- I have drafted my PoC (screenshots and code) for your reference 
https://paper.dropbox.com/doc/CSRF-Custom-Header--AayOGYyZAM63b2pRXhwdt_eeAg-5XQqI8q6PiC4dIQ0qdZNg

Now the **Question** is - did I miss (🤔 ) something to exploit it or the tested chrome `52.0.2743.82` / the latest chrome is not really vulnerable?

However, as per @manideepkonakandla comment, latest browsers are vulnerable.

>There is a workaround linked in use of custom headers where it can be exploited even in latest browsers.

I agree with @righettod's request. In order to resolve all confusion, it would be great if some one can take the initiative to perform a PoC to **confirm** if this attack is still possible on `recent` chrome / other browsers.

In case the attack is not possible then kindly update the `Use of Custom Request Headers` section to avoid reader’s confusion.

manideepkonakandla:
Sorry for the delay. I felt only [this conversation (https://github.com/OWASP/CheatSheetSeries/files/3052217/Riyaz.discussion.pdf) might be interest to you guys. All other email threads are where people agreed to what I said and all the references which I used there are currently linked in "use of custom headers" section. Linked document has details about which specific browsers they tested etc. You guys can feel free to test it on them and if you still find any issues, I will loop Riyaz here and we all can get questions clarified from him.

@greyshell You have a pretty interesting point about blog post having content-type header and not custom header. My reason to link them is because content-type triggers pre-flight (OPTIONS request) just like a custom header when it has a value other than "application/x-www-form-urlencoded , multipart/form-data, text/plain" (so application/json should trigger a pre-flight just like a custom header would do). In his article, you can see he uses the word custom header word a lot as well.

I really like how this discussion is going towards (parameter/body discussion to checking the efficiency of the custom headers technique). Thanks a lot everyone for contributing your knowledge here and having an healthy debate. If there is something wrong in the details I wrote, I will be more than happy to edit the content to avoid readers confusion just as @greyshell mentioned. At the end of the day, we all are here to provide accurate content to our readers. 

P.S: I have a very tight week ahead and might not be active over here but will try to look periodically. Let me know if you want me to include Riyaz/ you guys can reach out him directly as well. In the meanwhile, if you guys decided that my stance is wrong, you can wait for me to review your reasoning or else feel free to go ahead and merge this PR and do necessary changes under "use of custom headers" section. I am perfectly fine with either of them.

mackowski:
I will summarise what I learned from your comments/changes/referenced documents. 
Please correct me if I am wrong or if I misunderstood something :-) 

1. If we add strong and random CSRF token in the HTTP *custom* request header and if we do not log this header anywhere it is a little-bit stronger (because of CORS Preflighted requests) than the same CSRF token in body. - I think that this is clear.
2. But there is the danger of logging this header, because some frameworks tend to log all request headers.
3. We also know that custom headers itself, are not defence against CSRF (it can be for example bypassed by flash as @manideepkonakandla wrote, CORS was not designed to be defence against CSRF etc.).

I think that we should add this knowledge to this CS but I want to suggest a small change to what @greyshell proposed. 

IMO if we do not leak the token both methods are just secure. If you know that your infrastructure and application does not log/leak tokens using custom HTTP header seems like a good idea. But, if you are logging this headers, or you are not sure if they may be logged, I would recommend to put tokes in body. From @greyshell change developer can deduct that custom header with predicable value is good enough to protect from CFRS. I've asked young friend of mine how he understand this statement and he told me that "If attacker cannot forge the POST request with custom header, and this is safe even when token is stolen I can just add any header and no worry about the value there, it will make implementation simpler!"

My suggestion is to change it to something like this:
```In modern browsers inserting the CSRF token in the custom HTTP request header via JavaScript is considered more secure than adding the token in the hidden field form parameter. In this situation, even if the CSRF token is weak, predictable or leaked it is harder to exploit it. As per the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests) when the attacker tries to set any custom header through any XMLHttpRequest, the browser sends the OPTIONS (pre-flight) request. *Caution* you should not add CSRF token to headers (and use hidden field form parameter) if you or web framework of your choice log all the headers from requests because it can leak tokens. [Read here for more details](link to other section in CS where we summarise differences between tokens in custom HTTP request vs tokens in body)```

In addition, we can think how to convert all of good stuff, that we have here (like PoCs, samples, arguments from both sides) and allow others to get familiar with them. 
We can try to extract all of it from this discussion and add it to [use-of-custom-request-headers](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#use-of-custom-request-headers) section or create new section (e.g. Token in body vs Token in custom header).

greyshell:
@mackowski 
I am fine with the suggested change.

> We also know that custom headers itself, are not defence against CSRF (it can be for example bypassed by flash as @manideepkonakandla wrote, CORS was not designed to be defence against CSRF etc.).

I appreciate, if you can share your proof of concept result / screenshot like the way I did, to establish this fact that you can add custom header(for example  `Anti-CSRF-Token: 123`) through flash on recent chrome / other browsers.

manideepkonakandla:
@greyshell Did u get a chance to look at the document I shared where it explained that Riyaz already tested this and shared a PoC on his website? He also said that he tested it in a recent chrome version with the wrapper I linked. Since someone (I know Riyaz and his team from few days and they are pretty good at Appsec stuff) already tested it and provided the information, I believe it. If you believe that they might have provided some wrong information, please try to replicate it with the same browser versions etc. on your side and if it still does not work, then we can talk to Riyaz. If its decided that he must have missed something and what you found is accurate, you can definitely go ahead and change the recommendation. No issues in doing that at all. My full support will be to you then.

@mackowski If we are agreeing to [1], I am just thinking if we really need to do the additional blurb of tokens in headers?

_[1] We also know that custom headers itself, are not defence against CSRF (it can be for example bypassed by flash as @manideepkonakandla wrote, CORS was not designed to be defence against CSRF etc.)._

mackowski:
> > We also know that custom headers itself, are not defence against CSRF (it can be for example bypassed by flash as @manideepkonakandla wrote, CORS was not designed to be defence against CSRF etc.).
> 
> I appreciate, if you can share your proof of concept result / screenshot like the way I did, to establish this fact that you can add custom header(for example `Anti-CSRF-Token: 123`) through flash on recent chrome / other browsers.

To be clear I agree that custom headers with random token and with assurance that this token is not logged anywhere is better than token in body because they can provide additional protection without any other disadvantages.


My opinion is that custom headers itself - means without random token (or with predictable/leaked one) is not enough CSRF protection. Even if we do not consider flash crossdmain.xml issue. 
Why? Because it will only work in the simplest scenario where you have all applications served from one domain (e.g. one backend and frontend in the same domain) and you have robust CORS configuration. 
PoC that you provided are showing the simplest scenario where it is a sufficient protection. 
But this can be bypassed if CORS policy is badly implemented or configured or if your backend is used by more than one domain. 
In the last example (more than one application in different domains is using your backend) custom header can decrease attack surface but the vulnerability is still there. 
If you have API on api.example.com and two frontend applications admins.example.com and contractors.example.com via XSS on contractors.example.com you can attack admins with cookies from admins.example.com. The result is that XSS on any whitelisted website will allow to CSRF all other applications. And this is not the worst scenario because all of these 3 domains and applications are controlled by one company but it is also common that companies use third-party applications issued on domains allowed by CORS, in that scenario XSS in any of them causes complete circumvention of this protection in your application.
And this setup is pretty common nowadays, in fact in microservices world, in big companies, architecture can be more complex. 


Summarising I still think that we can improve this CS by adding a slightly changed version of what @greyshell suggested and by adding additional section where we can put all advantages and disadvantages of both methods that we discussed here.

manideepkonakandla:
> Even if we do not consider flash crossdmain.xml issue.

In the example you gave, even if there is XSS on atleast one subdomain and CORS is weak, we are vulnerable even if we use a random token in custom header. Since this XSS can request the page, grab token and then make a forged request but its a totally different problem of having XSS. I have slightly different reasoning on why your proposed solution is definitely a good idea.

> To be clear I agree that custom headers with random token and with assurance that this token is not logged anywhere is better than token in body because they can provide additional protection without any other disadvantages.
> 
> My opinion is that custom headers itself - means without random token (or with predictable/leaked one) is not enough CSRF protection.

Completely agreed!! We can say that including tokens in headers/ use of custom headers is fine as long as a random unique token/user specific token etc. (we spoke about how to generate tokens at length in our current guide) is used instead of standard one (ex: XMLHTTPRequest) or weak tokens AND headers are not being logged on servers. People who cant adhere to these can use parameter bodies. This works because
a) If attacker visits a site which is trying to exploit this flash based attack. Even though attacker script can make a request bypassing CORS/crossdomain.xml thing, it will not have a valid token (which we store on our original website DOM and attacker script has no access to it)
b) If attacker stores swf on the original website itself, we shouldn't be worried about CSRF anymore (if attacker can inject a malicious SWF file to a website, he can probably do lot of other malicious stuff).

I think we can just do these changes

@greyshell @mackowski Any concerns here in the explanation I gave? If not, let me know if one of you wanted to do these changes and if none of you want to volunteer, I can make the changes next week. Totally fine with either. We need to update "use of custom request headers section" as well. Again, thanks a lot guys - this really indeed is a interesting discussion and I got a chance to refresh my knowledge on this custom headers technique!!

mackowski:
Maybe I wrongly explained my example with api.example.com, admins.example.com and contractors.example.com. 
CORS is set that api accept requests from admins and contractors. admins do not allow requests from contractors or vice versa. In that scenario even when you have XSS on contractors you cannot just request the admin page and grab token because such cross-domain request will be blocked. 
CORS is often configured on domain and not on endpoint level and XSS on one page can be used to bypass CSRF protection based on custom headers without tokens.

mackowski:
I tried to exploit custom header via flash/SWF files and I was only able to change ContentType to application/json but I was unable to craft request with custom header.  
Appseco research https://blog.appsecco.com/exploiting-csrf-on-json-endpoints-with-flash-and-redirects-681d4ad6b31b also is about ContentType header any not any header.
The newest PoC that I was able to find https://github.com/sp1d3r/swf_json_csrf directly says that this is impossible: 

> Does it possible to craft custom headers like X-Req uested-With, Origin or Referrer?
> Answer: no (it was possible in the past, but not now

So from what I learned it is possible to craft ContentType header but not custom Anti-CSRF header as @greyshell also confirmed before.

mackowski:
My opinion is still that @greyshell PR is valid but we have to add waring against logging headers and I will still change 

> “In this situation, even if the CSRF token is weak, predictable or leaked but still an attacker cannot forge the POST request directly by setting the custom request header through XmlHttpRequest”

To something what I suggested before like: 

> “In this situation, even if the CSRF token is weak, predictable or leaked it is harder to exploit it. “ because of CORS and old browsers.

And add additional section with comparison between token in body and token in header

@manideepkonakandla can you invite appseco to review this?
@greyshell, @righettod and @ThunderSon  what is your opinion?

greyshell:
>So from what I learned it is possible to craft ContentType header but not custom Anti-CSRF header as @greyshell also confirmed before.

Thanks a lot @mackowski for confirming that . I would also like to request you to test on old chrome that I tested `52.0.2743.82`and released around `2016-08-3`. 

My understanding is this issue(adding custom header) was fixed long before in chrome.
https://bugs.chromium.org/p/chromium/issues/detail?id=332023

Yes. I agreed with you adding warning against logging headers is good idea.

manideepkonakandla:
Ya, I think he meant to say recent browsers.

> "Inserting the CSRF token in the HTTP request header via JavaScript, is considered more secure than adding in the hidden field form parameter/AJAX parameter as long as the token is random enough and is not weak or predictable. This type of implementation gives an extra layer of defense and is also not vulnerable to Flash based CSRF bypasses that were found earlier. When you transfer the token this way, you need to make sure that this header is not logged on your server side"

Does this look to both of you? This is inline to what we were discussing above.

>vulnerable to Flash based CSRF bypasses that were found earlier
This line I added above will cover if there are any old browsers vulnerable to it and also using random tokens (unlike weak/predictable one) in headers provides additional defense (As me and @mackowski were trying to explain above).

manideepkonakandla:
I am in a conference entire week. If you guys still want me to include Appsecco, I can do it early next (I can also try doing PoC of these next week to see if they apply only for content type pre-flights and not the custom header pre-flights)

greyshell:
@manideepkonakandla Thanks. I agree with that change. It makes more clear and to the point.

ThunderSon:
Hello everyone. Since everyone is at the same page and up to level with the technology, I'll make it quick.
CORS was created to be a `defense in depth` concept. It gives the browser extra powers to protect the user. If the application is weak, it's vulnerable in all cases. The discussion took a very bad detour from what @greyshell was trying to do. 
The PR is valid as adding the custom header will create an `extra layer` of protection, and that layer can be bypassed if the `user` allows it or even works to bypass it (request can be proxied).
And a point to add in here, if some web frameworks are storing the headers, they are logging the cookie/authorization headers as well. If the CSRF's life span is longer than the cookie/authorization, then there is a bigger issue to discuss in here. I have no idea how CSRF became more relevant than the login token itself. And if a user is using a browser that is 2 years old, it's not the issue of the CS, it's the user's.
We'll wait for @righettod 's final approval to merge the close and merge the PR. Thank you all for this discussion and contribution.

righettod:
As @ThunderSon and @mackowski have managed globally the PR I let them merge the PR.
I totally share their remarks and opinion on this topic 😃

mackowski:
I will marge this PR. If someone wants to add new section with comparison between tokens in body and header or update FLASH based re-direct attacks content it is more than welcome.

Issue #76, opened 2019-04-16
HMAC based token pattern scheme


69slimm:
Excusme, I wanna ask you about this :

> https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md

i want to implement this but i still got no idea about the flow work HMAC based token, pointly on verify the token using HMAC. can you explain more detail sir ? Any help from you will be useful for me in the future. Thank you

righettod:
Hi,
We will try to come back to you ASAP in order to help you.

69slimm:
thanks for the responds, i will waiting.

manideepkonakandla:
Hey @69slimm 

I wrote that HMAC based token method few months back but on re-reading it now, I can see why my explanation was not clear. I will make the explanation more clear on the cheat sheet by end of this week (Basically your token will be a combination of HMAC and couple of other identifiers). I am also going to make it more simplified by removing the operation/data/payload field as well.  

Thank you so much for raising this issue to us!

righettod:
@manideepkonakandla Can I assign this action to you in order that you help @69slimm on this topic ? We can assist you in case of need.
Thanks you very much in advance.

69slimm:
@manideepkonakandla ok sir. im gonna waiting. Just text me again here sir.

manideepkonakandla:
@righettod  Yes sure, you can assign it to me.

righettod:
GitHub do not let me assign it to @manideepkonakandla because i cannot find him in the assignee list, it's strange, anyone have already meet this case ?
@ThunderSon @mackowski @jmanico

jmanico:
https://github.com/manideepkonakandla looks real to me...

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

righettod:
Do not work unfortunately...

![NotFound](https://user-images.githubusercontent.com/1573775/56292035-ac5df100-6126-11e9-9df5-758b33a93267.png)

manideepkonakandla:
Not sure why. If you were not able to figure it out, dont worry! I have it on my task list and will complete it over the weekend.

righettod:
Thanks you, we will investigate in parallel to fix the issue...

69slimm:
@manideepkonakandla @righettod 
hey guys what about this scenario : 
[...]

PR #78, opened 2019-04-22
Multiple changes to Encryption and HMAC based token protections

manideepkonakandla:
I made below changes, please let me know if you guys see anything wrong in this. Will be glad to discuss :)

a) Though we mentioned in earlier versions that these techniques can be used if we don't want to maintain state at the server, our description of techniques were contrary (we ask them to include nonce, data/payload field etc. for which state needs to be maintained at the server).  I removed such fields as they are not needed. My threat analysis made me feel that we can get good security even without them.
b) Earlier encryption based protection asks to use userID to generate token and when received back, it says to match userID to sessionID. I felt its actually not needed and token can be generated using sessionID itself, so made the changes accordingly.
c) Few other cosmetic changes are made to make the implementation easy to read for our users

These changes triggered from https://github.com/OWASP/CheatSheetSeries/issues/76

Thank you for submitting a Pull Request to the Cheat Sheet Series. 

If your PR is related to an issue. Please end your PR text with the following line:

```
```

Thanks you again for your contribution :smiley:

### DIFF
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -93,9 +93,7 @@ For these cases, attempting to retrofit this pattern in existing applications re
 
 The Encrypted Token Pattern leverages an encryption, rather than comparison method of Token-validation. It is most suitable for applications that do not want to maintain any state at server side. 
 
-Server generates a token comprised of the user's session ID and a timestamp (to prevent replay attacks) using a unique key available only on the server (AES256 or stronger encryption algorithms are recommended). This token is returned to the client and embedded in a hidden field for forms, in the request-header/parameter for AJAX requests. On receipt of this request, the server reads and decrypts the token value with the same key used to create the token. Inability to correctly decrypt suggest an intrusion attempt (recommend to block and log the attack for incident response purposes). Once decrypted, the users sessionId and timestamp contained within the token are validated; session-id is compared against the currently logged in user, and the timestamp is compared against the current time to verify that its not beyond the defined token expiry time. If session-id matches and the timestamp is under the defined token expiry time, request can be allowed.
-
-If you would like to employ key rotations (recommended), you can include key related information (number representing rotation or any other identifier basing on your organizational/individual key management systems being used) in the token generation process and use that identifier while validating back at the server.
+Server generates a token comprised of the user's session ID and a timestamp (to prevent replay attacks) using a unique key available only on the server (AES256-with GCM mode or stronger encryption algorithms are recommended). This token is returned to the client and embedded in a hidden field for forms, in the request-header/parameter for AJAX requests. On receipt of this request, the server reads and decrypts the token value with the same key used to create the token. Inability to correctly decrypt suggest an intrusion attempt (recommend to block and log the attack for incident response purposes). Once decrypted, the users sessionId and timestamp contained within the token are validated; session-id is compared against the currently logged in user, and the timestamp is compared against the current time to verify that its not beyond the defined token expiry time. If session-id matches and the timestamp is under the defined token expiry time, request can be allowed.
 
 ### HMAC Based Token Pattern
 
@@ -109,9 +107,7 @@ a) Generating the token
 b) Include the token (i.e., HMAC+timestamp) in a hidden field for forms and in the request-header/parameter for AJAX requests 
 c) Validating the token 
    When the request is received at the server, re-generate the token with same key K (parameters are sessionID from the request and timestamp in the received token). If the HMAC in the received token and the one generated in this step match, verify if timestamp received is less than defined token expiry time. If both of them are success, then request is treated as legitimate and can be allowed. If not, we block the request and log the attack for incident response purposes.
-   
-Key rotation concept explained in encryption based token protection applies to this as well.
-  
+     
 ## Auto CSRF Mitigation Techniques
 
 Though the technique of mitigating tokens is widely used (stateful with synchronizer token and stateless with encrypted/HMAC token), the major problem associated with these techniques is the human tendency to forget things at times. If a developer forgets to add the token to any state changing operation, they are making the application vulnerable to CSRF. To avoid this, you can try to automate the process of adding tokens to CSRF vulnerable resources (mentioned earlier in this document). You can achieve this by doing the following:
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
 
 ### Encryption based Token Pattern
 
-The Encrypted Token Pattern leverages an encryption, rather than comparison method of Token-validation. It is most suitable for applications that do not want to maintain any state at server side.
+The Encrypted Token Pattern leverages an encryption, rather than comparison method of Token-validation. It is most suitable for applications that do not want to maintain any state at server side. 
 
-After successful authentication, the server generates a unique token comprised of the user's ID, a timestamp value and a [nonce](http://en.wikipedia.org/wiki/Cryptographic_nonce), using a unique key available only on the server. This token is returned to the client and embedded in a hidden field for forms, in the request-header/parameter for AJAX requests. On receipt of this request, the server reads and decrypts the token value with the same key used to create the token. Inability to correctly decrypt suggest an intrusion attempt. Once decrypted, the UserId and timestamp contained within the token are validated; the UserId is compared against the currently logged in user, and the timestamp is compared against the current time.
+Server generates a token comprised of the user's session ID and a timestamp (to prevent replay attacks) using a unique key available only on the server (AES256 or stronger encryption algorithms are recommended). This token is returned to the client and embedded in a hidden field for forms, in the request-header/parameter for AJAX requests. On receipt of this request, the server reads and decrypts the token value with the same key used to create the token. Inability to correctly decrypt suggest an intrusion attempt (recommend to block and log the attack for incident response purposes). Once decrypted, the users sessionId and timestamp contained within the token are validated; session-id is compared against the currently logged in user, and the timestamp is compared against the current time to verify that its not beyond the defined token expiry time. If session-id matches and the timestamp is under the defined token expiry time, request can be allowed.
 
-On successful token-decryption, the server has access to parsed values, ideally in the form of [claims](http://en.wikipedia.org/wiki/Claims-based_identity). These claims are processed by comparing the UserId claim to any potentially stored UserId (in a Cookie or Session variable, if the site already contains a means of authentication). The Timestamp is validated against the current time, preventing replay attacks. Alternatively, in the case of a CSRF attack, the server will be unable to decrypt the poisoned token, and can block and log the attack.
-
-This technique addresses some of the shortfalls in other stateless approaches, such as the need to store data in a Cookie, circumnavigating the Cookie-subdomain and HTTPONLY issues.
+If you would like to employ key rotations (recommended), you can include key related information (number representing rotation or any other identifier basing on your organizational/individual key management systems being used) in the token generation process and use that identifier while validating back at the server.
 
 ### HMAC Based Token Pattern
 
-[HMAC (hash-based message authentication code)](https://en.wikipedia.org/wiki/HMAC) is a cryptographic function that helps to guarantee integrity and authentication of a message. It is another way that CSRF mitigation can be achieved without maintaining any state at the server and is similar to an encryption token-based pattern with two main differences:
+This mitigation is also achieved without maintaining any state at the server. [HMAC](https://en.wikipedia.org/wiki/HMAC) based CSRF protection works similar to encryption based CSRF protection with couple of minor differences
+1. We use strong HMAC function (SHA256 or stronger algorithms are recommended) instead of an encryption function to generate the token
+2. Token is HMAC+timestamp
 
-- Uses a strong HMAC function instead of an encryption function to generate the token
-- Includes an additional field called ‘operation’ that would indicate the purpose of the operation for which you are including the CSRF token (may it be form tag/ajax call)
-
-Example:
-
-```text
-oneclickpurchase (or) buy/asin=SDFH&category=2&quantity=3)
-```
-
-**Note:** Fields mentioned in encryption token pattern (user's ID, a timestamp value and a nonce) are included.
-
-The operation field helps in mitigating the fact that the hash function generates the same value irrespective of multiple iterations (unlike strong encryption functions that generate different values when they are encrypted each time). So, it would help in avoiding having repeated token values across your application. Nonce field serves the same purpose as in encrypted token pattern (i.e., to avoid rare collisions due to weak cryptographic functions and acts as a defense-in-depth measure).
-
-Generate the token using HMAC including all four fields mentioned previously (user's ID, a timestamp value, nonce, and operation) and then include it in hidden fields for form tags, headers/parameters for ajax calls. Once you receive the HMAC from the client in the requests, re-generate HMAC with the same fields that you used to generate it, and then verify that the HMAC you re-generated matches the HMAC received from the client. If it does, it is a legitimate user request and if it does not, flag it as a CSRF intrusion and alert your incident response teams. Because an attacker has no visibility into the key used for generating the hash fields used in generating it, there is no way for them to re-generate it to use in forged request.
-
+Below steps explain implementing the HMAC based CSRF protection
+a) Generating the token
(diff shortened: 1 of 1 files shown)


cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+Server generates a token comprised of the user's session ID and a timestamp (to prevent replay attacks) using a unique key available only on the server (AES256 or stronger encryption algorithms are recommended). This token is returned to the client and embedded in a hidden field for forms, in the request-header/parameter for AJAX requests. On receipt of this request, the server reads and decrypts the token value with the same key used to create the token. Inability to correctly decrypt suggest an intrusion attempt (recommend to block and log the attack for incident response purposes). Once decrypted, the users sessionId and timestamp contained within the token are validated; session-id is compared against the currently logged in user, and the timestamp is compared against the current time to verify that its not beyond the defined token expiry time. If session-id matches and the timestamp is under the defined token expiry time, request can be allowed.
```
righettod: About encryption using AES algorithm, it can be interesting to warn the reader to use a [mode of operation](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation) as well as an implementation that do not expose the ciphered token to [padding oracle attack](https://www.owasp.org/index.php/Testing_for_Padding_Oracle_(OTG-CRYPST-002)) and propose the reader to use [AES with GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) or [AES with SIV](https://github.com/miscreant/miscreant/wiki/AES-SIV)  for example.

What do you think?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
 
-On successful token-decryption, the server has access to parsed values, ideally in the form of [claims](http://en.wikipedia.org/wiki/Claims-based_identity). These claims are processed by comparing the UserId claim to any potentially stored UserId (in a Cookie or Session variable, if the site already contains a means of authentication). The Timestamp is validated against the current time, preventing replay attacks. Alternatively, in the case of a CSRF attack, the server will be unable to decrypt the poisoned token, and can block and log the attack.
-
-This technique addresses some of the shortfalls in other stateless approaches, such as the need to store data in a Cookie, circumnavigating the Cookie-subdomain and HTTPONLY issues.
+If you would like to employ key rotations (recommended), you can include key related information (number representing rotation or any other identifier basing on your organizational/individual key management systems being used) in the token generation process and use that identifier while validating back at the server.
```
righettod: You can propose a example of representation of the key related information (or a link to an example) to guide the reader to apply your proposal.

What do you think?

righettod:
Thanks you very much for your PR,

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Hey, 

Yes, its a valid point. I thought of keeping it in the beginning but then I thought it might look more of cryptography document and did not keep it. I will keep it now.  

Also, I was trying to see if I can link any OWASP cryptography document and then found [this](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cryptographic_Storage_Cheat_Sheet.md) and [this](https://www.owasp.org/index.php/Testing_for_Weak_Encryption_(OTG-CRYPST-004)) where we still recommend AES 128. Maybe we should update them recommending to 256 bits?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Yeah, I can now see that the current version is not easy to understand. 

For now, I will re-submit without any information about key rotations (as it was before this PR) and when I get some free time over next week, I will submit a new PR with clear example about key rotation.

manideepkonakandla:
@righettod Basing on your comments, I made the changes in this same branch. I also responded to each of your comments. Thank you so much for your time :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
righettod: *I am not a crypto-analyst so perhaps the thing i said below are totally wrong:*

I have seen that there no signifiant security strengh increase between AES 256 vs 128 bits when the target protected data is not a critical asset. I think here that the important point is the fact that the anti-CSRF token is ciphered but we can accept the risk that it can be theoretically "cracked".

https://www.keylength.com/en/4/

Keylength-NIST-Reco.

As you said references to the *Cryptographic Storage Cheat Sheet* can be useful and we can, in this section of the CSRF CS, just add a simple note about the importance of using GCM or SIV mode of operation and not CBC based mode of operation.

What do you think?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
righettod: A small example is enough, it's just to guide a developer to use your recommandation "from scrach" with a simple representation of the key related information in the token. 

It's a kind of getting-started hint...

righettod:
You are welcome 😃

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,23 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
-
+This mitigation is also achieved without maintaining any state at the server. [HMAC](https://en.wikipedia.org/wiki/HMAC) based CSRF protection works similar to encryption based CSRF protection with couple of minor differences
+1. We use strong HMAC function (SHA256 or stronger algorithms are recommended) instead of an encryption function to generate the token
+2. Token is HMAC+timestamp
+
+Below steps explain implementing the HMAC based CSRF protection
```
ThunderSon: `Below are the steps for the proper implementation of the HMAC based CSRF protection:`
How does this sound to you?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,23 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+a) Generating the token
```
ThunderSon: `Generate`. Since we are giving steps, to use the present tense

ThunderSon: Following markdown best practices, `a)` doesn't exist for creating lists. Would it be possible to use numbered lists, as specified in the MD best practices?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,23 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+   Using key K, generate HMAC(usersessionID+timestamp) and append the same timestamp value to it which results in your CSRF token.
+b) Include the token (i.e., HMAC+timestamp) in a hidden field for forms and in the request-header/parameter for AJAX requests
```
ThunderSon: `_i.e_ HMAC+timestamp`. And it's only the request-header. What parameter? They are all headers. Unless you mean header fields?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,23 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
-
+This mitigation is also achieved without maintaining any state at the server. [HMAC](https://en.wikipedia.org/wiki/HMAC) based CSRF protection works similar to encryption based CSRF protection with couple of minor differences
+1. We use strong HMAC function (SHA256 or stronger algorithms are recommended) instead of an encryption function to generate the token
+2. Token is HMAC+timestamp
+
+Below steps explain implementing the HMAC based CSRF protection
+a) Generating the token
+   Using key K, generate HMAC(usersessionID+timestamp) and append the same timestamp value to it which results in your CSRF token.
```
ThunderSon: Might you know any reference/tutorial/blog that discusses implementing this? It'll be a great addition for developers to read up on actual implementation of this.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,23 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+b) Include the token (i.e., HMAC+timestamp) in a hidden field for forms and in the request-header/parameter for AJAX requests 
+c) Validating the token 
+   When the request is received at the server, re-generate the token with same key K (parameters are sessionID from the request and timestamp in the received token). If the HMAC in the received token and the one generated in this step match, verify if timestamp received is less than defined token expiry time. If both of them are success, then request is treated as legitimate and can be allowed. If not, we block the request and log the attack for incident response purposes.
```
ThunderSon: `we` usage. Rephrase if possible.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Yes, I agree with what you quoted for 128 but just as a defense in depth/safe side its better to recommend 256. 128 is OK too.

I think CBC is safe to use as long as few conditions are met. ECB is strictly not recommended. Made changes accordingly.

manideepkonakandla:
Addressed comments. Thanks both for your feedback :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Updated you already on this :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Hey, can u sync up with @righettod on this?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: If you agree, you can resolve this conversation.

righettod: My following word has to be confirmed with @ThunderSon and @mackowski and @jmanico

I think that it's preferable to propose to use `AES 128 GCM/SIV` over `AES 256 CBC` because:
* A strengh of 128 bits seems to be enough here according to the target value protected.
* CBC seems to tend to be more and more deprecated in flavor of GCM/SIV: [link](https://security.stackexchange.com/a/150749) and [link](https://security.googleblog.com/2013/11/a-roster-of-tls-cipher-suites-weaknesses.html)

I'm not blocked on the topic, I just try to find to right balance and advice for a developer implementing the crypto on the CSRF token and not aware about CBC vs GCM vs SIV usage context and caveats...

righettod: I have read your update on this part and it's OK for me.
Thanks for the update.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
righettod: Perhaps i will be wrong but according to the kind of value protected here (CSRF token) we can skip the key rotation topic because it can be overkill according to the crypto usage here. 

It can be interesting to add a link to [this section](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Key_Management_Cheat_Sheet.md#key-management-lifecycle-best-practices) of the Key management CS in order to provide hint about key storage/lifecyle.

What do you think?

ThunderSon: As this is discussing keys, it doesn't hurt to let the reader know about the key management CS.

mackowski: I like the idea of cross-reference to Key management CS

manideepkonakandla:
Hey, I resolved all the comments. Let me know any if any further issues need to be cleared else we can merge it :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
-
+HMAC Based Token Pattern mitigation is also achieved without maintaining any state at the server. [HMAC](https://en.wikipedia.org/wiki/HMAC) based CSRF protection works similar to encryption based CSRF protection with a couple of minor differences
+1. Use strong HMAC function (SHA256 or stronger algorithms are recommended) instead of an encryption function to generate the token
+2. Token is HMAC+timestamp
+
+Below are the steps for the proper implementation of the HMAC based CSRF protection:
+1. Generate the token
+   Using key K, generate HMAC(usersessionID+timestamp) and append the same timestamp value to it which results in your CSRF token.
```
ThunderSon: I don't think that this is behaving as you wish to have it. You meant to have it as a subpoint? Or what exactly? One new line in markdown keeps the text on the same line in the rich text.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+2. Include the token (i.e., HMAC+timestamp) in a hidden field for forms and in the request-header field/request body parameter for AJAX requests
```
ThunderSon: `*i.e.*` instead of `i.e.,`

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -91,33 +91,27 @@ For these cases, attempting to retrofit this pattern in existing applications re
...
+1. Use strong HMAC function (SHA256 or stronger algorithms are recommended) instead of an encryption function to generate the token
+2. Token is HMAC+timestamp
+
+Below are the steps for the proper implementation of the HMAC based CSRF protection:
+1. Generate the token
+   Using key K, generate HMAC(usersessionID+timestamp) and append the same timestamp value to it which results in your CSRF token.
+2. Include the token (i.e., HMAC+timestamp) in a hidden field for forms and in the request-header field/request body parameter for AJAX requests 
+3. Validating the token 
+   When the request is received at the server, re-generate the token with same key K (parameters are sessionID from the request and timestamp in the received token). If the HMAC in the received token and the one generated in this step match, verify if timestamp received is less than defined token expiry time. If both of them are success, then request is treated as legitimate and can be allowed. If not, block the request and log the attack for incident response purposes.
```
ThunderSon: Same about here about the newline comment.
`are successful, then the request ...`

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Not an english expert but 

https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/
https://en.wiktionary.org/wiki/i.e.

ThunderSon: Oh, haha. I am not talking about the meaning of it. I am just saying how it should be put in the sentence. It should be italicized and without the comma 😄

ThunderSon:
Thank you @manideepkonakandla for this update!

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: I got that. Links I shared seems to say that "," is not needed besides the i.e.

Anyway, I am not too much worried about these punctuation's ;) so I will change them as you requested.

righettod:
@ThunderSon @mackowski Can I merge it?

Issue #100, opened 2019-05-08
Cheat sheet update/refactor proposal: Password_Storage_Cheat_Sheet


fabienleite:
Please provides the following information about your proposal:

- What update do you want to perform? 

**Correct typo or number formatting on line 65.**

The problem is that the way it is now is confusing (at least to me), we don't know if the number meant is "one billion" or "a hundred thousand" :

```In the example above, as PBKDF2 [...]  at least 1 second (like 1000.000 for example).```

Should probably become :

```In the example above, as PBKDF2 [...]  at least 1 second (like 100.000 for example).```

Or : 

```In the example above, as PBKDF2 [...]  at least 1 second (like 1.000.000 for example).```

Based on popular implementations, I think 100.000 was meant but I'm unsure.  
I'll submit a PR if you validate the change and provide which one of the two is the correct number.

righettod:
Thank you very much for your issue.

**PBKDF2** is not time constant, it depend on the target execution machine.

For example, this code:

```java
import javax.crypto.SecretKeyFactory;
import javax.crypto.spec.PBEKeySpec;
import java.security.SecureRandom;

/**
 * Sandbox for code try
 */
public class Sandbox {

    public static void main(String[] args) throws Exception {
        //Initialization
        SecureRandom rnd = new SecureRandom();
        byte[] salt = new byte[64];
        SecretKeyFactory skf = SecretKeyFactory.getInstance("PBKDF2WithHmacSHA512");
        char[] password = "mypassword".toCharArray();
        //Test for 10.000 iterations
        rnd.nextBytes(salt);
        PBEKeySpec spec = new PBEKeySpec(password, salt, 10000, 256);
        long start = System.currentTimeMillis();
        skf.generateSecret(spec);
        System.out.printf("Computation time is %s milliseconds for 10.000 iterations with a key size of 256 bits\n", (System.currentTimeMillis() - start));
        //Test for 100.000 iterations
        rnd.nextBytes(salt);
        spec = new PBEKeySpec(password, salt, 100000, 256);
        start = System.currentTimeMillis();
        skf.generateSecret(spec);
        System.out.printf("Computation time is %s milliseconds for 100.000 iterations with a key size of 256 bits\n", (System.currentTimeMillis() - start));
        //Test for 500.000 iterations
        rnd.nextBytes(salt);
        spec = new PBEKeySpec(password, salt, 500000, 256);
        start = System.currentTimeMillis();
        skf.generateSecret(spec);
[...]

PR #102, opened 2019-05-09
Update Password_Storage_Cheat_Sheet: correct iteration_count number typo

fabienleite:
Hello, as stated in #100, here is my PR for review.

This PR covers Issue #100

### DIFF
--- a/assets/Password_Storage_Cheat_Sheet_Test_PBKDF2_Iterations.java
+++ b/assets/Password_Storage_Cheat_Sheet_Test_PBKDF2_Iterations.java
@@ -1,0 +1,39 @@
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import java.security.SecureRandom;
+
+// PLEASE RENAME THIS FILE TO PBKDF2ItEval.java BEFORE COMPILING.
+public class PBKDF2ItEval {
+
+    public static void main(String[] args) throws Exception {
+        //Initialization
+        SecureRandom rnd = new SecureRandom();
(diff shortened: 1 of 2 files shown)


righettod:
Thank you very much for the PR. We will process it ASAP.

mackowski:
I am not sure if this statement is clear to someone who is not familiar with slow-hashing algorithms.

What does it means `computation time on the target system must take at least 1 second`. What is target system and why 1 second? If target system is attacker system it make sense to make it about 1 second but developer does not have capabilities to check it. On the other hand if target system is your application server I do not fully agree with this statement.

righettod:
It is the main issue with pbkdf2, as it is not time computation constant,
you cannot ensure that the setting to take 1 sec on your server will take
also 1 sec on an attacker machine. Here, the goal is to made aware of that
point, people that decide to use pbkdf2. Usage of Argon2 is highly
recommanded.

mackowski:
But in current form this is not clear for me. For example I do not understand what is `target system` in this case. Maybe I should read it one more time...
Anyway maybe if the goal is to made aware that pbkdf2 is not time constant, instead of telling that you should `must have a number implying that the computation time on the target system must take at least 1 second.` or `test the appropriate number for your specific situation` (what is impossible) we should just write that `pbkdf2 is not time constant and therefore is hard to asses number of iterations etc.`

mackowski:
Apologies I read it one more time and it is better than I thought.

righettod:
No problem @mackowski it's very important to share the different opinions on this point 😃

Issue #114, opened 2019-05-22
Cheat sheet update/refactor proposal: Session Management Cheat Sheet


aussieklutz:
Thanks you for proposing a update/refactor of a cheat sheet.

Please provides the following information about your proposal:

- What update/refactoring do you want to perform? 

Web apps are starting to use the HTML5 Web Storage API's, in particular the window.sessionStorage api for session identifier storage. This provides protection from cross protocol (http -> https), cross-window, and retention beyond the window lifetime. It does however allow, by design, access from JS. It may also be vulnerable to TLS downgrade (cipher: null) attacks. Some addressing of this new mechanism would be valuable.

Thanks you again for your contribution :smiley:

mackowski:
Thanks for this proposal. Looks fine for me!
@righettod what is your view on that?

righettod:
This topic can effectively be interesting to deep dive to provide useful technical hints.

@aussieklutz Do you want to work on it?

aussieklutz:
Happy to draft a first pass at it...

aussieklutz:
Just made a pull request.

righettod:
Thank you very much, i have affected the issue to you 😃

righettod:
PR #115 merged, thanks again for your contribution 👍

PR #115, opened 2019-05-24
Add section on HTML5 Web Storage APIs

aussieklutz:
Address the mode of operation of the HTML5 Web Storage APIs (localStorage and sessionStorage) and potential security implications of using them to store session identifiers.

This PR covers issue #114.

### DIFF
--- a/cheatsheets/Session_Management_Cheat_Sheet.md
+++ b/cheatsheets/Session_Management_Cheat_Sheet.md
@@ -192,8 +192,8 @@ WHATWG suggests the use of `localStorage` for data that needs to be accessed acr
 ### Scope
 
 The `sessionStorage` API stores data within the window context from which it was called, meaning that Tab 1 cannot access data which was stored from Tab 2.
-Also, like the `localStorage` API, data stored using the `sessionStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`)..
-This provides similar access to this data as would be achieved by using the `secure` flag on a cookie, meaning that data stored from `https` could not be retrieved by `http`.
+Also, like the `localStorage` API, data stored using the `sessionStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`).
+This provides similar access to this data as would be achieved by using the `secure` flag on a cookie, meaning that data stored from `https` could not be retrieved via `http`.
 
 ### Duration
 
@@ -209,9 +209,9 @@ WHATWG suggests the use of `sessionStorage` for data that is relevant for one-in
 
 ## Security Risks
 
-In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage on shared systems. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than cookies with the `httponly` flag applied.
-While a case could be made for storing workflow specific data in `sessionStorage` for use by that specific tab/window across reloads, the Web Storage API's should be treated as insecure storage. Because of this, if a business solution requres the use of the `localStorage` or `sessionStorage` to store sensitive data, such a solution should encipher data and apply replay protections.
-Due to the potential to access Web Storage API's via an XSS attack, session identifiers should be stored using non-persistent cookies, with the appropriate flags to protect from [insecure access](Transport_Layer_Protection_Cheat_Sheet.md) (`Secure`), [XSS](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) (`HttpOnly`) and [CSRF](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md) issues (`SameSite`).
+In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage on shared systems. Because the Web Storage mechanisms are APIs, this also permits access from injected scripts, making it less secure than cookies with the `httponly` flag applied.
+While a case could be made for storing workflow specific data in `sessionStorage` for use by that specific tab/window across reloads, the Web Storage APIs should be treated as insecure storage. Because of this, if a business solution requires the use of the `localStorage` or `sessionStorage` to store sensitive data, such a solution should encipher data and apply replay protections.
+Due to the potential to access Web Storage APIs via an XSS attack, session identifiers should be stored using non-persistent cookies, with the appropriate flags to protect from [insecure access](Transport_Layer_Protection_Cheat_Sheet.md) (`Secure`), [XSS](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) (`HttpOnly`) and [CSRF](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md) issues (`SameSite`).
 
 ## References
 
--- a/cheatsheets/Session_Management_Cheat_Sheet.md
+++ b/cheatsheets/Session_Management_Cheat_Sheet.md
@@ -165,46 +165,60 @@ Typically, session management capabilities to track users after authentication m
 
 # HTML5 Web Storage API
 
-The Web Hypertext Application Technology Working Group (WHATWG) describes the HTML5 Web Storage APIs, localStorage and sessionStorage, as mechanisms for storing name-value pairs client-side.
-Unlike HTTP cookies, the contents of localStorage and sessionStorage are not automatically shared within requests or responses by the browser and are used for storing data client-side.
+The Web Hypertext Application Technology Working Group (WHATWG) describes the HTML5 Web Storage APIs, `localStorage` and `sessionStorage`, as mechanisms for storing name-value pairs client-side.
+Unlike HTTP cookies, the contents of `localStorage` and `sessionStorage` are not automatically shared within requests or responses by the browser and are used for storing data client-side.
 
 ## The localStorage API
 
 ### Scope
 
-Data stored using the localStorage API is accessible by pages which are loaded from the same protocol, port and host.
-This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved via http.
+Data stored using the `localStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`).
+This provides similar access to this data as would be achieved by using the `secure` flag on a cookie, meaning that data stored from `https` could not be retrieved via `http`. Due to potential concurrent access from separate windows/threads, data stored using `localStorage` may be susceptible to shared access issues (such as race-conditions) and should be considered non-locking ([Web Storage API Spec](https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute)). 
 
 ### Duration
 
-Data stored using the localStorage API is persisted across browsing sessions, meaning that this data is susceptible to access by other system users. 
+Data stored using the `localStorage` API is persisted across browsing sessions, extending the timeframe in which it may be accessible to other system users. 
+
+### Offline Access
+
+The standards do not require `localStorage` data to be encrypted-at-rest, meaning it may be possible to directly access this data from disk.
 
 ### Use Case
 
-WHATWG suggests the use of localStorage for data that needs to be accessed across windows or tabs, across multiple sessions, and where large (multi-megabyte) volumes of data may need to be stored for performance reasons.
+WHATWG suggests the use of `localStorage` for data that needs to be accessed across windows or tabs, across multiple sessions, and where large (multi-megabyte) volumes of data may need to be stored for performance reasons.
 
 ## The sessionStorage API
 
 ### Scope
 
-The sessionStorage API is accessible only to the window or tab that it was stored from, meaning that Tab 1 cannot access data for Tab 2.
-Also, like the localStorage API, data stored using the sessionStorage API is accessible by pages which are loaded from the same protocol, port and host.
-This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved by http.
+The `sessionStorage` API stores data within the window context from which it was called, meaning that Tab 1 cannot access data which was stored from Tab 2.
+Also, like the `localStorage` API, data stored using the `sessionStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`)..
+This provides similar access to this data as would be achieved by using the `secure` flag on a cookie, meaning that data stored from `https` could not be retrieved by `http`.
 
 ### Duration
 
-The sessionStorage API only stores data for the duration of the current browsing session. Once the tab is closed, that data is no longer retrievable.
+The `sessionStorage` API only stores data for the duration of the current browsing session. Once the tab is closed, that data is no longer retrievable. This does not necessarily prevent access, should a browser tab be reused or left open. Data may also persist in memory until a garbage collection event.
+
+### Offline Access
+
+The standards do not require `sessionStorage` data to be encrypted-at-rest, meaning it may be possible to directly access this data from disk.
 
 ### Use Case
 
-Despite the name, WHATWG does not suggest that sessionStorage be used to store session identifiers.
-Rather, WHATWG suggests the use of sessionStorage for data that is relevant for one-instance of a workflow, such as details for a ticket booking, but where multiple workflows could be performed in other tabs concurrently. The window/tab bound nature will keep the data from leaking between workflows in separate tabs.
+WHATWG suggests the use of `sessionStorage` for data that is relevant for one-instance of a workflow, such as details for a ticket booking, but where multiple workflows could be performed in other tabs concurrently. The window/tab bound nature will keep the data from leaking between workflows in separate tabs.
 
 ## Security Risks
 
-In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage to other users on the same host. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than a cookie with the “httponly” flags applied.
-These characteristics would restrict the use of localStorage to public data. While a case could be made for storing workflow specific data in sessionStorage for use by that specific tab/window across reloads, sensitive data should not be stored this way.
-An authentication request to an API endpoint should result in a Set-Cookie header with an appropriate identifier as raised above, which is restricted to the domain and path of the API, with no persistence, and with the recommended flags implemented.
+In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage on shared systems. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than cookies with the `httponly` flag applied.
+While a case could be made for storing workflow specific data in `sessionStorage` for use by that specific tab/window across reloads, the Web Storage API's should be treated as insecure storage. Because of this, if a business solution requres the use of the `localStorage` or `sessionStorage` to store sensitive data, such a solution should encipher data and apply replay protections.
+Due to the potential to access Web Storage API's via an XSS attack, session identifiers should be stored using non-persistent cookies, with the appropriate flags to protect from [insecure access](Transport_Layer_Protection_Cheat_Sheet.md) (`Secure`), [XSS](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) (`HttpOnly`) and [CSRF](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md) issues (`SameSite`).
+
+## References
+
+- https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
+- https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
+- https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
+- https://html.spec.whatwg.org/multipage/webstorage.html#webstorage
 
 # Session ID Life Cycle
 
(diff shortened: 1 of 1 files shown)


righettod:
Thank you very much for the PR, we will review it ASAP.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+# HTML5 Web Storage API
+
+The Web Hypertext Application Technology Working Group (WHATWG) describes the HTML5 Web Storage APIs, localStorage and sessionStorage, as mechanisms for storing name-value pairs client-side.
+Unlike HTTP cookies, the contents of localStorage and sessionStorage are not automatically shared within requests or responses by the browser and are used for storing data client-side.
+
+## The localStorage API
+
+### Scope
+
+Data stored using the localStorage API is accessible by pages which are loaded from the same protocol, port and host.
```
righettod: According to [this](https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute) + [this](https://html.spec.whatwg.org/multipage/origin.html#concept-origin), the origin that define access policy is defined by:
* A scheme (a scheme).
* A host (a host).
* A port (a port).
* A domain (null or a domain). Null unless stated otherwise. 

So it's important to mention the 4 elements explicitly.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved via http.
+
+### Duration
+
+Data stored using the localStorage API is persisted across browsing sessions, meaning that this data is susceptible to access by other system users.
```
righettod: It's the same for `sessionStorage` during the time when the tab is opened. 

So, it's important to explictly mention that data stored by the browser are not ciphered when they are stored for both `sessionStorage` and `localStorage`.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+Data stored using the localStorage API is accessible by pages which are loaded from the same protocol, port and host.
+This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved via http.
+
+### Duration
+
+Data stored using the localStorage API is persisted across browsing sessions, meaning that this data is susceptible to access by other system users. 
+
+### Use Case
+
+WHATWG suggests the use of localStorage for data that needs to be accessed across windows or tabs, across multiple sessions, and where large (multi-megabyte) volumes of data may need to be stored for performance reasons.
```
righettod: It can be interesting to point to the warning about the concurrent RW access mentioned [here](https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute) at the bottom of the section:

 > The localStorage attribute provides access to shared state. This specification does not define the interaction with other browsing contexts in a multiprocess user agent, and authors are encouraged to assume that there is no locking mechanism. A site could, for instance, try to read the value of a key, increment its value, then write it back out, using the new value as a unique identifier for the session; if the site does this twice in two different browser windows at the same time, it might end up using the same "unique" identifier for both sessions, with potentially disastrous effects.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+
+## The sessionStorage API
+
+### Scope
+
+The sessionStorage API is accessible only to the window or tab that it was stored from, meaning that Tab 1 cannot access data for Tab 2.
+Also, like the localStorage API, data stored using the sessionStorage API is accessible by pages which are loaded from the same protocol, port and host.
```
righettod: Same remark like above the `origin` criteria defining the access policy.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved by http.
+
+### Duration
+
+The sessionStorage API only stores data for the duration of the current browsing session. Once the tab is closed, that data is no longer retrievable.
```
righettod: I questioning myself if data stored by a tab in a `sessionStorage` still exist if the browser crash (ex: plugin error or browser instability), do you have some hint about this point?

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: Add link to the part of the WHATWG  spec where they mention the fact mentioned.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+
+### Use Case
+
+Despite the name, WHATWG does not suggest that sessionStorage be used to store session identifiers.
```
righettod: Add link to the part of the WHATWG  spec where they mention the fact mentioned.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+Rather, WHATWG suggests the use of sessionStorage for data that is relevant for one-instance of a workflow, such as details for a ticket booking, but where multiple workflows could be performed in other tabs concurrently. The window/tab bound nature will keep the data from leaking between workflows in separate tabs.
+
+## Security Risks
+
+In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage to other users on the same host. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than a cookie with the “httponly” flags applied.
```
righettod: If really needed by a business situation, you can store data using these storage but **it must be ciphered with user context additional information in order to prevent stealing + replay attack**.

righettod: When you say *on the same host*, you refer to a shared computer with a shared system user in a way that you can access the browser DB file where info are stored?

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+These characteristics would restrict the use of localStorage to public data. While a case could be made for storing workflow specific data in sessionStorage for use by that specific tab/window across reloads, sensitive data should not be stored this way.
+An authentication request to an API endpoint should result in a Set-Cookie header with an appropriate identifier as raised above, which is restricted to the domain and path of the API, with no persistence, and with the recommended flags implemented.
```
righettod: Here you recommands to use cookies to carry authentication token ?

If yes, then, add a warning about CSRF and point to the to the dedicated CS, precisely the section about the attribute [SameSite](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#samesite-cookie-attribute) (use relative link because CS are in the same folder).

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,49 @@ Typically, session management capabilities to track users after authentication m
...
+
```
righettod: Add the following documentation links to MDN and WHATWG:
* https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
* https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
* https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage
* https://html.spec.whatwg.org/multipage/webstorage.html#webstorage

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: This is a bit hard. Each UA is independent from the other. Check the note in [here](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API#Web_Storage_concepts_and_usage). A simple note can be mentioned about UA crashes, but no hard details on it.

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: I believe they're referencing the shared system. It's a good point from you so that it is explicitly specified as `host` can be used in different contexts.

ThunderSon: This should be added as a note for folks that need it. More implementation though is needed to properly save the session and secure it, don't you think? If you have any references to the comment, please mention them.

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: The author refrained from specifying any flag though. Why only reference same site flag?

ThunderSon: Some heavy JS based frameworks (Angular, React) and other frameworks use the `Authorization` header instead of setting a cookie. Do you think we should look into both cases?

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: OK, thanks for the remark 😃 so forget my note.

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: It was just a remark from me about the fact that you have the same issue from someone having access to the machine file system whatever the kind of storage used. The only difference is the timeframe in which the data is stored.

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: For sure, i was thinking to only add a link but a summary, as you proposed, is good too 😃

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: Done 😃

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: 100% agree, it's a very special case, I was just think of it so it's why i have added the remark. 
We can just add a note about this situation and create a ticket to add an example of implementation later.
What do you think?

righettod: As I was questioning myself, I have asked the question 😃

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: As the author propose to use cookie to carry authentication token then we must also take in account the CSRF attack, as the HttpOnly flag has been already mentioned and prevent access from JS (XSS).

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: Agree on that point, It was just a internal question from myself 😃

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: I don't mind at all to stay hand in hand with the official docs.

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: For the mentioned point, I believe there needs to be another section for it. Keep things well organized 😄 I am not against it.

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: Certainly, if you feel like it'll take time and push this PR further away. We'll await the author's input.

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: 100% agree

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: It's just to have time to create a good and useful example 😃

cheatsheets/Session_Management_Cheat_Sheet.md
righettod: Let's wait the feedback from the author.

aussieklutz: Happy to modify to use the scheme/host/port nomenclature.

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: Raised in a separate heading regarding Offline Access

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: Added to Scope section

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: Was a negative example, namely that they do not mention using session storage to store session identifiers. I can see how this could be leading, and will remove.

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: Rather than discouraging the use of these api's, will suggest that it be treated as insecure storage, and that additional safeguards such as ciphering and replay protection may be necessary to protect sensitive data.

aussieklutz: Rephrased to use the term "shared system"

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: Rephrased

cheatsheets/Session_Management_Cheat_Sheet.md
ThunderSon: Please go ahead. Switching between the `host` and `domain` sometimes makes a difference with `same-site` and `same-origin`.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -165,46 +165,60 @@ Typically, session management capabilities to track users after authentication m
...
 ## The sessionStorage API
 
 ### Scope
 
-The sessionStorage API is accessible only to the window or tab that it was stored from, meaning that Tab 1 cannot access data for Tab 2.
-Also, like the localStorage API, data stored using the sessionStorage API is accessible by pages which are loaded from the same protocol, port and host.
-This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved by http.
+The `sessionStorage` API stores data within the window context from which it was called, meaning that Tab 1 cannot access data which was stored from Tab 2.
+Also, like the `localStorage` API, data stored using the `sessionStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`)..
+This provides similar access to this data as would be achieved by using the `secure` flag on a cookie, meaning that data stored from `https` could not be retrieved by `http`.
```
ThunderSon: `via` instead of `by`

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -165,46 +165,60 @@ Typically, session management capabilities to track users after authentication m
...
 
 ## The sessionStorage API
 
 ### Scope
 
-The sessionStorage API is accessible only to the window or tab that it was stored from, meaning that Tab 1 cannot access data for Tab 2.
-Also, like the localStorage API, data stored using the sessionStorage API is accessible by pages which are loaded from the same protocol, port and host.
-This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved by http.
+The `sessionStorage` API stores data within the window context from which it was called, meaning that Tab 1 cannot access data which was stored from Tab 2.
+Also, like the `localStorage` API, data stored using the `sessionStorage` API is accessible by pages which are loaded from the same origin, which is defined as the scheme (`https:\\`), host (`example.com`), port (`443`) and domain/realm (`example.com`)..
```
ThunderSon: Double ending period

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -165,46 +165,60 @@ Typically, session management capabilities to track users after authentication m
...
+WHATWG suggests the use of `localStorage` for data that needs to be accessed across windows or tabs, across multiple sessions, and where large (multi-megabyte) volumes of data may need to be stored for performance reasons.
 
 ## The sessionStorage API
 
 ### Scope
 
-The sessionStorage API is accessible only to the window or tab that it was stored from, meaning that Tab 1 cannot access data for Tab 2.
-Also, like the localStorage API, data stored using the sessionStorage API is accessible by pages which are loaded from the same protocol, port and host.
-This provides similar access to this data as would be achieved by using the "secure" flag on a cookie, meaning that data stored from https could not be retrieved by http.
+The `sessionStorage` API stores data within the window context from which it was called, meaning that Tab 1 cannot access data which was stored from Tab 2.
```
ThunderSon: Are you sure that Tab 1 can't access what Tab 2 stored as long as they follow the same-origin rules and the session is not closed yet?

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -165,46 +165,60 @@ Typically, session management capabilities to track users after authentication m
...
-An authentication request to an API endpoint should result in a Set-Cookie header with an appropriate identifier as raised above, which is restricted to the domain and path of the API, with no persistence, and with the recommended flags implemented.
+In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage on shared systems. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than cookies with the `httponly` flag applied.
+While a case could be made for storing workflow specific data in `sessionStorage` for use by that specific tab/window across reloads, the Web Storage API's should be treated as insecure storage. Because of this, if a business solution requres the use of the `localStorage` or `sessionStorage` to store sensitive data, such a solution should encipher data and apply replay protections.
```
ThunderSon: `requires` instead of `requres`

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,63 @@ Typically, session management capabilities to track users after authentication m
...
+
+The standards do not require `sessionStorage` data to be encrypted-at-rest, meaning it may be possible to directly access this data from disk.
+
+### Use Case
+
+WHATWG suggests the use of `sessionStorage` for data that is relevant for one-instance of a workflow, such as details for a ticket booking, but where multiple workflows could be performed in other tabs concurrently. The window/tab bound nature will keep the data from leaking between workflows in separate tabs.
+
+## Security Risks
+
+In general, secure or sensitive data should not be stored persistently in browser data stores as this may permit information leakage on shared systems. Because the Web Storage mechanisms are API's, this also permits access from injected scripts, making it less secure than cookies with the `httponly` flag applied.
```
ThunderSon: APIs and not API's

ThunderSon:
@aussieklutz awesome update! I like how you modified it.

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: For example, a page could have a checkbox that the user ticks to indicate that they want insurance:

<label>
 <input type="checkbox" onchange="sessionStorage.insurance = checked ? 'true' : ''">
  I want insurance on this trip.
</label>
A later page could then check, from script, whether the user had checked the checkbox or not:

if (sessionStorage.insurance) { ... }
If the user had multiple windows opened on the site, each one would have its own individual copy of the session storage object.
-- https://html.spec.whatwg.org/multipage/webstorage.html#introduction-15

aussieklutz: An exclusive sessionData object is instantiated for each origin in each browsing context (ie tab or window) that is opened. The data that is written into one sessionData object is not shared to other sessionData objects (other origins or other browsing contexts).

aussieklutz:
Thanks @ThunderSon, I have implemented the further comments you made.

cheatsheets/Session_Management_Cheat_Sheet.md
aussieklutz: I think I have resolved this. Please let me know if there are additional concerns...

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -163,6 +163,63 @@ Typically, session management capabilities to track users after authentication m
...
+While a case could be made for storing workflow specific data in `sessionStorage` for use by that specific tab/window across reloads, the Web Storage APIs should be treated as insecure storage. Because of this, if a business solution requires the use of the `localStorage` or `sessionStorage` to store sensitive data, such a solution should encipher data and apply replay protections.
+Due to the potential to access Web Storage APIs via an XSS attack, session identifiers should be stored using non-persistent cookies, with the appropriate flags to protect from [insecure access](Transport_Layer_Protection_Cheat_Sheet.md) (`Secure`), [XSS](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) (`HttpOnly`) and [CSRF](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md) issues (`SameSite`).
+
+## References
+
+- https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API
```
righettod: Use named linked, change this section to this:
- [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API)
- [LocalStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
- [SessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
- [WHATWG Web Storage Spec](https://html.spec.whatwg.org/multipage/webstorage.html#webstorage)

aussieklutz: Updated.

Issue #91, opened 2019-05-01
Remove the "Authors and Primary Editors" section


righettod:
In order to only use the GH feature for contributors, the section **Authors and Primary Editors** of all CS will be removed.

The information will be transfered in a file named **CONTRIBUTOR-V1.md** in which there will be the list of contributors (of the project V1 in fact)  and the associated CSs.

righettod:
We will perform this task near the end of this year in order to let the commit history grow.

righettod:
I start actively working on it.

righettod:
Linked to PR #169

righettod:
PR merged.

PR #117, opened 2019-05-30
Add JavaScript guidance for AJAX calls on including CSRF tokens by default

arunjohnkuruvilla:
Thank you for submitting a Pull Request to the Cheat Sheet Series. 

Thanks you again for your contribution :smiley:


cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -523,22 +523,141 @@ public class CSRFValidationFilter implements Filter {
...
 ```
+# JavaScript Guidance for Auto-inclusion of CSRF tokens as an AJAX Request header
+
+The following guidance considers **GET**, **HEAD** and **OPTIONS** methods are safe operations. Therefore **GET**, **HEAD**, and **OPTIONS** method AJAX calls need not be appended with a CSRF token header. However, if the verbs are used to perform state changing operations, they will also require a CSRF token header.
+
+**POST**, **PUT**, **PATCH**, **DELETE** and **TRACE** verbs, being state changing verbs, should have a CSRF token attached to the request. The following guidance will demonstrate how to create overrides in JavaScript libraries to have CSRF tokens included automatically with every AJAX request for the state changing methods mentioned above.
+
+## Storing the CSRF Token Value in the DOM
+
+A CSRF token can be included in the <meta> tag as shown below. All subsequent calls in the page can extract the CSRF token from this <meta> tag. It can also be stored in a JS variable / anywhere on the DOM. However, it is not recommended to store it in cookies or browser local storage.
```
mackowski: Explain why it is not recommended to store it in cookies or browser local storage.
I understand that this is not recommended because of the reasons that were explained before [in double submit cookie](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#double-submit-cookie) - if so please refer/link to it.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -551,3 +670,5 @@ Paul Petefish
 Eric Sheridan - eric.sheridan@owasp.org
 
 Dominique Righetto - dominique.righetto@owasp.org
+
+[Arun John Kuruvilla (Amazon Application Security Team)](https://github.com/arunjohnkuruvilla)
```
mackowski: Please do not add your name because we want to use GH feature for contributors. This sections will be deleted near the end of this year - see issue #91.

mackowski:
I think that in case of Angular we should also explain Angular's CSRF protection build in feature. @arunjohnkuruvilla do you want to cover that since you crafted Angular's example?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
arunjohnkuruvilla: Removed.

arunjohnkuruvilla:
I can explain Angular's CSRF feature as a separate PR later on.

ThunderSon:
Solid update. What do you think about the naming of the header? By convention, any user set header should be X-<What it is>, so in our case `X-CSRF-Token`. And by looking at other frameworks, they use XSRF-Token for cookies and X-XSRF-Token for header. What do you think about following the same trend?

arunjohnkuruvilla:
> Solid update. What do you think about the naming of the header? By convention, any user set header should be `X-<What it is>`, so in our case `X-CSRF-Token`. And by looking at other frameworks, they use XSRF-Token for cookies and X-XSRF-Token for header. What do you think about following the same trend?

The name is just an example. The application developer can change the header to `X-CSRF-Token` if desired as per use case.

PR #122, opened 2019-06-07
Add missing space

horgh:


righettod:
Hello, I cannot identify the space fixed, can you spot me the one fixed?
Thanks in advance.

mackowski:
Fix is in "In a negative access model,you have to be much more careful to" space was added after comma.

"In a negative access model,you have to be much more careful to" -> "In a negative access model, you have to be much more careful"

PR #123, opened 2019-06-07
Add a reference to the JWT BCP

yaronf:
Apologies for the self-advertisement, but I think this would be useful for developers that follow the JWT cheatsheet.

### DIFF
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
@@ -603,6 +603,10 @@ You can also use:
 - [JohnTheRipper](https://github.com/hashcat/hashcat/issues/1057#issuecomment-279651700) to perform the password dictionary attack.
 - [Hashcat](https://hashcat.net/hashcat/).
 
+### Additional Reading
+
+- [JWT Best Practices Internet Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-jwt-bcp/)
+
 # Authors and Primary Editors
 
 Jim Manico - jim.manico@owasp.org


righettod:
I will review it ASAP :)

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -603,10 +603,14 @@ You can also use:
...
+
 # Authors and Primary Editors
 
 Jim Manico - jim.manico@owasp.org
 
 Dominique Righetto - dominique.righetto@owasp.org
 
-Paul Ionescu - paul.ionescu@owasp.org
\ No newline at end of file
+Paul Ionescu - paul.ionescu@owasp.org
```
ThunderSon: Kindly remove the author name as we are looking to shift to Github's contributors model.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -603,10 +603,14 @@ You can also use:
 - [JohnTheRipper](https://github.com/hashcat/hashcat/issues/1057#issuecomment-279651700) to perform the password dictionary attack.
 - [Hashcat](https://hashcat.net/hashcat/).
 
+### Additional Reading
+
+- [JWT Best Practices Internet Draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-jwt-bcp/)
```
ThunderSon: So far we are not accepting `Drafts` in general, as in, work in progress projects. @righettod what do you think about this one though?

yaronf:
Confused - why was it assigned to me? I proposed a PR, and I'm not seeing any suggested changes.

ThunderSon:
@yaronf I assigned you so that you can check out the feedback I left and tell me your thoughts

yaronf:
Yes this is draft, i.e. work in progress. However it is now in "IESG evaluation" (a stage in the IETF standardization process that's very late), which means that most likely there will not be any major changes. It could take another 6 months before this is published as RFC, it's fine if you prefer to wait but personally I think it's a useful resource in its current form.

PR #137, opened 2019-06-19
Add warning that RFC 5321 email addresses can include XSS payloads

rbsec:
There have been some fairly high profile vulnerabilities (such as one in [Magento](https://blog.sucuri.net/2016/01/security-advisory-stored-xss-in-magento.html)) where the assumption that valid (RFC 5321) email addresses are safe, when they can in fact include XSS payloads and other dangerous strings.

This adds a warning that you should treat them as untrusted  even if they've validated as proper email addresses.

### DIFF
--- a/cheatsheets/Input_Validation_Cheat_Sheet.md
+++ b/cheatsheets/Input_Validation_Cheat_Sheet.md
@@ -188,6 +188,8 @@ Following [RFC 5321](https://tools.ietf.org/html/rfc5321), best practice for val
 - Ensure the domain is no longer than **255 octets**.
 - Ensure the address **is deliverable**.
 
+Note that [RFC 5321](https://tools.ietf.org/html/rfc5321) allows potentially dangerous characters in email addresses. For example, `"><script>alert(1);</script>"@example.org` is a valid email address. As such, it should not be assumed that strings which are valid email addresses are safe to display unencoded or to include in SQL queries.
+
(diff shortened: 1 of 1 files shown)


cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -188,6 +188,8 @@ Following [RFC 5321](https://tools.ietf.org/html/rfc5321), best practice for val
 - Ensure the domain is no longer than **255 octets**.
 - Ensure the address **is deliverable**.
 
+Note that [RFC 5321](https://tools.ietf.org/html/rfc5321) allows potentially dangerous characters in email addresses. For example, `"><script>alert(1);</script>"@example.org` is a valid email address. As such, it should not be assumed that strings which are valid email addresses are safe to display unencoded or to include in SQL queries.
```
righettod: [here](https://tools.ietf.org/html/rfc5321#section-4.1.2) for the `Local-part` I see that quoted string is allowed but I didn't find where the RFC allow the `( ) < / >` characters. 

It is present into the RFC?

rbsec: `Local-part` allows `Quoted-string`
`Quoted-string` is defined as `DQUOTE *QcontentSMTP DQUOTE`
`QcontentSMTP` allows `qtextSMTP`
`qtextSMTP` is defined as `%d32-33 / %d35-91 / %d93-126`
 Which includes `<` (60) and `>` (62), as well as pretty much any other ASCII character

righettod: OK thank you 😃

rbsec: I'm sure the RFC authors get used to it, but for the rest of us BNF is a real pain to try and read..

Issue #151, opened 2019-07-09
Update Cross-Site Request Forgery Prevention Cheat Sheet


mackowski:
This CS is very big. It is a good resource about CSRF but it is not a cheat sheet per say.

I propose to delete or strongly refactor sections:
 - What's new?
 - Not So Popular CSRF Mitigations
 - CSRF Mitigation Myths
 - Personal Safety CSRF Tips for Users

Refactor (to make it shorter) all Defense In Depth Techniques descriptions

mackowski:
I want to create PR this week.
Do you have any other suggestions what should be changed in this CS?

righettod:
I'm in sync with the update that you want to perform and your objective 😃

righettod:
PR #160 merged

PR #160, opened 2019-07-18
CSRF update 

mackowski:

This PR covers issue  #151


righettod:
I will review it this week end 😃

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -45,7 +41,7 @@ This defense is one of the most popular and recommended methods to mitigate CSRF
...
+Any state changing operation requires a secure random token (e.g., CSRF token) to prevent CSRF attacks. A CSRF token should be unique per user session, large random value, and also generated by a cryptographically secure random number generator. The CSRF token is added as a hidden field for forms, headers/parameters for AJAX calls (If you need to add it to header, it is better to make sure that the token header is not being logged at your server) and within the URL if the state changing operation occurs via a GET. See "Disclosure of Token in URL" section below. The server rejects the requested action if the CSRF token fails validation.
```
righettod: Add a link to the section instead of the name "Disclosure of Token in URL"

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -135,6 +127,27 @@ If sub-domains under your master domain are treated as not trusty in your threat
...
+
+The default Lax value provides a reasonable balance between security and usability for websites that want to maintain user's logged-in session after the user arrives from an external link. In the above GitHub scenario, the session cookie would be allowed when following a regular link from an external website while blocking it in CSRF-prone request methods such as POST. Only cross-site-requests that are allowed in Lax mode are the ones that have top-level navigations and are also [safe](https://tools.ietf.org/html/rfc7231#section-4.2.1) HTTP methods.
+
+For more details on the `SameSite` values, check the following [section](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7.1) from the [rfc](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02).
```
righettod: Add a section explaining what does the `None` value like for the 2 others ones.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -242,7 +217,7 @@ If this is the case for your system, you can simply verify the presence of this
...
+This technique obviously works for Ajax calls and you have to still need protect `<form>` tags with approaches described in this document such as tokens. Setting custom headers on form tags is not possible directly. Also, CORS configuration should also be robust to make this solution work effectively (as custom headers for requests coming from other domains trigger a pre-flight CORS check).
```
righettod: *Setting custom headers on form tags is not possible directly*

Explains what are the possibilities to address that or add a ref to the section of the document talking about this.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
righettod: Check if you can handle the grammar fix of the PR #151 in your PR directly

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: @ThunderSon you recently updated `SameSite` section. From what I remember there was a reason why `None` is not covered.
I know that `None` will be soon treated different in Chrome

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: done!

manideepkonakandla:
@mackowski Awesome efforts in making the cheatsheet simple. Thanks :)

Since I originally wrote/rewrote some of these sections and as this seems like major update/refactor, do you think this can wait till I review it? I wish I could review it now but I'm on vacation in INDIA till end of July with very limited laptop access. I can review it in early first week of August. Will really appreciate if this can wait till then.

mackowski:
> @mackowski Awesome efforts in making the cheatsheet simple. Thanks :)
> 
> Since I originally wrote/rewrote some of these sections and as this seems like major update/refactor, do you think this can wait till I review it? I wish I could review it now but I'm on vacation in INDIA till end of July with very limited laptop access. I can review it in early first week of August. Will really appreciate if this can wait till then.

@manideepkonakandla of course! Your input is very valuable. There is no rush with this but my goal is to make this CS simpler and shorter. Now it is becoming a pretty long and exhaustive article. We can even think to split it to more cheat sheets if needed.

manideepkonakandla:
Update: I had to extend my vacation due to some personal emergencies and hence the delay. All set now. I will review this week and update by Sunday.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -190,44 +190,27 @@ If you are behind a proxy, there are a number of options to consider.
...
 
-Usually, a minor percentage of traffic does fall under above categories ([1-2%](http://homakov.blogspot.com/2012/04/playing-with-referer-origin-disquscom.html)) and no enterprise would want to lose even this minor % of traffic. One of the popular technique used across Internet to make this technique more usable is to accept the request if the Origin/referrer matches your configured list of domains "OR" a null value (Examples [here](http://homakov.blogspot.com/2012/04/playing-with-referer-origin-disquscom.html). null value is to cover the edge cases mentioned above where these headers are not sent). Please note that, attackers can use it to exploit CSRF using the edge cases but many prefer to deploy this technique it as a defense in depth measure because of minor effort involved in creating it.
```
manideepkonakandla: I am fine removing the last three paragraphs but the first paragraph is what that makes this technique "usable" as even as "defense in depth".

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -190,44 +190,27 @@ If you are behind a proxy, there are a number of options to consider.
...
 
-So, unless you are sure that your subdomains are fully secured and only accept HTTPS connections (we believe it’s difficult to guarantee at large enterprises), you should not rely on the Double Submit Cookie technique as a primary mitigation for CSRF.
-
-Scenarios a and b mentioned above are possible only if the CSRF token is not derived/tied to the session in which case an attacker can overwrite the token in the parent domain cookie with XSS in child domain. A variant of how this can be mitigated by linking token and session/auth cookie is explained below.
```
manideepkonakandla: Not sure why this is removed. I added this workaround to mitigate the two issues mentioned in this technique above.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -135,6 +131,27 @@ If sub-domains under your master domain are treated as not trusty in your threat
...
+Example of cookies using this attribute:
+
+```
+Set-Cookie: JSESSIONID=xxxxx; SameSite=Strict
+Set-Cookie: JSESSIONID=xxxxx; SameSite=Lax
+```
+
+All desktop browsers and almost all mobile browsers now support the `SameSite` attribute. To keep track of the browsers implementing it and the usage of the attribute, refer to the following [service](https://caniuse.com/#feat=same-site-cookie-attribute).
+
+It is important to note that this attribute should be implemented as an additional layer *defense in depth* concept. This attribute protects the user through the browsers supporting it, and it contains as well 2 ways to bypass it as mentioned in the following [section](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7.1). This attribute should not replace having a CSRF Token. Instead, it should co-exist with that token in order to protect the user in a more robust way.
```
manideepkonakandla: In the earlier paragraph, we mention that all desktop browsers and almost all mobile browsers now support this attribute and here we say that "protects the user through the browsers supporting it". Its not very clear on what we are trying to explain. How about we re-frame the second section as below??

Even though most newer browser versions support this attribute, old versions do not in which case browsers just ignore this attribute sent by the server. Since we cannot rely on user browsers for providing security, this defense should only be implemented as an *defense in depth* concept. This defense also contains 2 ways to bypass it as mentioned in the following [section](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7.1). Hence, it should not replace having a CSRF Token and should co-exist with it in order to protect users in a more robust way.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -254,39 +250,6 @@ While all the techniques referenced here do not require any user interaction, so
 
 While these are a very strong CSRF defense, it does create a huge impact on the user experience. For applications that are in need of high security for some operations (password change, money transfer etc.), these techniques should be used along with token based mitigation. Please note that tokens by themselves can mitigate CSRF, developers should use these techniques only to achieve additional security for their high sensitive operations.
 
-# Not So Popular CSRF Mitigations
```
manideepkonakandla: I  personally know lot of developers who think content-type header validation can be used as a CSRF defense and know few folks who think triple submit cookie is better than double submit cookie.

My understanding of cheat sheet is that when people come to it, they need to know whats right way to do and whats wrong way. We dont need to highlight all possible "wrong" ways but at least mention the most widely known one's. Hence, I recommend adding these things back. We can add it with different title/ strip them to a line/two instead of how they are expanded now. What do you guys think?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: I remember adding this statement long back. I think its safe to remove it.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -135,6 +131,27 @@ If sub-domains under your master domain are treated as not trusty in your threat
 
 # Defense In Depth Techniques
 
+## Samesite Cookie Attribute
```
manideepkonakandla: Any specific reason on why we are moving this section ahead in "defense in depth" measures?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: I agree, it would be awesome to leave them but in shorter from!

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: Ok so I will delete this

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: Agree! I will add them again

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: true, I will add it again

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: This PR is for for improving Samesite Cookie Attribute I just moved this paragraph without changes. You can open a new issue for it.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: The same answer as I give to Dom - This PR is for for improving Samesite Cookie Attribute I just moved this paragraph without changes. You can open a new issue for it.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: Because it is the easiest one and IMO will add at least the same level of security as other techniques in this section. In addition to that this is true defense in depth because it protect against CSRF on other layer (browser instead of server side).

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -194,6 +194,8 @@ This mitigation is working properly when origin or referrer headers are present
...
 
+Usually, a minor percentage of traffic does fall under above categories ([1-2%](http://homakov.blogspot.com/2012/04/playing-with-referer-origin-disquscom.html)) and no enterprise would want to lose even this minor % of traffic. One of the popular technique used across Internet to make this technique more usable is to accept the request if the Origin/referrer matches your configured list of domains "OR" a null value (Examples [here](http://homakov.blogspot.com/2012/04/playing-with-referer-origin-disquscom.html). null value is to cover the edge cases mentioned above where these headers are not sent). Please note that, attackers can use it to exploit CSRF using the edge cases but many prefer to deploy this technique it as a defense in depth measure because of minor effort involved in creating it.
```
manideepkonakandla: Better rephrase? (I understand that I initially wrote it but on re-reading it, I felt it can be improved a bit)

_"Please note that, attackers can exploit this but people prefer to use this technique as a defense in depth measure because of minor effort involved in deploying it "_

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Yup, fine with me.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: Once this change is made and stripped version of "not so popular mitigation" techniques is added , I am fine with this pull request. Thanks again for this refactoring :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -45,7 +41,7 @@ This defense is one of the most popular and recommended methods to mitigate CSRF
...
+Any state changing operation requires a secure random token (e.g., CSRF token) to prevent CSRF attacks. A CSRF token should be unique per user session, a large random value, and also generated by a cryptographically secure random number generator. The CSRF token is added as a hidden field for forms, headers/parameters for AJAX calls (If you need to add it to header, it is better to make sure that the token header is not being logged at your server) and within the URL if the state changing operation occurs via a GET. See "Disclosure of Token in URL" section below. The server rejects the requested action if the CSRF token fails validation.
```
ThunderSon: 1. Why are we saying that most practices implement CSRF in a hidden field? We can simply mention the techniques (in the headers, hidden fields, etc.) to do and where they can be used (forms, AJAX calls, etc.)
2. Instead of saying `See Disclosure of Token ...`, directly link to the section

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -81,14 +77,10 @@ Synchronizer token defenses have been built into many frameworks, so we strongly
 
 **Disclosure of Token in URL**
```
ThunderSon: Why is this not a subsection?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
ThunderSon: Exactly. This shouldn't be handled in this PR.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: @ThunderSon I am not sure if I got your first point. Where did we mention that "most practices implement CSRF in a hidden field"?? We mentioned "The CSRF token is added as a hidden field for forms, headers/parameters for AJAX calls"

ThunderSon: Exactly. Who said that? We can simply mention that CSRF token can be added through hidden fields, headers, additional parameters in forms, AJAX calls, etc. Just mention all possible techniques to add the token for all state changing actions

righettod:
Any updates regarding this PR?

mackowski:
> Any updates regarding this PR?

I have to fix minor issues after @ThunderSon review last week and it will be ready to merge.

righettod:
Thank you for the update 😉

mackowski:
@ThunderSon @righettod please review all the issues should be resolved :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -45,7 +41,7 @@ This defense is one of the most popular and recommended methods to mitigate CSRF
...
+Any state changing operation requires a secure random token (e.g., CSRF token) to prevent CSRF attacks. A CSRF token should be unique per user session, a large random value, and also generated by a cryptographically secure random number generator. The CSRF token can be added through hidden fields, headers, additional parameters in forms, AJAX calls and within the URL. Token should be added to all state changing actions. Make sure that token will not leakue for example in server logs (e.g. when server log all of the headers) or [in the URL](#disclosure-of-token-in-url). The server rejects the requested action if the CSRF token fails validation.
```
ThunderSon: What do you think about the following suggestion where I:
- Improved the way the CSRF token is described
- Improved the ease of readability
- Removed sending the CSRF token as part of the URL. It doesn't make sense sending it in the URL and stating right after that URLs get logged and we are saying not to log it?
- Made a small separation between where the CSRF token is added, and what functions use it.
```suggestion
Any state changing operation requires a secure random token (e.g., CSRF token) to prevent CSRF attacks. A CSRF token should be:

- unique per user session;
- large random value;
- generated by a Cryptographically Secure Pseudo-Random Number Generator (CSPRNG).

The CSRF token can be added through hidden fields, headers, and can be used with forms, and AJAX calls. That token should be added to all state changing actions. Make sure that the token is not leaked in the server logs, or [in the URL](#disclosure-of-token-in-url). The server must reject the requested action if the CSRF token fails validation.
```

mackowski: Looks much better! 
I see also one one change that we can add here. In the first sentence we write "Any state changing operation requires a secure random token" and that we repet it "That token should be added to all state changing actions". Maybe we should delete this repetition?

ThunderSon: Certainly!

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: done :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: done

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: it is now :)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
manideepkonakandla: I have not seen many code samples where people include tokens in headers with form submissions and hence suggested to mention it considering the regular trend. Anyway, I am fine with the way its edited now.

PR #161, opened 2019-07-18
Fix grammar in CSRF Prevention Cheat Sheet

alexsaalberg049:
https://en.wikipedia.org/wiki/Parallelism_(grammar)


righettod:
Thank you very much for your PR.
We are checking if we can handle it via the PR #161 directly

mackowski:
I fixed it in #160

ThunderSon:
As both PRs tackle the same CS and are inclusive to each other, it was best to let #160 handle it. @alexsaalberg049 looking forward to you identifying more issues in the project 😄

PR #174, opened 2019-08-06
Fix small typo in TLS Cipher String cheat sheet

james-otten:
Fix a small typo.


jmanico:
“When you’re willing to help us fix the small errors like this, it helps us see the bigger problems more clearly.”

Someone wise once told me that. Not sure if it’s true but it sounds good.

THANK YOU!

--
Jim Manico
@Manicode

PR #175, opened 2019-08-07
Reference long-password DoS

Neil-Smithline:
Mention the importance of setting a password maximum to prevent long password DoSs.

### DIFF
--- a/cheatsheets/Authentication_Cheat_Sheet.md
+++ b/cheatsheets/Authentication_Cheat_Sheet.md
@@ -25,7 +25,7 @@ A key concern when using passwords for authentication is password strength. A "s
 
 - Password Length
     - **Minimum** length of the passwords should be **enforced** by the application. Passwords **shorter than 8 characters** are considered to be weak ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)). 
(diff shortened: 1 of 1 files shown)


cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -25,7 +25,7 @@ A key concern when using passwords for authentication is password strength. A "s
 
 - Password Length
     - **Minimum** length of the passwords should be **enforced** by the application. Passwords **shorter than 8 characters** are considered to be weak ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)). 
-    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. Typical maximum length is 128 characters. 
+    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. Typical maximum length is 128 characters. It is important to set a maximum password length to prevent [long password Denail of Service attacks](https://www.helpnetsecurity.com/2013/09/17/too-long-passwords-can-dos-some-servers/).
```
ThunderSon: Denial typo.
I'd prefer linking the following article, as it is more technical and is more appropriate: https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/

Neil-Smithline: Happy to switch, though I think they are both crappy links. Not sure why this isn't better documented. Thanks for following up.

ThunderSon: I believe cz it is a DoS attack, and most frameworks (which are the bigger part of the web) have certain field restrictions in place (if you notice, Drupal and Wordpress had those bugs like 6 years ago, and fixed them).
The second point might be cause Application Security is slowly booming, and not that much thought was put into it back then?

PR #176, opened 2019-08-08
PR not referencing issues

ThunderSon:
- Remove backticks around the issue number.
- Remove bold styling


ThunderSon:
@righettod I think you need to make the PR checks run only on certain file types?

PR #188, opened 2019-08-27
Add discussion of using SameSite cookies to protect against Clickjacking

rbsec:
The `SameSite` cookie attribute is quite well known as a defence against CSRF, but it can also provide protection against Clickjacking attacks. When the attribute is set to either `strict` or `lax`, the cookies won't be sent by requests inside an `<iframe>`, which means that CSRF is only exploitable if it works anonymously.

This doesn't seem to be widely known or discussed - the only direct discussion I found for it was here:

https://javascript.info/clickjacking#samesite-cookie-attribute

There's a brief mention of the fact that `SameSite` cookies won't be sent for requests in `<iframes>` on the NetSparker blog, but they don't talk at all about using this as a Clickjacking defence:

https://www.netsparker.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/

I've played around with this myself in modern browsers and it seems to be effective, although I've not had a really deep dig into it to see if there's any way to bypass this.

I'd welcome any thoughts or comments on this.

Also fixed up a few minor things like trailing whitespace while I was at it.


righettod:
Interesting discussion indeed, give us some days to look at the topic and to come back to you 😃

righettod:
@rbsec Can you share your prototype with us?

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -77,6 +77,20 @@ Meta-tags that attempt to apply the X-Frame-Options directive DO NOT WORK. For e
 - **X-Frame-Options Deprecated** While the X-Frame-Options header is supported by the major browsers, it was never standardized and has been deprecated in favour of the frame-ancestors directive from the CSP Level 2 specification.
 - **Proxies** Web proxies are notorious for adding and stripping headers. If a web proxy strips the X-Frame-Options header then the site loses its framing protection.
 
+# Defending with SameSite Cookies
+
+The `SameSite` cookie attribute defined in [RFC 6265bis](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7) is primarily intended to defend against [cross-site request forgery (CSRF)](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md#samesite-cookie-attribute); however it can also provide protection against Clickjacking attacks.
+
+Cookies with a `SameSite` attribute of either `strict` or `lax` will not be included in requests made to a page within an `<iframe>`. This means that if the session cookies are marked as `SameSite`, any Clickjacking attack that requires the victim to be authenticated will not work, as the cookie will not be sent.
```
righettod: Can be interesting to add the 2 links mentioned in the description of the PR because they provide good explanation.

rbsec: Added.

rbsec:
@righettod unfortunately it was on a web application belonging to a client which wasn't externally facing, but I can throw together a quick PoC tonight.

rbsec:
For a proof of concept:

This page will set three cookies (called "none", "lax" and "strict") with the respective SameSite policies:

http://rbsec.net/setcookies.php

Once you've browsed there to create them, this page will print all your cookies (note that if you click the link below rather than browsing directly, you won't get the strict one):

http://rbsec.net/getcookies.php

If you browse to it directly, you should see all three cookies. However, if you try and load it in an `<iframe>`, you should only see the "none" cookie.

```html
<iframe src="http://rbsec.net/getcookies.php" />
```

Since the cookies don't load in the frame, it would only be possible to do anonymous clickjacking against the site.

Tested and working for me in Firefox, Chrome and Edge (although my Edge shows no cookies in an `<iframe>`, possibly because I have the non-default setting to block third party cookies).

righettod:
Thanks a lot for the POC, I will take a look ASAP and come back to you 😃

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -2,7 +2,7 @@
 
 This cheat sheet is focused on providing developer guidance on [Clickjack/UI Redress](https://www.owasp.org/index.php/Clickjacking) attack prevention.
 
-The most popular way to defend against Clickjacking is to include some sort of "frame-breaking" functionality which prevents other web pages from framing the site you wish to defend. This cheat sheet will discuss two methods of implementing frame-breaking: first is X-Frame-Options headers (used if the browser supports the functionality); and second is javascript frame-breaking code.
+The most popular way to defend against Clickjacking is to include some sort of "frame-breaking" functionality which prevents other web pages from framing the site you wish to defend. This cheat sheet will discuss two methods of implementing frame-breaking: first is X-Frame-Options headers (used if the browser supports the functionality); and second is JavaScript frame-breaking code.
```
ThunderSon: - Add a link just like this [X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) header; it is a singular header, and not multiple ones.
- and `the` second is ... Add `the`
- This introduction looks to be old, and I believe could be updated. Could you possible open up an issue about this in order to handle it? 😄

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -77,6 +77,22 @@ Meta-tags that attempt to apply the X-Frame-Options directive DO NOT WORK. For e
...
+
+Cookies with a `SameSite` attribute of either `strict` or `lax` will not be included in requests made to a page within an `<iframe>`. This means that if the session cookies are marked as `SameSite`, any Clickjacking attack that requires the victim to be authenticated will not work, as the cookie will not be sent. An article on the [Netsparker blog](https://www.netsparker.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/) provides further details on which types of requests cookies are sent for with the different SameSite policies.
+
+This approach is discussed on the [JavaScript.info website](https://javascript.info/clickjacking#samesite-cookie-attribute).
+
+## Limitations
+
+As of 2019 the `SameSite` attribute is supported by [most modern browsers](https://caniuse.com/#feat=same-site-cookie-attribute), although it it not widely supported on mobile browsers. Support in Internet Explorer will varies depending on the updates installed and the underlying operating system.
```
ThunderSon: This is not a limitation. It almost is supported by even all mobile browsers. Could you possibly restructure this sentence?
- will `vary` instead of `varies`

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -77,6 +77,22 @@ Meta-tags that attempt to apply the X-Frame-Options directive DO NOT WORK. For e
...
+
+Additionally, if the Clickjacking attack does not require that the user is authenticated, this attribute will not provide any degree of protection.
```
ThunderSon: replace `that the user is authenticated` with `the user to be authenticated`

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -77,6 +77,22 @@ Meta-tags that attempt to apply the X-Frame-Options directive DO NOT WORK. For e
...
+
+This use of this attribute should be considered as part of a defence-in-depth approach, and it should not be relied upon as the sole protection against Clickjacking.
```
ThunderSon: as the sole `protective measure`

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
rbsec: I've reworded the sentence to soften it a bit and removed the note about mobile browsers - but according the the caniuse stats there's currently 12% of users with browsers that don't support it, which is non-trivial.

Thoughts on the new wording?

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
rbsec: I hadn't really look at the intro in any depth, but you're right that it didn't read too well. I've had a stab at rewriting it in the latest commit.

cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -87,11 +93,11 @@ This approach is discussed on the [JavaScript.info website](https://javascript.i
 
 ## Limitations
 
-As of 2019 the `SameSite` attribute is supported by [most modern browsers](https://caniuse.com/#feat=same-site-cookie-attribute), although it it not widely supported on mobile browsers. Support in Internet Explorer will varies depending on the updates installed and the underlying operating system.
+If the Clickjacking attack does not require the user to be authenticated, this attribute will not provide any protection.
 
-Additionally, if the Clickjacking attack does not require that the user is authenticated, this attribute will not provide any degree of protection.
+Additionally, while `SameSite` attribute is supported by [most modern browsers](https://caniuse.com/#feat=same-site-cookie-attribute), there are still a significant number of users (approximately 12% as of August 2019) with browsers which do not support it.
```
ThunderSon: there `is` still ... and not are

rbsec: "number" is a plural here, so I believe "are" is correct:

https://www.grammar-monster.com/lessons/singular_plural_number.htm

victoriadrake: The use of “is” and “are” depends on whether the thing being referred to is singular or plural; the question in this case is whether we mean to say “a high number of users” (in which case, we use “is” to refer to the numerical value - see the link @rbsec provided) or if we mean to say “many users” (in which case, the sentence should be reworded).

I suggest the following options for revision:

> ... there is still a significant number of users...

> ... there are still many users...

On an unrelated note, the latter part of this sentence should read, “with browsers that do not support it.” See [_that_ vs _which_](https://www.quickanddirtytips.com/education/grammar/which-versus-that-0).

This has been a PSA from your friendly neighbourhood grammarist 👋

rbsec: @victoriadrake thanks - I've updated it to `many users` and also amended the final sentence.

righettod:
@victoriadrake what is PSA? (thanks in advance for your light 😃 )

victoriadrake:
> @victoriadrake what is PSA? (thanks in advance for your light 😃 )

Oops. “Public Service Announcement” - in jest.

PR #189, opened 2019-08-27
Fancy quotes

rbsec:
The vast majority of the cheatsheets use normal quotes `"` or `'`', but there are a number of instances where fancy quotes `“”` or `’‘` are used.

This seems to be about 5% of the cases from a quick grep:

```
CheatSheetSeries[master]$ grep '“|”' -or| wc -l
176
CheatSheetSeries[master]$ grep '"' -or| wc -l
4341
```

Most of the time this doesn't matter, but there are a few cases (like the OS command injection sheet) where fancy quotes are used in example code, which may or may not work depending on the interpreter. This PR replaces all the fancy quotes with normal ones (and also a couple of fancy hyphens as well).

This is mostly a stylistic/consistency thing, so feel free to reject and close this PR if you don't want to merge it. If you do merge, it might also be worth updating the style guidelines (or even adding a CI check for them).


righettod:
Hello, 
We want to avoid PR modifying many files because it become the hell to review and merge moreover when differentPR  are pending in the  pipe.
But thanks for the point, we will take care of it during the update of each PR...
Thank in advance for your understanding.

rbsec:
@righettod no problem, I thought that this kind of PR might cause some problems. Mass changes like this can often be a bit of a pain for others.

Thanks

righettod:
I have discussed with @ThunderSon and he will handle it. I'm sorry for the quick close.
In fact we try to avoid mass PR in order to be able to review PR quickly and give quick feedback to the contributor.
As it's spare time task, I take of the spare time of the core team 😃

rbsec:
No worries - this might be the kind of change that it's easier to implement yourselves rather than from a PR. I basically just did it with a find command, something like:

`find ./cheatsheets -name "*.md" -exec sed -i 's/”/"/g' {} \;`

righettod:
OMG, now that i have slept enough, I understand the replacement...
In fact, it's a easy and very useful update indeed.
Thanks a lot for this PR 👍

righettod:
@rbsec I have a flavor to ask to you, can you fix the conflict into the file **cheatsheets/Cryptographic_Storage_Cheat_Sheet.md** so we can merge the PR quickly?
Thank you very much in advance 😃 
I will create a ticket to update the config of the linter to prevent future usage of fancy quote...

rbsec:
@righettod Should be good to merge now.

ThunderSon:
@rbsec Thank you for this update. Can you share the second command u used that the first one missed? Might you have Slack as well? 😊 Would make discussions a bit easier since you're a seasoned contributor.

rbsec:
@ThunderSon it was the same command, but I also include the `cheatsheets_excluded` folder as well (although there was only one in there).

I don't have Slack at the moment, but I can certainly setup an account.

ThunderSon:
Not bothered by checking the excluded folder, but it's just more work you did. That CS won't be included in our final version most probably, as discussed in a previous issue. Reviewing the PR right now.

PR #191, opened 2019-08-27
Unvalidated forwards

rbsec:
Some of the points in the section about safe redirects talk about validation, but don't really give much guidance on how to do this. There are some really common mistakes that people make here (such as allowing redirects to URLs starting with a `/` assuming that means the same site), so I added some notes for this.

Also updated the safe PHP example to call `exit` after the redirect (to avoid access control issues), and a bit of minor cleanup.


cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.md
```diff
@@ -157,15 +159,24 @@ public class ForwardServlet extends HttpServlet
...
-- Sanitize input by creating a list of trusted URL's (lists of hosts or a regex).
+- Sanitize input by creating a list of trusted URLs (lists of hosts or a regex).
 - Force all redirects to first go through a page notifying users that they are going off of your site, and have them click a link to confirm.
 
+## Validating URLs
+When attempting to validate and sanitise user-input to determine whether the URL is safe, wherever possible you should use a built in library or function to parse the URLs, such as `parse_url()` in PHP, rather than rolling your own parser using regex. Additionally, make sure that you take the following into account:
+
+- Input starting with a `/` to redirect to local pages is **not safe**. `//example.org` is a valid URL.
+- Input starting with the desired domain name is **not safe**. `https://example.org.attacker.com` is valid.
+- JavaScript URIs such as `javascript:alert(1)` should be blocked
```
righettod: I will advise to only allow the HTTP/HTTPS protocol to be sure that no other protocol are used like Javascript/Ftp/...

cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.md
```diff
@@ -157,15 +159,24 @@ public class ForwardServlet extends HttpServlet
 Safe use of redirects and forwards can be done in a number of ways:
 
 - Simply avoid using redirects and forwards.
-- If used, do not allow the url as user input for the destination. This can usually be done. In this case, you should have a method to validate URL.
+- If used, do not allow the URL as user input for the destination. This can usually be done. In this case, you should have a method to validate URL.
 - If user input can’t be avoided, ensure that the supplied **value** is valid, appropriate for the application, and is **authorized** for the user.
 - It is recommended that any such destination input be mapped to a value, rather than the actual URL or portion of the URL, and that server side code translate this value to the target URL.
-- Sanitize input by creating a list of trusted URL's (lists of hosts or a regex).
+- Sanitize input by creating a list of trusted URLs (lists of hosts or a regex).
```
righettod: I will say to made an emphasis on the point above **It is recommended that any such destination input be mapped to a value, rather than the actual URL or portion of the URL, and that server side code translate this value to the target URL.** because, for me, it's the solution to adopt due to the fact that the parsing of URL is prone to bypass depending on the technology used for the implementation like described in this [presentation](https://www.blackhat.com/docs/us-17/thursday/us-17-Tsai-A-New-Era-Of-SSRF-Exploiting-URL-Parser-In-Trending-Programming-Languages.pdf).

cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.md
rbsec: Added this as a point.

cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.md
rbsec: I've updated a few of the points in this section to emphasise the use of IDs/Tokens (and a potential issue with this approach). Does it look reasonable to you?

righettod:
Thank you very much for the update, can you fix the linter issue please ?
Thanks a lot in advance 😃

rbsec:
@righettod whoops, my bad. Fixed the linter error and also the merge conflict from the fancy quotes PR.

cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.md
righettod: Yes thanks a lot

Issue #99, opened 2019-05-06
CS Credential Stuffing Grammar Update


ThunderSon:
- What update/refactoring do you want to perform? 

Grammar update

righettod:
Hello, 

As the issue is assigned since many time and we do not have received any PR for this then we send it back to the backlog and we set it as **HELP_WANTED**.

Thank you very much in advance for your understanding 😃

righettod:
Handling by @rbsec pending.

Issue #96, opened 2019-05-04
Multi-Factor Authentication CS


mackowski:
List of common factors (SMS, e-mail, application, hardware tokens etc) that can be used for Multi-Factor Authentication with comparison.
Example implementation of Multi-Factor Authentication.

righettod:
Hello, 

As the issue is assigned since many time and we do not have received any PR for this then we send it back to the backlog and we set it as **HELP_WANTED**.

Thank you very much in advance for your understanding 😃

rbsec:
I've made a start on a skeleton for this here:

https://github.com/rbsec/CheatSheetSeries/blob/mfa/cheatsheets/Multifactor_Authentication_Cheat_Sheet.md

Any comments or feedback are welcome.

rbsec:
Draft CS submitted for review.

PR #194, opened 2019-08-28
Credential stuffing

rbsec:
This is an initial attempt at cleaning up the credential stuffing cheatsheet, as per issue #99.

It's currently a work in progress and not ready for merging (as you can probably see I'm only about half way through the cheatsheet), but I thought I'd open a draft PR in case you have any feedback, or there are specific parts of the cheatsheet you think need extra attention.

The main technical content isn't changing - I'm mostly focusing on rewriting it to read better, although I have increased the emphasis on implementing MFA since the original as it's by far the best defence that can be implemented.

I'll try and finish this in the next few days, but any comments you have in the meantime are welcome.


righettod:
Very good idea, i was not aware that it was possible to create a "draft" PR thanks a lot for the hint 😃 

@mackowski This topic is one that interest you a lot no?

righettod:
I affect the issue #99 to you @rbsec

mackowski:
Thanks @rbsec! I will review it with pleasure :) 

There is not a lot about MFA because we wanted to create a new cheatsheet #96 that will be about MFA and here we will add a reference to that CS.
Purpose of it is to decouple these two topics. If we do this in 2 CSs it will be easier to find and it is also more aligned with our vision of this project where CS is short and concrete. 

What you already done looks fine! I like it!

rbsec:
@mackowski that makes sense - I wasn't planning to go into any real detail about implementing MFA, more just to emphasise that they should be using it.

What do you think about expanding the scope of this cheatsheet a bit to also cover password spraying attacks? It's a distinct, different type of attack, but pretty much all of the defences here would be effective in protecting against it. I feel like if I was going to create a separate cheatsheet for that, it would be almost identical to this so, which suggests they should probably be combined.

I was thinking about splitting the introduction into two parts (password spraying and credential stuffing), and then the rest of the cheatsheet would be more or less the same as it is, with with a few comments about where certain controls would only protect against one type of attack.

mackowski:
@rbsec it makes sense to cover password spraying attacks here. 

One challange I see here is that we try not to change names os cheatsheets in order to not break external links. Therefore I try to think how to make it discoverable for others that content about password spraying attacks is here. 

@righettod what do you think? Can we change the name of the CS since it was created in V2 and is still pretty new or do you see other solution for it?

rbsec:
@mackowski @righettod first draft is ready for review. Main technical changes are:

* Increase emphasis on MFA
* Add secondary authentication (pins, etc)
* Add password spraying

Looking forward to your feedback.

righettod:
@rbsec @mackowski Regarding the renaming of the credential stuffing CS, cross reference are crucial so I prefer to add more extra content to the credential stuffing CS and keep the name.

ThunderSon:
I am looking the usage with `number`. I'll get back to you with some feedback. No need to reply to every requested change about it. English ...

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
```diff
@@ -1,84 +1,112 @@
...
+This cheatsheet covers defences against two common types of authentication-related attacks: credential stuffing and password spraying. Although these are separate, distinct attacks, in many cases the defences that would be implemented to protect against them are the same, and they would also be effective at protecting against brute-force attacks. A summary of these different attacks is listed below:
 
-This is a subset of the brute force attack category: large numbers of spilled credentials are automatically entered into websites looking for matches to existing accounts, which the attacker can then hijack for their own purposes.
+| Attack Type | Description |
+|-------------|-------------|
+| Brute Force | Testing multiple passwords from dictionary or other source against a single account. |
+| Credential Stuffing | Testing username/password pairs obtained from the breach of another site. |
+| Password Spraying | Testing a single weak password against a large number of different accounts.
```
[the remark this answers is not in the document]
rbsec: Table was done by hand - I've fixed the typo. I've found that most markdown renderers are pretty lax about tables, you can usually have a row that's just `foo|bar` and they won't complain. No reason not to do it properly though.

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
```diff
@@ -1,84 +1,112 @@
...
 
-# Primary Defenses
+# Multi-Factor Authentication
 
-It should be noted that defense mechanisms are intended to be used in a layered approach. In most cases, a single defense option would be inadequate to stop most Credential Stuffing attacks.
+Multi-factor authentication (MFA) is by far the best defense against the majority of password-related attacking, including credential stuffing and password spraying, with analysis by Microsoft suggesting that it would have stopped [99.9% of account compromises](https://techcommunity.microsoft.com/t5/Azure-Active-Directory-Identity/Your-Pa-word-doesn-t-matter/ba-p/731984). As such, it should be implemented wherever possible; however depending on the audience of the application it may not be practical or feasible to enforce the use of MFA.
```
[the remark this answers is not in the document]
ThunderSon: @rbsec this was not resolved. (had to reopen it)

rbsec:
@ThunderSon there's quite a lot of discussion about it online, and both forms seem quite widely used. The general consensus I've found that "A number" is plural (and should have an "are"), whereas "The number" is singular (and would be is).

However ultimately this doesn't matter - the cheatsheets are meant to be best practices in security, not a lesson on the finer points of English grammar (which is horribly inconsistent anyway), so I've just slightly reworded both sentences to avoid the problem.

`..there are a number of alternative defenses` > `..there are many alternative defenses..`

Let's leave the arguments to the linguists...

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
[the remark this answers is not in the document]
rbsec: @ThunderSon this should be fixed in 1c092d9b6b775aa0af447bc0e05537afe42f2e87

righettod:
A small reminder: It's important to check the english but the technical content is the real attention point because our goal is to provide real/direct/actionable  useful content to dev team and unfortunately we have a huge legacy of CS...
So, it must never be a blocking point to merge a PR once the english is "correct". If needed, we can perform several PR on a CS to enhance the english over the time...

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
```diff
@@ -1,84 +1,114 @@
...
-In many cases, brute force protections will overlap with credential stuffing defenses.
+In order to balance security and usability, multi-factor authentication can be combined with other techniques to require for 2nd factor only in specific circumstances where there is reason to suspect that the login attempt may not be legitimate, such as a login from:
 
-Keep in mind that application can have different security levels for different users/roles or actions. For example for casual customers multi-factor authentication may be optional but for admins or some other special roles it should be enforced.
+- A new browser/device or IP address.
+- An unusual country or location.
+- Specific countries that are considered untrusted.
+- An IP address that appears on known blacklists.
+- An IP address that has tried to login to multiple accounts.
+- A login attempt that appears to be scripted rather than manual.
```
righettod: For specific business action, a second authentication factor might be needed like for example banking transaction on a new account, update a beneficiary...It can be interesting to add the point that business feature can lead to MFA...

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
```diff
@@ -1,84 +1,114 @@
...
-Similar to Multi-Factor Authentication this defence can be combined with other techniques to ask user to solve CAPTCHA only in special situations. 
+## Secondary Passwords, PINs and Security Questions
 
-Lots of CAPTCHA examples can be reviewed [here](https://www.whoishostingthis.com/resources/captcha/)
+As well as requiring a user to enter their password when authentication, they can also be prompted to provide additional security information such as:
 
-## Defense Option 3: IP blacklists
+- A PIN
+- Specific characters from a secondary passwords or memorable word
+- Answers to security questions
```
righettod: Can be interesting to add a link to the [Choosing and Using Security Questions Cheat Sheet](Choosing_and_Using_Security_Questions_Cheat_Sheet.md).

rbsec: Added in 03ebaabe3b903b0917ccef93f02b451329fb228a.

cheatsheets/Credential_Stuffing_Prevention_Cheat_Sheet.md
rbsec: I like the use of MFA on that kind of sensitive functionality (and also stuff like changing passwords or recovery email addresses). I'm not sure it really fits in too well here, but certainly something to include in the new MFA cheatsheet (#96).

righettod:
Thanks a lot.

righettod:
It was just a note but it is true that the list provided is enough. We can
as you said add more cases in the MFA CS.

rbsec:
@righettod I believe that all those changes should not be implemented.

righettod:
> @righettod I believe that all those changes should not be implemented.

I do not understand this message...

rbsec:
@righettod apologies, that should have read "..changes should **now** be implemented"

righettod:
@rbsec No problem, thanks for the clarification 😃

I have merged the PR, the content is great, thanks you so much for your help on the project. 

As you have seen, we have a huge legacy of CS to refresh but as the project is now fully on GitHub it will help...

PR #208, opened 2019-09-14
Add warning about enterprise proxies with certificate auth to Authentication CS

rbsec:
Client certificates and enterprise proxies don't play nicely together (you either have to install the cert on the proxy, which is bad, or whitelist the site so there's not SSL/TLS decryption).


righettod:
Thanks a lot for the PR 👍

PR #209, opened 2019-09-14
Auth CS: improve brute-force protection section

rbsec:
Main changes are:

* Widen scope to cover other automated attacks (linking to credential stuffing CS)
* Discuss some other defences (MFA, CAPTCHA, security questions/words)
* Provide some more guidance on a lockout policy

Do you think it would be useful to give examples of a sensible lockout policy (something like 5 attempts, 30 minute lockout, 30  minute observation window), and if so, what values should we suggest? This kind of thing should be environment and application specific, so I'm a bit hesitant to give actual numbers.

The MFA guidance is essentially just taken from the credential stuffing CS. Once we have the MFA cheatsheet (#96) it can be cut down and just link to that.

### DIFF
--- a/cheatsheets/Authentication_Cheat_Sheet.md
+++ b/cheatsheets/Authentication_Cheat_Sheet.md
@@ -181,15 +181,15 @@ There are a number of different types of automated attacks that attackers can us
 | Credential Stuffing | Testing username/password pairs obtained from the breach of another site. |
 | Password Spraying | Testing a single weak password against a large number of different accounts.|
 
-Various different protection mechanisms can be implemented to protect against them. In many cases these defences do not provide complete protection, but when a number of them are implemented in a defence-in-depth approach, a reasonable level of protection can be achieved.
+Different protection mechanisms can be implemented to protect against these attacks. In many cases these defences do not provide complete protection, but when a number of them are implemented in a defence-in-depth approach, a reasonable level of protection can be achieved.
 
-The follow sections will focus primarily on preventing brute-force attacks, although these controls can also be effective against other types of attacks. For a further guidance on defending against credential stuffing and password spraying, see the [Credential Stuffing Cheat Sheet](Credential_Stuffing_Prevention_Cheat_Sheet.md).
+The following sections will focus primarily on preventing brute-force attacks, although these controls can also be effective against other types of attacks. For further guidance on defending against credential stuffing and password spraying, see the [Credential Stuffing Cheat Sheet](Credential_Stuffing_Prevention_Cheat_Sheet.md).
 
 ### Multi-Factor Authentication
 
 Multi-factor authentication (MFA) is by far the best defense against the majority of password-related attacks, including brute-force attacks, with analysis by Microsoft suggesting that it would have stopped [99.9% of account compromises](https://techcommunity.microsoft.com/t5/Azure-Active-Directory-Identity/Your-Pa-word-doesn-t-matter/ba-p/731984). As such, it should be implemented wherever possible; however, depending on the audience of the application, it may not be practical or feasible to enforce the use of MFA.
 
-In order to balance security and usability, multi-factor authentication can be combined with other techniques to require for 2nd factor only in specific circumstances where there is reason to suspect that the login attempt may not be legitimate, such as a login from:
+In order to balance security and usability, multi-factor authentication can be combined with other techniques to require for 2nd factor when performing sensitive actions (such as adding a new payment destination), or in specific circumstances where there is reason to suspect that the login attempt may not be legitimate, such as a login from:
 
 - A new browser/device or IP address.
 - An unusual country or location.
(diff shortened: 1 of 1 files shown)


cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -171,11 +171,58 @@ The application may return a different [HTTP Error code](https://developer.mozil
...
+| Attack Type | Description |
+|-------------|-------------|
+| Brute Force | Testing multiple passwords from dictionary or other source against a single account. |
+| Credential Stuffing | Testing username/password pairs obtained from the breach of another site. |
+| Password Spraying | Testing a single weak password against a large number of different accounts.|
+
+Various different protection mechanisms can be implemented to protect against them. In many cases these defences do not provide complete protection, but when a number of them are implemented in a defence-in-depth approach, a reasonable level of protection can be achieved.
+
+The follow sections will focus primarily on preventing brute-force attacks, although these controls can also be effective against other types of attacks. For a further guidance on defending against credential stuffing and password spraying, see the [Credential Stuffing Cheat Sheet](Credential_Stuffing_Prevention_Cheat_Sheet.md).
```
ThunderSon: - `following` instead of `follow`
- What do you think about this? Just looking at what looks/sounds better:
   - against other automated attacks **vs** against other types of attacks
- `For further` instead of `For a further`

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -171,11 +171,58 @@ The application may return a different [HTTP Error code](https://developer.mozil
...
+
+### Multi-Factor Authentication
+
+Multi-factor authentication (MFA) is by far the best defense against the majority of password-related attacks, including brute-force attacks, with analysis by Microsoft suggesting that it would have stopped [99.9% of account compromises](https://techcommunity.microsoft.com/t5/Azure-Active-Directory-Identity/Your-Pa-word-doesn-t-matter/ba-p/731984). As such, it should be implemented wherever possible; however, depending on the audience of the application, it may not be practical or feasible to enforce the use of MFA.
+
+In order to balance security and usability, multi-factor authentication can be combined with other techniques to require for 2nd factor only in specific circumstances where there is reason to suspect that the login attempt may not be legitimate, such as a login from:
```
ThunderSon: How about discussing the need for MFA only as well when there is a special action being done?

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -171,11 +171,58 @@ The application may return a different [HTTP Error code](https://developer.mozil
...
+- An IP address that has tried to login to multiple accounts.
+- A login attempt that appears to be scripted rather than manual.
+
+Additionally, for enterprise applications, known trusted IP ranges could be added to a whitelist so that MFA is not required when users connect from these ranges.
+
+### Account Lockout
+
+The most common protection against these accounts is to implement account lockout, which prevents any more login attempts for a period after a certain number of failed logins.
+
+The counter of failed logins should be associated with the account itself, rather then the source IP address, in order to prevent an attacker making login attempts from a large number of different IP addresses. There are a number of different factors that should be considered when implementing an account lockout policy in order to find a balance between security and usability:
```
ThunderSon: rather `than`

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -171,11 +171,58 @@ The application may return a different [HTTP Error code](https://developer.mozil
...
+
+- The number of failed attempts before the account is locked out (lockout threshold).
+- The time period that these attempts must occur within (observation window).
+- How long the account is locked out for (lockout duration).
+
+Rather than implementing a fixed lockout duration (e.g, ten minutes), some applications use and exponential lockout, where the lockout duration starts as a very short period (e.g, one second), but doubles after each failed login attempt.
```
ThunderSon: - use `an` exponential
- using exponential and then doubles is kind of misleading?

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -171,11 +171,58 @@ The application may return a different [HTTP Error code](https://developer.mozil
...
+
+### CAPTCHA
+
+The use of an effective CAPTCHA can help to prevent automated login attempts against accounts. However, many CAPTCHA implementations have weaknesses that allow them to be solved using automated techniques, or can be outsourced to services which can solve them. As such, the use of CAPTCHA should be viewed as a defence-in-depth control to make brute-force attacks more time consuming and expensive, rather than as a preventative.
+
+It may be more user-friendly to only require a CAPTCHA be solved after a small number of failed login attempts, rather than requiring it from the very first login.
+
+### Security Questions and Memorable Words
+
+The additional of a security question or memorable word can also help protect against automated attacks, especially when the user is asked to enter a number of randomly chosen characters from the word. It should be noted that this does **not** constitute multi-factor authentication, as both factors are the same (something you know).
```
ThunderSon: - The `addition`.
- What do you think about adding the risk of having weak security questions?

cheatsheets/Authentication_Cheat_Sheet.md
rbsec: Added

cheatsheets/Authentication_Cheat_Sheet.md
rbsec: Added a link to the security questions CS

cheatsheets/Authentication_Cheat_Sheet.md
rbsec: Some of the defences would also help against non-automated attacks (such as MFA protecting against phishing/etc), so I think the current wording is a bit more general.

cheatsheets/Authentication_Cheat_Sheet.md
rbsec: The term `exponential backoff` seems to be fairly well used, and a timeout that doubles each time is definitely an exponential function (2^n). Would `exponential backoff` be better here?

ThunderSon: No no, I believe I was mistaken. Apologies.

Issue #207, opened 2019-09-13
Cheat sheet update/refactor proposal: XML External Entity (XXE) Prevention Cheat Sheet


Jurrie:
###  What is missing or needs to be updated in the current iteration of the cheat sheet?

Section Java -> JAXB Unmarshaller needs to be updated. It basically says that default Unmarshaller instances are unsafe for XXE injection. I believe that, since [JDK-8010393](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8010393) (which is in Java 8 beta 86), this is no longer true. I base this on [this commit](https://github.com/JetBrains/jdk8u_jaxws/commit/5a04cd1fb06c900cdc2be686e15aad26f4093a60#diff-295f96af589d485569513d34d9aec081R276). There, the property `com.sun.xml.internal.bind.v2.runtime.JAXBContextImpl.disableSecurityProcessing` is added, with a default of `true`.

Java 8 was released on 18-03-2014 and the bug was marked as resolved on 09-04-2013. So it's safe to assume it made it into the first public release of Java 8.

Also, I tested with OpenJDK 1.8.0_77, and I can not do XXE injection with a default Unmarshaller.

Looking at the source code of OpenJDK 1.7.0_231, I see the bugfix is not present. So Java 7 is not safe from XXE injection by default.

On a side note, I can reproduce this with OpenJDK 1.8.0_77 when I deliberately break security:
```
final Map<String, Object> properties = new HashMap<>();
properties.put(com.sun.xml.internal.bind.api.JAXBRIContext.DISABLE_XML_SECURITY, Boolean.TRUE);
JAXBContext.newInstance(new Class[] { Foo.class }, properties).createUnmarshaller();
```

### How do you intend to resolve this with an update/refactoring?
State that Java 8 and up give safe defaults for Unmarshaller.

All other mentioned classes are still vulnerable by default.

righettod:
Thanks for your proposal, it is very interesting.
Do you want to work on it?

Jurrie:
Sure :) I have made a pull request here: https://github.com/OWASP/CheatSheetSeries/pull/211

rbsec:
#211 has been merged, so closing this down. Thanks for the PR.

PR #211, opened 2019-09-16
Add note on Unmarshaller instances being safe by default on OpenJDK 8+

Jurrie:

This PR covers issue #207

### DIFF
--- a/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
@@ -310,6 +310,10 @@ Unmarshaller um = jc.createUnmarshaller();
 um.unmarshal(xmlSource);
 ```
 
+### Java 8 and up
+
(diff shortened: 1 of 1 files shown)


Jurrie:
Hello @ThunderSon. No, I do not think that line should be updated. It merely states that, like the Unmarshaller, the XPathExpression is unconfigurable. So, like the Unmarshaller, the untrusted data should be parsed through another parser first. That still is true, even with Java 8 and up.

Unlike the Unmarshaller, the XPathExpression is still vulnerable by default on Java 8. But that is noted in the added section on Unmarshaller:

> The other classes mentioned here are still unsafe by default in Java 8.

If you feel this should be clarified more, maybe you could suggest an addition? Thanks!

ThunderSon:
What I mean is the following:
In the past, before that JDK version, the developer was not able to secure `Unmarshaller`. After that update, it comes secure by default, and I believe the developer can modify it if they wish to.
Based on your answer on the above claim, if the developer can actually modify it, the line 319 doesn't actually fit in the CS since `Unmarshaller` is now configurable by itself to be secure.

Does this make any sense?

Jurrie:
I don't think I completely follow :) XPathExpression is a completely unrelated class and has little to do with Unmarshaller. But I do get that it causes confusion to reference Unmarshaller in the paragraph on XPathExpression. So I've removed that reference.

@ThunderSon: is this more clear?

PR #217, opened 2019-09-23
Update Clickjacking_Defense_Cheat_Sheet.md

ben-stock:
Bugfix: origins and *.domain.com may not have single ticks around them

### DIFF
--- a/cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
+++ b/cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
@@ -27,7 +27,7 @@ Common uses of CSP frame-ancestors:
 - `Content-Security-Policy: frame-ancestors 'self' *.somesite.com https://myfriend.site.com;`
     - This allows the current site, as well as any page on `somesite.com` (using any protocol), and only the page `myfriend.site.com`, using HTTPS only on the default port (443).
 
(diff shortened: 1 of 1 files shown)


cheatsheets/Clickjacking_Defense_Cheat_Sheet.md
```diff
@@ -24,10 +24,10 @@ Common uses of CSP frame-ancestors:
     - This prevents any domain from framing the content. This setting is recommended unless a specific need has been identified for framing.
 - `Content-Security-Policy: frame-ancestors 'self';`
     - This only allows the current site to frame the content.
-- `Content-Security-Policy: frame-ancestors 'self' '\*.somesite.com' 'https://myfriend.site.com';`
+- `Content-Security-Policy: frame-ancestors 'self' *.somesite.com https://myfriend.site.com;`
     - This allows the current site, as well as any page on `somesite.com` (using any protocol), and only the page `myfriend.site.com`, using HTTPS only on the default port (443).
 
-Note that the single quotes are required.
+Note that the single quotes are required around 'self' and 'none', but may not occur around other source expressions.
```
ThunderSon: What do you think about the following suggestion, since we want to showcase them as keywords, and that formatting does that better:
```suggestion
Note that the single quotes are required around `self` and `none`, but may not occur around other source expressions.
```

ThunderSon:
Thank you for your PR 😄

Issue #213, opened 2019-09-17
New cheat sheet proposal : Node.js Security Cheat Sheet


mleblebici:
1. Which security issues are commonly encountered when using the technology or technique described in the cheat sheet?
Node.js is gaining popularity rapidly as a back end language. Like any other programming language or framework, it may be susceptible to various web application vulnerabilities.

2. What is the objective of the cheat sheet?
CS aims to help Node.js developers in developing secure Node.js applications.

3. What will the cheat sheet bring to the reader?
Node.js developers can benefit from this CS by learning about some tools, techniques and best practices on secure programming of Node.js applications.

mackowski:
looks fine for me :) 
@mleblebici  Do you want to create this Cheat Sheet?

mleblebici:
@mackowski I have created a new PR #219 for adding the cheat sheet.

ThunderSon:
Closed in #219

PR #219, opened 2019-09-23
New cheat sheet: Node.js Security Cheat Sheet

mleblebici:
Added Node.js Security Cheat Sheet.

This PR covers issue #213 .


cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
...ions are one of the strongest features of Node.js. However, increasing layers of nesting within callback functions can become a problem. Any multistage process can become nested 10 or more levels deep. This problem is called as Pyramid of Doom or Callback Hell. In such a code, the errors and results get lost within the callback. Promises are a good way to write asynchronous code without getting into nested pyramids. Promises provide top-down execution while being asynchronous by delivering errors and results to next .then function. Another advantage of Promises is the way Promises handle the errors. If an error occurs in a Promise class, it skips over the .then functions and invokes the first .catch function it finds. This way Promises bring a higher assurance of capturing and handling errors. As a principle, you can make all your asynchronous code(apart from emitters) return promises. However, it should be noted that Promise calls can also become a pyramid. In order to completely stay away from callback hells, flat Promise chains should be used. If the module you are using does not support Promises, you can convert base object to a Promise by using Promise.promisifyAll() function.
```
rbsec: Love `Pyramid of Doom` as a name.

This paragraph is quite long and dense - could it be split?

Backtick the `.catch` and other functions

Space after the bracket after `asynchronous code`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Return sanitized user objects
+
+Information about the users of an application is among the most critical information about the application. Therefore, user information should be stored on a different database. User tables generally include fields like id, username, full name, email address, birth date, credit card number and in some cases social security numbers. Therefore, when querying and using user objects, you need to return only needed fields as it may be vulnerable to personal information disclosure. This is also correct for other objects stored on the database. If you just need a certain field of an object, you should never return it with all of its fields. As an example, you can use a function like the following whenever you need to get information on a user. By doing so, you can only return the fields that are needed for your specific operation. In other words, if you only need to list names of the users available, you are not returning their email addresses or credit card numbers in addition to their full names.
```
rbsec: Not sure what you mean by `stored on a different database`?

`credit card number should never be stored in a users table. Maybe `password` instead?

`you should only return the specific fields required` rather than `you should never return it with all the fields`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+    id: user.id,
+    username: user.username,
+    fullName: user.fullName
+  };
+};
+```
+
+## Set request size limits
+
+Buffering and parsing of request bodies can be cumbersome for the server. If there is no limit on the size of requests, attackers can send request with large request bodies so that they can exhaust server memory or fill disk space. However, fixing a request size limit for all requests may not be the correct behavior, since some requests like those for uploading a file to the server have more content to carry on the request body. Also, input with a JSON type is more dangerous than a multipart input, since parsing JSON is a blocking operation. Therefore, you should set request size limits for different content types. You can accomplish this very easily with express middlewares as follows:
```
rbsec: `resource intensive` rather than `cumbersome`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+app.use(express.urlencoded({ limit: “1kb” }));
+app.use(express.json({ limit: “1kb” }));
+app.use(express.multipart({ limit:”10mb” }));
```
rbsec: Doesn't this mean that an attacker can just set their content type to a multipart form and bypass this?

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+app.use(express.limit(“5kb”)); // this will be valid for every other content type
+```
+
+## Use strict mode
+
+JavaScript has some unsafe features that lies within the language itself. In order to remove these features, ES5 included a strict mode for developers. With this mode, errors that were silent previously are thrown. It also restricted optimizations for JavaScript engines. With strict mode, previously accepted bad syntax causes real errors. use "use strict". Because of these improvements, you should always use strict mode in your application. In order to enable strict mode, you just need to write _‘use strict’;_ on top of your code.
```
rbsec: Replace first sentence with something like `JavaScript has a number of unsafe and dangerous legacy features that should not be used.`

`use 'use strict'` isn't really a complete sentence

Put it into a JavaScript code block like the other examples?

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Use object property descriptors
+
+Object properties include 3 hidden attributes: writable (if false, property value cannot be changed), enumerable (if false, property cannot be used in for loops) and configurable (if false, property cannot be deleted). When defining an object property through assignment, these three hidden attributes are set to true by default. These properties can be set as follows:
```
rbsec: Might look better having the three attributes in a table? Or if not, backtick them

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+    configurable: true,
+    value: “A”
+});
+```
+
+Apart from these, there are some special functions for object attributes. Object.preventExtensions() prevents new properties from being added to the object.
+
+## Handle errors in asynchronous calls
+
+Errors that occur within asynchronous callbacks can be easily forgettable. Therefore, as a general principle first argument to the asynchronous calls should be an Error object. Also, express routes handle errors itself, but it should be always remembered that errors occurred in asynchronous calls made within express routes are not handled, unless an Error object is sent as a first argument.
```
rbsec: `are easy to miss` rather than `are easily forgettable`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+Errors in these callbacks can be propagated as many times as possible. Each callback that the error has been propagated to can ignore, handle or propagate the error.
+
+## Listen to errors when using EventEmitter
+
+When using EventEmitter, errors can occur anywhere in the event chain. Normally, if an error occurs in an EventEmitter object, an error event with an Error object as its argument is called. However, if there are no attached listeners to that error event, the Error object that is sent as argument is thrown and becomes an uncaught exception. In short, if you do not handle errors within an EventEmitter object properly, these unhandled errors may crash your application. Therefore, you should always listen to error events when using EventEmitter objects.
```
rbsec: Second sentence is quite hard to read.

`as an argument` instead of `as argument`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+}
+var emitter = new myEventEmitter();
+emitter.on(‘error’, function(err){
+    //Perform necessary error handling here
+});
+```
+
+## Handle uncaughtException
+
+Node.js behavior for uncaught exceptions is to print current stack trace and then terminate the thread. However, Node.js allows customization of this behavior. It provides a global object named process which is available to all Node.js applications. It is an EventEmitter object  and in case of an uncaught exception, “uncaughtException” event gets emitted and it is brought up to the main event loop. In order to provide a custom behavior for uncaught exceptions, you can bind to this event. However, resuming the application after such an uncaught exception can lead to further problems. Therefore, if you do not want to miss any uncaught exception, you should bind to uncaughtException event and cleanup any allocated resources like file descriptors, handles and similar before shutting down the process. Resuming the application is strongly discouraged as the application will be in an unknown state.
```
rbsec: Also good not to print a stacktrace to the users - as it leaks information

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+process.on(“uncaughtException”, function(err) {
+    // clean up allocated resources
+    // log necessary error details to log files
+    process.exit(); // exit the process to avoid unknown state
+});
+```
+
+## Monitor the event loop
+
+When your application server is under heavy network traffic, it may not be able to serve its users. This is essentially a type of Denial of Service (DoS) attack. Toobusy module allows you to monitor the event loop. It keeps track of lags and when it goes beyond a certain threshold, this module can indicate your server is too busy. In that case, you can stop processing incoming requests and send them 503 Server Too Busy message so that your application stay responsive. Sample use of toobusy module is shown here:
```
rbsec: `_The_ Toobusy module`

`keeps track of the response time, and when`...

`Example` rather than `Sample`

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+        res.send(503, “Server Too Busy”);
+    } else {
+    next();
+    }
+});
+```
+
+## Perform application activity logging
+
+Logging application activity is an encouraged good practice. It makes it easier to debug any errors encountered during application runtime. It is also useful for security concerns, since it can be used during incident response. Also, these logs can be used to feed Intrusion Detection/Prevention Systems (IDS/IPS). In Node.js, there are some modules like Winston or Bunyan to perform application activity logging. These modules enable streaming and querying logs. Also, they provide a way to handle uncaught exceptions. With the following code, you can log application activities in both console and a desired log file.
```
rbsec: Link to logging cheatsheet?

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+});
+```
+
+Also, you can provide different transports so that you can save errors to a separate log file and general application logs to a different log file.
+
+## Perform input validation
+
+Input validation is a crucial part of application security. Input validation failures can result in many different types of application attacks. These include SQL Injection, Cross-Site Scripting, Command Injection, Local/Remote File Inclusion, Denial of Service, Directory Traversal, LDAP Injection and many other injection attacks. In order to avoid these attacks, input to your application should be sanitized first. The best input validation technique is to use a white list of accepted inputs. However, if this is not possible, input should be first checked against expected input scheme and dangerous inputs should be escaped. In order to ease input validation in Node.js applications, there are some modules like validator and mongo-express-sanitize. Also, you should escape all HTML and JavaScript content shown to users via application. You can use [escape-html](https://github.com/component/escape-html) or [node-esapi](https://github.com/ESAPI/node-esapi) libraries to perform output escaping.
```
rbsec: Link to input validation cheatsheet.

Should output encoding be its own section?

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Adhere to general application security principles
+
+This list has mainly focused on issues that are common in Node.js applications. Also, recommendations against these issues are given specific to Node.js environment. Apart from these, there are general principles that apply to web applications regardless of technologies used in application server. You should also keep those principles in mind while developing your applications. A very good reference document on these principles is developed and maintained by OWASP. You can always refer to [OWASP Web Application Security Testing Cheat Sheet](https://www.owasp.org/index.php/Web_Application_Security_Testing_Cheat_Sheet) to learn about vulnerabilities that may exist in your application.
```
rbsec: The cheatsheets on the wiki are deprecated, should point here instead: https://cheatsheetseries.owasp.org/

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+# Objective
+
+This cheat sheet aims to provide a list of best practices to follow during development of Node.js applications. [Node Security Platform (nsp)](https://github.com/nodesecurity/nsp) provides a CLI tool to identify packages that have known vulnerabilities. You can install it easily with the following command:
+
+```bash
+npm install -g nsp
+```
+
+Then you can check your projects by going to the directory where your package.json file is and executing nsp check command. In addition to nsp, you can also use [Snyk](https://snyk.io) to keep your packages safe.
```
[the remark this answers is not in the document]
mleblebici: Suggesting OWASP Dependency Check and Retire.js only

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Keep your packages up-to-date
+
+Security of your application depends directly on how secure the third-party packages you use in your application are. Therefore, it is important to keep your packages up-to-date.
+
+## Use appropriate security headers
+
+There are several different HTTP headers that can help you prevent some common attack vectors. These are listed below:
+
+* __Strict-Transport-Security__: HTTP Strict Transport Security (HSTS) dictates browsers that the application can only be accessed via HTTPS connections. This header takes two parameters: max-age to determine how long this configuration will be valid and includeSubDomains to state if subdomains are to be treated in the same way. In order to use it in your application, add the following codes:
```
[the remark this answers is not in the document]
mleblebici: Added reference to HSTS cheat sheet

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+app.use(helmet.hsts()); // default configuration
+app.use(helmet.hsts("<max-age>", "<includeSubdomains>")); // custom configuration
+```
+
+* __X-Frame-Options:__ It determines if a page can be loaded via a \<frame> or an \<iframe> element. Allowing to do so may result in clickjacking attacks which aims to deceive users to click on something different than they perceive to be clicking on. This header has 3 directives: DENY to never allow framing, SAMEORIGIN to only allow framing within the same origin and ALLOW-FROM to only allow framing from specified URIs. These behaviors can be achieved with helmet module as follows:
```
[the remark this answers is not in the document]
mleblebici: Changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+app.use(hemlet.xframe()); // default behavior (DENY)
+helmet.xframe(‘sameorigin’); // SAMEORIGIN
+helmet.xframe(‘allow-from’, ‘http://alloweduri.com’); //ALLOW-FROM uri
+```
+
+* __X-XSS-Protection:__ This header enables browsers to stop loading pages when browsers detect reflected cross-site scripting attacks. In order to implement this header in your application, you can use the following code:
```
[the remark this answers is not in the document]
mleblebici: added reference XSS prevention cheat sheet

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+   directives: {
+       scriptSrc: ["'self'", "'unsafe-inline'"],   // helps prevent XSS attacks
+       frame-ancestors: ["'none'"],  // helps prevent Clickjacking attacks
+       img-src: ["'self'", "'http://imgexample.com'"],
+       style-src: ["'none'"]
+    }
+}))
+```
+
+* __Cache-Control and Pragma:__ Cache-Control header can be used to prevent browsers from caching the given responses. It is desired for pages which contains sensitive information about either the user or the application. However, when implemented, it can raise serious performance issues. Therefore, the choice to use it or not should be considered thoroughly. It can be used easily by the following code:
```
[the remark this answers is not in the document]
mleblebici: modified the related part

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+app.use(helmet.noCache());
+```
+
+The above code sets Cache-Control, Surrogate-Control, Pragma and Expires headers accordingly.
+
+* __x-Download-Options:__ This header prevents Internet Explorer from executing downloaded files in the site’s context. This is achieved with noopen directive. You can do so with the following piece of code:
```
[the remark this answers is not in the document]
mleblebici: corrected the typo

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+app.use(helmet.ieNoOpen());
+```
+
+* __Expect-CT:__ Certificate Transparency is a new mechanism developed to fix some structural problems regarding current SSL infrastructure. It has three directives. Enforce directive dictates if the policy should be enforced or be used in report-only mode. Max-age directive specifies how long this setting will be valid. Report-uri directive specifies where the browser should send invalid CT information reports. These can be implemented in your application as follows:
```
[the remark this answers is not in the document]
mleblebici: edited accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+var expectCt = require(‘expect-ct’);
+app.use(expectCt({ maxAge: 123 }));
+app.use(expectCt({ enforce: true, maxAge: 123 }));
+app.use(expectCt({ enforce: true, maxAge: 123, reportUri: ‘http://example.com’}));
+```
+
+* __Public-Key-Pins:__ This header increases the security of HTTPS. With this header, a specific cryptographic public key is associated with a specific web server. If the server does not use the pinned keys in future, the browser regards the responses as illegitimate. It has 2 optional (reportUri, includeSubDomains) and 2 required (pin-sha256, max-age) directives. These can be used as follows:
```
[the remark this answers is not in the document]
mleblebici: added the warning and the reference regarding HPKP

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+```JavaScript
+app.use(helmet.hpkp({
+    maxAge: 123,
+    sha256s: [‘Ab3Ef123=’, ‘ZyxawuV45=’],
+    reportUri: ‘http://example.com’,
+    includeSubDomains: true
+}));
+```
+
+* __X-Powered-By:__ X-Powered-By header is used to inform what technology is used in the server side. This is an unnecessary header causing information leakage, so it should be removed from your application. To do so, you can use the hidePoweredBy as follows:
```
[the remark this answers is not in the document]
mleblebici: done

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+```JavaScript
+app.use(helmet.hsts()); // default configuration
+app.use(helmet.hsts("<max-age>", "<includeSubdomains>")); // custom configuration
+```
+
+* __X-Frame-Options:__ It determines if a page can be loaded via a \<frame> or an \<iframe> element. Allowing to do so may result in clickjacking attacks which aims to deceive users to click on something different than they perceive to be clicking on. This header has 3 directives: DENY to never allow framing, SAMEORIGIN to only allow framing within the same origin and ALLOW-FROM to only allow framing from specified URIs. These behaviors can be achieved with helmet module as follows:
+
+```JavaScript
+app.use(hemlet.xframe()); // default behavior (DENY)
+helmet.xframe(‘sameorigin’); // SAMEORIGIN
```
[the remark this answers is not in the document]
mleblebici: fixed fancy quotes

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+// route to protect
+app.post(“/login”, bouncer.block, function(req, res) {
+    if (LoginFailed){  }
+    else {
+        bouncer.reset( req );
+    }
+});
+```
+
+Apart from express-bouncer modules, there are several other modules that aims to mitigate brute-forcing. Ratelimiter and express-brute are examples of these modules.
```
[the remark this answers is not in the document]
mleblebici: modified accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+app.use(helmet.hidePoweredBy());
+```
+
+Also, you can lie about the technologies used with this header. For example, even if your application does not use PHP, you can set X-Powered-By header to seem so.
+
+```JavaScript
+app.use(helmet.hidePoweredBy({ setTo: ‘PHP 4.2.0’ }));
+```
+
+## Take precautions against brute-forcing
```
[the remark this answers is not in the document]
mleblebici: added references

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+var svgCaptcha = require(‘svg-captcha’);
+app.get(‘/captcha’, function (req, res) {
+    var captcha = svgCaptcha.create();
+    req.session.captcha = captcha.text;
+    res.type(‘svg’);
+    res.status(200).send(captcha.data);
+});
+```
+
+Also, account lockout is a recommended solution to keep attackers away from your valid users. Account lockout is possible with many modules like mongoose. You can refer to [this blog post](http://devsmash.com/blog/implementing-max-login-attempts-with-mongoose) to see how account locking is implemented in mongoose.
```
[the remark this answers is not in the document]
mleblebici: added reference

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Set cookie flags appropriately
+
+Generally, session information is sent over cookies in web applications. However, the usage of cookies can eliminate some attack vectors related to session management. There are some flags that can be set for each cookie. For session cookies, httpOnly, Secure and SameSite flags are very important. httpOnly flag prevents the cookie from being accessed by client-side JavaScript. This is an effective counter-measure for XSS attacks. Secure flag lets the cookie to be sent only if the communication is over HTTPS. SameSite flag can prevent cookies from being sent in cross-site requests which helps protect against Cross-Site Request Forgery (CSRF) attacks. Apart from these, there are other flags like domain, path and expires. Setting these flags appropriately is encouraged, but they are mostly related to cookie scope not the cookie security. Sample usage of these flags is given in the following example:
```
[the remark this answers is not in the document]
mleblebici: modified accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+app.use(session({
+    secret: ‘your-secret-key’,
+    key: ‘cookieName’,
+    cookie: { secure: true, httpOnly: true, path: ‘/user’, sameSite: true}
+}));
+```
+
+## Use Anti-CSRF tokens
+
+Cross-Site Request Forgery (CSRF) aims to perform authorized actions on behalf of an authenticated user, while the user is unaware of this action. CSRF attacks are generally performed for state-changing requests like password change, adding users or placing orders. Csurf is an express middleware that can be used to mitigate CSRF attacks. It can be used as follows:
```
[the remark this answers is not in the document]
mleblebici: added CSRF CS reference

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+    res.render(‘send’, { csrfToken: req.csrfToken() })
+})
+app.post(‘/process’, parseForm, csrfProtection, function(req, res) {
+    res.send(‘data is being processed’);
+});
+```
+
+After writing this code, you also need to add csrfToken to your HTML form, which can be easily done as follows:
+
+```JavaScript
```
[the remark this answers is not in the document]
mleblebici: corrected :)

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+<input type=”hidden” name=”_csrf” value=”{{ csrfToken }}”>
+```
+
+## Do not use dangerous functions
+
+There are some JavaScript functions that are too dangerous to use. To the fullest possible extent, use of such functions and modules should be avoided. The first example is the eval() function. This function takes a string argument and executes it as any other JavaScript source code. This behavior inherently leads to remote code execution vulnerability. Similarly, calls to child_process.exec are also very dangerous. This function acts as a bash interpreter and sends its arguments to /bin/sh. By injecting input to this function, attackers can execute arbitrary commands on the server. Therefore, its use is highly discouraged.
```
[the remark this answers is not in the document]
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+In addition to these functions, there are some modules that require special attention when being used. As an example, fs module handles filesystem operations. However, if improperly sanitized user input is fed into this module, your server’s content can be tampered. Similarly, vm module provides APIs for compiling and running code within V8 Virtual Machine contexts. Since it can perform dangerous actions by nature, it should be used within a sandbox.
```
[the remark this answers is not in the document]
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Stay away from evil regexes
+
+Denial of Service (DoS) attack aims to make one or more of an application’s resources or services unavailable for its legitimate users. Some Regular Expression (Regex) implementations cause extreme situations that makes the application very slow. Attackers can use such regex implementations to cause application to get into these extreme situations and hang for a long time.  Such regexes are called evil if application can be stuck on crafted input.  Generally, these regexes are exploited by grouping with repetition and alternation with overlapping. (a+)+, (a|a?)+ are some examples of evil regexes. Fortunately, there is a Node.js module that can be used to check if a specific regex is evil or not. However, as it is stated in the module’s Github page, you cannot “be absolutely sure that this module will catch all exponential-time cases”. Its usage is as simple as follows:
```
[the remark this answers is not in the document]
mleblebici: backticked the regex codes

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+```JavaScript
+node safe.js <regex>
```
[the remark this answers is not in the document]
mleblebici: for improved accuracy, vuln-regex-detector is recommended

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+node safe.js '(x+x+)+y'
+```
+
+## Remove unnecessary routes
+
+A web application should not contain any page that is not used by users. Leaving such pages on the website can bring advantage to attackers. Such pages may increase the attack surface of the application. This principle is also valid for Node.js applications. All unused API routes should be disabled in Node.js applications. This occurs especially in frameworks like Sails and Feathers, as they automatically generate REST API endpoints. For example, in Sails, if a URL does not match a custom route, it may match one of the automatic routes and still generate a response. This situation may lead to results ranging from information leakage to arbitrary command execution. Therefore, before using such frameworks and modules, it is important to know the routes they automatically generate and remove or disable these routes.
```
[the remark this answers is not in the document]
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+  // perform actions on file content
+});
+fs.unlinkSync(‘/file.txt’);
+```
+
+In the above example, unlinkSync function may run before the callback, which will delete the file before the desired actions on the file content is done. Such race conditions can also impact the security of your application. An example would be a scenario where authentication is performed in callback and authenticated actions are run synchronously. In order to eliminate such race conditions, you can write all operations that rely on each other in a single non-blocking function. By doing so, you can guarantee that all operations are executed in the correct order.
+
+## Prevent HTTP Parameter Pollution
+
+HTTP Parameter Pollution(HPP) is an attack in which attackers send multiple HTTP parameters with the same name and this causes your application to interpret them in an unpredictable way. When multiple parameter values are sent, Express populates them in an array. In order to solve this issue, you can use hpp module. This module puts array parameters in req.query and/or req.body aside and just selects the last parameter value. You can use it as follows:
```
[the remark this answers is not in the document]
mleblebici: updated accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: done

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: Added a kind of warning about this in a separate paragraph just after the code example

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: modified accordingly and added an example

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: added backticks

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: modified accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: added the recommendation

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: changed accordingly

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: added the link

cheatsheets/Nodejs_security_cheat_sheet.md
mleblebici: Added the link and create a new section for output escaping

cheatsheets/Nodejs_security_cheat_sheet.md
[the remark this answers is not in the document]
mleblebici: changed list items and parameters

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,371 @@
...
+
+This cheat sheet aims to provide a list of best practices to follow during development of Node.js applications. [Node Security Platform (nsp)](https://github.com/nodesecurity/nsp) provides a CLI tool to identify packages that have known vulnerabilities. You can install it easily with the following command:
+
+```bash
+npm install -g nsp
+```
+
+Then you can check your projects by going to the directory where your package.json file is and executing nsp check command. In addition to nsp, you can also use [Snyk](https://snyk.io) to keep your packages safe.
+
+# Proposition
```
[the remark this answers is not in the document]
mleblebici: categorized the recommendations

ThunderSon:
Hello @mleblebici ! Thank you for your contributions. Can you please push at the end? We've been hammered with more than 90 emails. Once you're done with your commits, make a single push to the PR. Thank you for your understanding

mleblebici:
@ThunderSon I am really sorry for the emails, I did not realize it sends a different email per comment. I am done with the commits. So, you guys can check it.
Again sorry for spamming :(

mackowski:
Hey thanks for this CS. I will try to review it till end of this week since it is huge :)

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,532 @@
...
+
+```JavaScript
+app.use(helmet.hidePoweredBy({ setTo: 'PHP 4.2.0' }));
+```
+
+## Platform Security
+
+### Keep your packages up-to-date
+
+Security of your application depends directly on how secure the third-party packages you use in your application are. Therefore, it is important to keep your packages up-to-date. It should be noted that [Using Components with Known Vulnerabilities](https://www.owasp.org/index.php/Top_10-2017_A9-Using_Components_with_Known_Vulnerabilities) is still in the OWASP Top 10. You can use [OWASP Dependency-Check](https://jeremylong.github.io/DependencyCheck/analyzers/nodejs.html) to see if any of the packages used in the project has a knwon vulnerability. Also you can use [Retire.js](https://github.com/retirejs/retire.js/) to check JavaScript libraries with known vulnerabilities. In order to use it, you can run the following commands in the source code folder of your application:
```
mackowski: For more tools we can link to: https://cheatsheetseries.owasp.org/cheatsheets/Vulnerable_Dependency_Management_Cheat_Sheet.html#tools.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,532 @@
...
+
+```bash
+npm install -g retire
+retire
+```
+
+### Do not use dangerous functions
+
+There are some JavaScript functions that are dangerous and should only be used where absolutely necessary. To the fullest possible extent, use of such functions and modules should be avoided. The first example is the `eval()` function. This function takes a string argument and executes it as any other JavaScript source code. Combined with user input, this behavior inherently leads to remote code execution vulnerability. Similarly, calls to `child_process.exec` are also very dangerous. This function acts as a bash interpreter and sends its arguments to /bin/sh. By injecting input to this function, attackers can execute arbitrary commands on the server.
```
mackowski: Are these all of dangerous functions? Can you review https://github.com/wisec/domxsswiki/wiki/Direct-Execution-Sinks? IMO advice "there are dangerous functions e.g. X or Y - do not use them" is not complete because from developer perspective I still do not know which functions I should avoid.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,532 @@
...
+```JavaScript
+var session = require('express-session');
+app.use(session({
+    secret: 'your-secret-key',
+    key: 'cookieName',
+    cookie: { secure: true, httpOnly: true, path: '/user', sameSite: true}
+}));
+```
+
+### Use appropriate security headers
```
mackowski: I wonder if we should create a new CS about security headers - this can be used as a foundation to it. @ThunderSon @rbsec what do you think?

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,532 @@
...
+});
+```
+
+Apart from these, there are some special functions for object attributes. `Object.preventExtensions()` prevents new properties from being added to the object.
+
+### Use access control lists
+
+Authorization prevents users from acting outside of their intended permissions. In order to do so, users and their roles should be determined with consideration of the principle of least privilege. Each user role should only have access to the resources they must use. For your Node.js applications, you can use [acl](https://www.npmjs.com/package/acl) module to provide ACL (access control list) implementation. With this module, you can create roles and assign users to these roles.
+
+## Error & Exception Handling
```
mackowski: We have https://github.com/OWASP/CheatSheetSeries/blob/33253edaa2218511193f91ad3da939b2d307f954/cheatsheets/Error_Handling_Cheat_Sheet.md maybe it should go there and here we will link to it? Any thoughts on that? @rbsec @ThunderSon @mleblebici

cheatsheets/Nodejs_security_cheat_sheet.md
rbsec: I was actually thinking about this the other day - I certainly think it's one to discuss, but not sure how it'd work.

It'd either be the authoritative CS for a lot of different areas (with links from various other CS like the XSS and TLS ones), or it'd have not much content and link out to the others for most of it.

Certainly worth discussing outside this PR.

cheatsheets/Nodejs_security_cheat_sheet.md
rbsec: I think keeping it here (maybe with a link to the error handling CS) is probably the best option for now.

That CS looks a bit dated, and I think there needs to be a wider discussion about how this sort of thing is handled in the language-specific cheat sheets. Easy enough to move it between them later if we decide to do that.

mleblebici:
I added the references, and did not change anything on security headers and error handling as you seem to agree on deciding later

ThunderSon:
Haha I'll have my review in before the end of week. Great job guys!

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+app.use(xssFilter());
+```
+
+- **X-Content-Type-Options:** Even if the server sets a valid Content-Type header in the response, browsers may try to sniff the MIME type of the requested resource. This header is a way to stop this behavior and tell the browser not to change MIME types specified in Content-Type header. It can be configured in the following way:
+
+```JavaScript
+app.use(helmet.noSniff());
+```
+
+- **Content-Security-Policy:** Content Security Policy is developed to reduce the risk of attacks like XSS and Clickjacking. Basically, it allows content from a whitelist you decide. Other content from different sources is not accepted if Content-Security-Policy headers are set correctly. It has several directives each of which prohibits loading specific type of a content. These are `connect-src`, `font-src`, `frame-src`, `img-src`, `media-src`, `object-src`, `script-src`, `style-src` and `default-src`. These can be assigned to self, none, unsafe-inline or unsafe-eval. You can implement these settings in your application as follows:
```
ThunderSon: You can link to the CSP CS instead of describing it yourself 😊 This is a CS, shouldn't tackle every subject in details.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+app.use(session({
+    secret: 'your-secret-key',
+    key: 'cookieName',
+    cookie: { secure: true, httpOnly: true, path: '/user', sameSite: true}
+}));
+```
+
+### Use appropriate security headers
+
+There are several different HTTP headers that can help you prevent some common attack vectors. These are listed below:
```
ThunderSon: You can reference the following project. Check the tabs, I think Best Practices helps the most for this.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+app.use(helmet.noSniff());
+```
+
+- **Content-Security-Policy:** Content Security Policy is developed to reduce the risk of attacks like XSS and Clickjacking. Basically, it allows content from a whitelist you decide. Other content from different sources is not accepted if Content-Security-Policy headers are set correctly. It has several directives each of which prohibits loading specific type of a content. These are `connect-src`, `font-src`, `frame-src`, `img-src`, `media-src`, `object-src`, `script-src`, `style-src` and `default-src`. These can be assigned to self, none, unsafe-inline or unsafe-eval. You can implement these settings in your application as follows:
+
+```JavaScript
+const csp = require('helmet-csp')
+app.use(csp({
+   directives: {
+       scriptSrc: ["'self'", "'unsafe-inline'"],   // helps prevent XSS attacks
```
ThunderSon: Unsafe-inline isn't that great of a recommendation

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+```JavaScript
+app.use(helmet.noSniff());
+```
+
+- **Content-Security-Policy:** Content Security Policy is developed to reduce the risk of attacks like XSS and Clickjacking. Basically, it allows content from a whitelist you decide. Other content from different sources is not accepted if Content-Security-Policy headers are set correctly. It has several directives each of which prohibits loading specific type of a content. These are `connect-src`, `font-src`, `frame-src`, `img-src`, `media-src`, `object-src`, `script-src`, `style-src` and `default-src`. These can be assigned to self, none, unsafe-inline or unsafe-eval. You can implement these settings in your application as follows:
+
+```JavaScript
+const csp = require('helmet-csp')
+app.use(csp({
+   directives: {
```
ThunderSon: I'd say having the `default-src` is the most crucial one.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+       scriptSrc: ["'self'", "'unsafe-inline'"],   // helps prevent XSS attacks
+       frame-ancestors: ["'none'"],  // helps prevent Clickjacking attacks
+       img-src: ["'self'", "'http://imgexample.com'"],
+       style-src: ["'none'"]
+    }
+}))
+```
+
+Also for further information on the usage of CSP directives, you can always refer to the [Content Security Policy Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html).
```
ThunderSon: Reference this in the start, and don't describe directives much in here.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+```JavaScript
+var session = require('express-session');
+app.use(session({
+    secret: 'your-secret-key',
+    key: 'cookieName',
+    cookie: { secure: true, httpOnly: true, path: '/user', sameSite: true}
+}));
+```
+
+### Use appropriate security headers
```
ThunderSon: I'd recommending removing lots of the wordings around each header, and linking to the MDN official documentation.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+
+It would not be fair to say that these functions and modules should not be used whatsoever, however, they should be used carefully especially when they use with user input. Also, there are [some other functions](https://github.com/wisec/domxsswiki/wiki/Direct-Execution-Sinks) that may render your application vulnerable. 
+
+### Stay away from evil regexes
+
+A Denial of Service (DoS) attack aims to make one or more of an application's resources or services unavailable for its legitimate users. Some Regular Expression (Regex) implementations cause extreme situations that makes the application very slow. Attackers can use such regex implementations to cause application to get into these extreme situations and hang for a long time.  Such regexes are called evil if application can be stuck on crafted input.  Generally, these regexes are exploited by grouping with repetition and alternation with overlapping. `(a+)+`, `(a|a?)+` are some examples of evil regexes. There are some tools to check if a regex has a potential for causing denial of service. One example is [vuln-regex-detector](https://github.com/davisjam/vuln-regex-detector).
```
ThunderSon: I think you should be talking about reDoS instead of regular DoS, and reference an explanation to it, with a small example on where it could be happening such as `(a+)+`.

cheatsheets/Nodejs_security_cheat_sheet.md
```diff
@@ -0,0 +1,534 @@
...
+```JavaScript
+"use strict";
+
+func();
+function func() {
+  y = 3.14;   // This will cause an error (y is not defined)
+}
+```
+
+### Adhere to general application security principles
```
ThunderSon: Maybe link to this? https://www.owasp.org/index.php/Security_by_Design_Principles

mleblebici:
I have made the changes requested by @ThunderSon. It now looks a lot more better than the one I first wrote.

Issue #132, opened 2019-06-17
New User Management CS


ThunderSon:
Please provide the following information about your proposal:

1. Which security issues are commonly encountered when using the technology or technique described in the cheat sheet?

- System configurations
- Bad practices
- No proper implementation for user management

2. What is the objective of the cheat sheet?

To help the reader understand the concepts behind user management.

3. What will the cheat sheet bring to the reader?

- Provide a general overview on user management concepts
- Provide a description of JWT and OAuth implementations
- Implement both technologies using Flask

righettod:
Hello, 

As the issue is assigned since many time and we do not have received any PR for this then we send it back to the backlog and we set it as **HELP_WANTED**.

Thank you very much in advance for your understanding 😃

jmanico:
stale, closing for now

PR #221, opened 2019-09-25
[WIP] Create OAuth CS skeleton

ThunderSon:
This PR tackles directly the following issue #67 and is a small part of #132 

Additions:
- Create a Draft folder for upcoming CSs
- Create a skeleton for the OAuth CS

### DIFF
--- a/cheatsheets_draft/OAuth_Cheat_Sheet.md
+++ b/cheatsheets_draft/OAuth_Cheat_Sheet.md
@@ -4,19 +4,24 @@ This CS will handle concepts related to using OAuth in different ways and method
 
 ## Table of Contents
 
-1. [OAuth Cheat Sheet](#oauth-cheat-sheet)
-   1. [Table of Contents](#table-of-contents)
-   2. [Terminologies - Getting Started](#terminologies---getting-started)
-   3. [Access Tokens](#access-tokens)
-   4. [Refresh Tokens](#refresh-tokens)
-   5. [Tokens TTL](#tokens-ttl)
-   6. [Managing Tokens](#managing-tokens)
-   7. [Redirect URI](#redirect-uri)
-   8. [Use Cases](#use-cases)
-      1. [Classic Web Applicaiton](#classic-web-applicaiton)
-      2. [Single Page Application](#single-page-application)
-      3. [Mobile Application](#mobile-application)
-      4. [Backend Service](#backend-service)
+- [Terminologies - Getting Started](#terminologies---getting-started)
+- [Access Tokens](#access-tokens)
+- [Refresh Tokens](#refresh-tokens)
+- [Tokens TTL](#tokens-ttl)
+- [Managing Tokens](#managing-tokens)
+- [Redirect URI](#redirect-uri)
+- [Security Protective Measures](#security-protective-measures)
+   - [Client Credentials Protection](#client-credentials-protection)
+   - [CSRF Protection](#csrf-protection)
+   - [Referer Header Leaks Protection](#referer-header-leaks-protection)
+   - [Token Logging Protection](#token-logging-protection)
+   - [Authorization Server Mix-Up Protection](#authorization-server-mix-up-protection)
+   - [PKCE Considerations](#pkce-considerations)
+- [Use Cases](#use-cases)
+   - [Classic Web Applicaiton](#classic-web-applicaiton)
+   - [Single Page Application](#single-page-application)
+   - [Mobile Application](#mobile-application)
+   - [Backend Service](#backend-service)
 
 ## Terminologies - Getting Started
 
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
Here is a draft list of topics that I think should be addressed:

- How to protect client credentials (`client_id` + `client_secret`)
- How to protect against CSRF attacks using the `state` parameter
- How to protect against authorization code leaks through the `Referer` header
- How to prevent tokens being accidentally logged
- How to prevent authorization server mix ups
- When to use the Proof Key for Code Exchange extension

I will soon file a PR to add those concerns into the initial table of content.

AkshayShar:
It is not merging

mackowski:
> It is not merging

@AkshayShar yes because this is a draft only for now

MarcinHoppe:
Should we mark it as a WIP somehow?

ThunderSon:
@bschach @MarcinHoppe Would love your review the soonest to push this and start writing down issues and pushing this CS forward 😊

ThunderSon:
@rbsec or @mackowski do you have any comments? Would love one approval from any of you 😄

bschach:
Looks good to me for the start! As I understand we are now focusing on client developers, we should state this clearly. Doesn't have to be in the structure yet, but I guess in the future, we can separate especially the "Security Protective Measures" section into Client Side, Resource Server and Authorization Server.

rbsec:
@ThunderSon Oauth isn't an area that I know too much about, but it looks like a reasonable skeleton to build from. Sometimes new areas become apparent when you're writing the rest of it.

A couple of areas to think about

- Logout/session revocation is often missing with these Oauth applications - anything worth discussing  about that?
- People should be using a library for this. We should recommend that, and is there any useful guidance to give around it?
- Is it worth giving a high level pros/cons of Oauth (when should you use it, and when is it not appropriate)?

MarcinHoppe:
These are all great points! I think we should add a section at the very beginning to flesh out the scope and focus. I think we should state that the initial focus is on OAuth Clients (applications). Then we can extend the CS to cover Resource Servers. I expect most people should not be creating their own Authorization Servers, so we can revisit that when we are happy with other use cases.

We can definitely discuss software reuse (i.e. use libraries vs build your own implementation).

ThunderSon:
A way we've been approaching new CSs is something similar to the released [Password Storage CS](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md). It starts off with an introduction, with the most important takeaways, followed by the ToC.

MarcinHoppe:
It really reads very well, I'd to follow the same format!

garethventon:
Format TBC but some ideas for any requirement/recommendations section:

*from a pen test of the native app flow

The state parameter should be verified by the server to prevent CSRF

The  redirect_URI  parameter  should  be  validated  to prevent  redirects resulting  in  authentication code leakage

The  authorisation  code grant  type  should  be  used rather than the implicit or any other grant type when authenticating

PKCE with  a  SHA256  hash should  be  used  by  the  application  and  server  to  provide protection  for  the authorisation code in place of a client secret

Authorisation codes should be single use only

Authorisation  codes  should  be  suitably  long  and random to deter guessing or brute force attacks

Authorisation   codes   should   not   live   longer   than  strictly required

Authorisation codes did not time out for a substantial amount of time after being issued

HTTPS should be used for all parts of the OAuth 2.0 flow

It should not be possible to modify the response type to change the flow from authorisation code to implicit

Access tokens should expire in a short time frame

Refresh tokens  should  expire  if  not  used  for  a  long period of time

Refresh  tokens  should  have  their  state  tracked  in  a database, to allow users or administrators to revoke existing  refresh  tokens  that  still  contained  a  valid  signature and to ensure that can no longer be used  to request access tokens

Refresh  tokens  should  be  reissued when  they  are used,  with state  tracking  to  indicate  which  refresh  token remains attached to the user. If a valid refresh token is used that is not the token currently attached, this  indicates  a  security  breach and  so  appropriate action   must   be   taken   (i.e.   alerting   the   user, invalidating all refresh tokens if there is a risk to the legitimate token)

If more than one device is required, and so more than one  valid  refresh  token,  these  tokens  should  be cryptographically bound to the device so appropriate checking  can  be  performed  to  ensure  only  one  is valid per device at any time

Secure  links  (Android  App  Links  or  iOS  Universal Links)  should  be  used  to  handle  the  authorisation code redirect wherever possible

Ensure  that  the  signature  on  the  JWT  access  and refresh tokens is validated appropriately, and cannot be replaced with the None algorithm or subjected to any other common signature attacks

Ensure  that  the  JWT  signature  is  suitably  long  and complex, and cannot be “cracked” by an attack

Some other thoughts out loud in the spirit of what is "good enough" is there anyway to use the L1-3 ASVS type ranking when stipulating a reference implementation pattern thats adequate pending the resource being protected,

bschach:
@garethventon Cool, that's really great input! I think we'll need to sort this into relevant for
- Client devs
- Resource Server devs
- Authorization Server devs
Many of the topics are under the responsibility of the Authorization Server and that is the one we wanted to handle last, as  those devs should usually be OAuth experts and know what they do. But fully agree, that it still makes sense to add a section for them, as the topic is so complex and offers sooo many pitfalls :-)

PR #227, opened 2019-09-28
Remove unsupported project and incomplete sections

rbsec:
The suPHP project is been unsupported for some time, so shouldn't be recommended any more. There is an unofficial fork, but really it's been replaced with functionality built into PHP-FPM.

The Nginx and CADDY sections are both empty, which makes the CS look incomplete.

To be honest the whole CS could probably do with a review and and update, but at the moment this section is actively making it look worse, so should be removed until a proper web server configuration guide can be written.


ThunderSon:
@rbsec Please open up an issue about what needs fixing, and we'll tackle it, or let someone from the community give out a hand

rbsec:
Create #232

PR #228, opened 2019-09-29
Update new cheatsheet template. Fixes #222

### LINKED ISSUE #222: Review and update new CS template
The current [pull request template](https://github.com/OWASP/CheatSheetSeries/blob/master/.github/pull_request_template.md) requires that you use the [new cheat sheet template](https://github.com/OWASP/CheatSheetSeries/blob/master/templates/New_CheatSheet.md) as one of the checks.

This template is quite old, isn't really used by any of the cheat sheets, and doesn't really fit in a many cases. It may also need updating if/when the table of contents becomes a standard. There are a few options:

* Remove the template and the requirement from the PR template.
* Update the PR template to make it clear that it's usage is optional.
* Update the template to something more general and make it clear which bits are required, and which bits are optional.
* ???

I think a minimal template could be useful which has:

* Table of Contents
* Introduction
* A couple of example headings to show the structure

rbsec:
As per the discussion on #222, this PR includes a simpler template for new cheatsheets.

If we're hoping to get a start on some of the new ones for Hacktoberfest, it would be good to get this resolved before that starts.

Comments and feedback welcome.

### DIFF
--- a/templates/New_CheatSheet.md
+++ b/templates/New_CheatSheet.md
@@ -5,7 +5,11 @@
 * Use this [editor and validation policy](https://github.com/OWASP/CheatSheetSeries#editor--validation-policy).
 * Use these [format rules](https://github.com/OWASP/CheatSheetSeries#conversion-rules).
 
-# Introduction
+# `Topic` Cheat Sheet
+
+**Replace `Topic` with the topic you're tackling, such as `Authentication` and remove this line**
+
+## Introduction
 
 Provide high level information about the topic in order to introduce it to people that do not know it.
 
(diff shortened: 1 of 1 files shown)


ThunderSon:
@rbsec @mackowski what do you think about my updates? I've been wanting to do this for some good time now. I recommend the level 2 headers because not everything is a Title, as that's what h1s are. Shouldn't be used for every section, and the CSs were lacking the name at the top.
The part about the Table of Contents is not included because I still want to decide on which extension to use between to options. I'll update accordingly and add guide on using it.

rbsec:
I think that makes sense, given how they're rendered on the generated site.

PR #247, opened 2019-10-04
Cheat sheet for Competitive Programming in python

KrishLalwani:


ThunderSon:
Hi @KrishLalwani ! This project is focused towards application security. Your PR is a general CS that doesn't actually pertain to security.

PR #264, opened 2019-10-15
Grammar

jsoref:
> :triangular_flag_on_post: If your PR is related to grammar/typo mistakes, please double check the file for other mistakes in order to fix all the issue on the current cheat sheet.

Done.

Note: most of these corrections were suggested by app.grammarly.com. The choices are all mine, but advertising credit goes to them. While I've marked the first commit as a spelling fix, that's arguable, both `double` and `check` are words, but they aren't the intended word. So, it's a misspelling of `double-check` just as `their` and `they're` are both words and using the wrong word would be a misspelling.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 Thank you for submitting a Pull Request (PR) to the Cheat Sheet Series. 
 
-> :triangular_flag_on_post: If your PR is related to grammar/typo mistakes, please double-check the file for other mistakes in order to fix all the issue on the current cheat sheet.
+> :triangular_flag_on_post: If your PR is related to grammar/typo mistakes, please double-check the file for other mistakes in order to fix all the issues on the current cheat sheet.
 
 Please make sure that for your contribution:
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@ Thank you for submitting a Pull Request (PR) to the Cheat Sheet Series.
 Please make sure that for your contribution:
 
 - [ ] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](https://github.com/OWASP/CheatSheetSeries/blob/master/templates/New_CheatSheet.md).
-- [ ] All the markdown files do not raise any validation policy violation, see the policy [here](https://github.com/OWASP/CheatSheetSeries#editor--validation-policy).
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,19 +1,19 @@
...
 
 - [ ] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](https://github.com/OWASP/CheatSheetSeries/blob/master/templates/New_CheatSheet.md).
-- [ ] All the markdown files do not raise any validation policy violation, see policy [here](https://github.com/OWASP/CheatSheetSeries#editor--validation-policy).
+- [ ] All the markdown files do not raise any validation policy violation, see the [policy](https://github.com/OWASP/CheatSheetSeries#editor--validation-policy).
 - [ ] All the markdown files follow these [format rules](https://github.com/OWASP/CheatSheetSeries#conversion-rules).
 - [ ] All your assets are stored in the **assets** folder.
 - [ ] All the images used are in the **PNG** format.
-- [ ] Any references to website have been formatted as [TEXT](URL)
-- [ ] You verified/tested the effectiveness of your contribution (e.g.: defensive code proposed is really an effective remediation? Please verify it works!).
+- [ ] Any references to the website have been formatted as [TEXT](URL)
```
rbsec: This should probably be `websites` rather than `the website`

.github/pull_request_template.md
```diff
@@ -1,19 +1,19 @@
...
+- [ ] You verified/tested the effectiveness of your contribution (e.g., the defensive code proposed is really an effective remediation? Please verify it works!).
 - [ ] The CI build of your PR pass, see the build status [here](https://travis-ci.org/OWASP/CheatSheetSeries/pull_requests).
 
-If your PR is related to an issue. Please end your PR text with the following line:
+If your PR is related to an issue. Please finish your PR text with the following line:
```
rbsec: This should probably be one sentence with a comma between the two parts - otherwise the first is just a fragment that doesn't make sense.

.github/pull_request_template.md
jsoref: Oh, that makes sense. Will do.

.github/pull_request_template.md
jsoref: Indeed.

jsoref:
@rbsec: fixed (thanks for your prompt response)

rbsec:
Thanks for the PR - merged.

PR #269, opened 2019-10-16
Changing comments

PauloVAS:


mackowski:
Thank you for this PR but it does not add much value in my opinion.

PR #270, opened 2019-10-17
More performant Double Submit Cookie

Xeoncross:
A simpler alternative to an encrypted cookie is to hash the token with a secret salt known only by the server and place this value in a cookie.

### DIFF
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -219,6 +219,8 @@ Scenarios a and b mentioned above are possible only if the CSRF token is not der
 
 Including the token in an encrypted cookie - other than the authentication cookie (since they are often shared within subdomains) - and then at the server side matching it (after decrypting the encrypted cookie) with the token in hidden form field or parameter/header for ajax calls mitigates both the issues mentioned above. This works because a sub domain has no way to over-write an properly crafted encrypted cookie without the necessary information such as encryption key.
 
+A simpler alternative to an encrypted cookie is to hash the token with a secret salt known only by the server and place this value in a cookie. This is similar to an encrypted cookie (both require knowledge only the server holds), but is less computationally intensive than encrypting and decrypting the cookie. Whether encryption or a salted-hash is used, an attacker won't be able to recreate the cookie value from the plain token without knowledge of the server secrets.
+
 **Note about Triple Submit Cookie** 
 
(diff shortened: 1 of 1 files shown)


mackowski:
Thank you for your PR. I will review it this week.

Xeoncross:
I would say that at some point, most of the section on "Double Submit Cookie" needs to be clarified and grouped. So far, we have unencrypted, encrypted, and now hashing approached listed without what might be considered clear separation of each.

For example, with encrypted or hashed cookie values there is a conflict with the advice given in a prior paragraph:

> "With double submit, if an attacker can write a cookie they can obviously defeat the protection."

This is not true unless they also have access to the encryption key _or_ hash salt (which ever method is used). Writing a cookie without knowledge of these secrets will still fail as the server would notice the invalid results.

mackowski:
@Xeoncross would you like to create a PR that will clarity, shorten and make this section more explicit? We can merge this now and create a separate issue for that work.

ThunderSon:
@rbsec so I was looking at the csurf module for express [here](https://github.com/pillarjs/csrf). They create a salt of 8 bytes, with a secret of 18 bytes. @Xeoncross can you give us some references to your text, or is this based on experience? It's a totally valid point, having some articles/references with it is better.

rbsec:
@ThunderSon 18 bytes seems a rather odd length - I wonder why they didn't go for 16 or 32. 16 bytes (128 bits) is more than enough to protect against brute-force (assuming it's been securely generated of course), so I don't see much value in going above that.

Xeoncross:
@mackowski I would be happy to open a separate PR when I have some time to sit down and rewrite this section. Double-submit cookies are a very important part of blocking unauthorized access from what I've seen and at the same time, are simple to implement - especially with single page apps like react/vue/angular.

@ThunderSon I've spoken to a number of developers about this approach and realized that most had not thought about simply hashing the token which led to my original PR. I'm not sure where I first ran across this approach, but it's one I've used.

@rbsec I'll just add that it's okay for the salt to be reused for multiple unique session tokens as each token should be random itself. So using a longer run of salt randomness (minimum 128bits) should not be a problem to generate once per day. The salt can be rotated every day or two (keeping the old salt around a while to "port" tokens to the new salt). Of course, you could generate a unique salt for each user - but you would have to store that with the session.

rbsec:
@Xeoncross to be honest, with a 128bit salt I don't really see a problem if the salt is essentially static on the website - it's going to be impossible to crack (assuming it's generated properly), so it would only really need changing if you have reason to believe that it's been compromised. Certainly wouldn't hurt to change it though (apart from invalidating any existing sessions).

Do you think it's worth adding some (brief) discussion of this? I think key points are:

1. Make it long enough (at least 128bits)
2. Generate it securely
3. Consider regular changes, or change if believed to be compromised.

mackowski:
I will merge this PR and create a new issue for clarify "Double Submit Cookie" section

PR #239, opened 2019-10-03
update on php hardening options

anantshri:
1. I have moved suhosin to the bottom as its only applicable for 5.X and not 7.X series and 5.x is EOL
2. I have corrected the link to project page instead of a 3 pager unknown wordpress blog
3. I have added snuggleupagus which is the successor of suhosin however its in beta phase hence added at the bottom.

Thank you for submitting a Pull Request (PR) to the Cheat Sheet Series.

Issue #232, opened 2019-10-01
Cheat sheet update/refactor proposal: PHP Configuration


rbsec:
There are a number of issues with this CS:

* Links to a guide on secure PHP development, which isn't really about configuration
* Links to a barely-used GitHub repo, which only has links to various other guides/pages
* Recommends Suhosin, which AFAIK doesn't work with PHP 7 and isn't really supported
* No real explanation of what the recommended settings do, or why to use them
* Are the recommendations actually still valid and useful for the latest PHP?

Adding this issue as a reminder for now - need to decide the best approach to handling this (and to what extend server hardening like this is in the scope of the project).

anantshri:
ah didn't saw this earlier i have just submitted a pull request to update suhosin and add another option. 
https://github.com/OWASP/CheatSheetSeries/pull/239

I am in for suggestions on how to move forward.

I would propose
1. We add details to what each of the option does
2. We remove non-needed link.

I would also propose spliting cheatsheet project in subsections BUILDERS and BREAKERS. this way we can attack and defence cheatsheets in same place.

ThunderSon:
The CS project is purely for builders. For breakers, the [testing guide](https://github.com/OWASP/OWASP-Testing-Guide-v5) is there to help them. Our main focus is developers trying to secure their systems and don't have enough time to learn about a given topic. 😄

mackowski:
this was already done in #276

PR #276, opened 2019-10-27
Update Suhosin text in PHP hardening CS

rbsec:
Implement the changes discussed in #239 about Suhosin, pending a wider review of the CS in #232.

Links to Snuffleupagus and Suhosin-ng are included with an appropriate warning not because we're recommending their use, but because it preempts the obvious question of "what about PHP 7" when it's stated that Suhosin only works for PHP 5.

### DIFF
--- a/cheatsheets/PHP_Configuration_Cheat_Sheet.md
+++ b/cheatsheets/PHP_Configuration_Cheat_Sheet.md
@@ -10,10 +10,6 @@ For general PHP codebase security please refer to the two following great guides
 
 # PHP Configuration and Deployment
 
-## Suhosin
-
-Consider using [Suhosin](http://www.hardened-php.net/suhosin/index.html) if you want to patch many custom security flaws in various parts of PHP. 
-
 ## php.ini
 
 Some of following settings need to be adapted to your system, in particular `session.save_path`, `session.cookie_path` (e.g. `/var/www/mysite`), and `session.cookie_domain` (e.g. `ExampleSite.com`). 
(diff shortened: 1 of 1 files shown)


cheatsheets/PHP_Configuration_Cheat_Sheet.md
```diff
@@ -111,3 +107,9 @@ report_memleaks         = On
 track_errors            = Off
 html_errors             = Off
 ```
+
+## Suhosin
+
+[Suhosin](http://www.hardened-php.net/suhosin/index.html) is a patch to PHP which provides a number of hardening and security features that are not available in the default PHP build. However, Suhosin only works with PHP 5, which is **unsupported** and **should not be used**.
+
+There are a number of projects that have attempted to implement similar features for PHP 7 such as [Snuffleupagus](https://snuffleupagus.readthedocs.io/) and [Suhosin-ng](https://github.com/sektioneins/suhosin-ng/wiki/News), but these are both in the pre-release stage, and as such **should not be used in production**.
```
anantshri: *attempting* would be more appropriate as both projects are work in progress at this point. attempted feels like attempted but never succeeded and then project closed.

ThunderSon: @rbsec Let me know what you think. For me, this is not a blocker, can be amended definitely.

rbsec: Hi both,

I think `have attempted` is more accurate, as there have also been other projects such as [suhosin7](https://github.com/sektioneins/suhosin7) which didn't make huge progress and have subsequently been abandoned.

mackowski:
I believe that we can merge that right? @rbsec

rbsec:
@mackowski I'm happy for it to be merged.

cheatsheets/PHP_Configuration_Cheat_Sheet.md
anantshri: fair enough.

anantshri:
good to go from my side too.

PR #277, opened 2019-10-27
Updates to license and DNS sections on readme

rbsec:
No change the the actual content - just reformatted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,32 +240,19 @@ A special thank you to the following people for their help provided during the m
 
 # License
 
-See [here](LICENSE.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -240,32 +240,19 @@ A special thank you to the following people for their help provided during the m
 
 # License
 
-See [here](LICENSE.md).
+The entire project content is under the **[Creative Commons v3.0](https://creativecommons.org/licenses/by-sa/3.0/)** license.
```
ThunderSon: Remove the license file to have a clean change.

rbsec: @ThunderSon since LICENSE.md is something of an informal standard I think it'd be better to leave it there - as it's a default place people (or scripts) might look.

PR #305, opened 2019-11-22
Rewrite of email validation guidance. Fixes #298

### LINKED ISSUE #298: Update: Input Validation CS - email subaddressing
## Input_Validation_Cheat_Sheet.md Update
I propose to add to the Email Address Validation section the "careful with mail widlcard section"

Add section: Careful With Wildcards

Popular mail services like Gmail offer wildcard modifiers in order to help users to organize inbox and filters.

1) Append a plus ("+") sign and any combination of words or numbers after your email address. For example, if your name was example@gmail.com, you could send mail to example+friends@gmail.com or example+mailinglists@gmail.com.

2) Insert one or several dots (".") anywhere in your email address. Gmail doesn't recognize periods as characters in addresses -- we just ignore them. For example, you could tell people your address was example@gmail.com, examp.le@gmail.com or ex.am.ple@gmail.com. (We understand that there has been some confusion about this in the past, but to settle it once and for all, you can indeed receive mail at all the variations with dots.)

Before register a new Gmail address, strip "+" and "." before check that the email is already registered.

rbsec:
Following form the discussion on #298, I've rewritten the email validation guidance. The old guidance had some slightly confusing terminology in places, and didn't properly cover some areas such as sub-addressing or disposable email addresses.

@aiacobelli2 I'm interested to hear your feedback.

@ThunderSon and @mackowski I split this into two sections (syntactic and semantic) - what do you think of this? Do you think this would be a good pattern to follow for each section in the CS? If so, are there better (more clear) names for the sections?

### DIFF
--- a/cheatsheets/Input_Validation_Cheat_Sheet.md
+++ b/cheatsheets/Input_Validation_Cheat_Sheet.md
@@ -214,11 +214,11 @@ Blocking disposable email addresses is almost impossible, as there are a large n
 
 If these lists are used to block the use of disposable email addresses then the user should be presented with a message explaining why they are blocked (although they are likely to simply search for another disposable provider rather than giving their legitimate address).
 
-If it is imperative that disposable email addresses are blocked, then registrations should only be allowed from a specifically whitelisted email providers. However, if this includes public providers such as Google or Yahoo, users can simply register their own disposable address with them.
+If it is essential that disposable email addresses are blocked, then registrations should only be allowed from specifically whitelisted email providers. However, if this includes public providers such as Google or Yahoo, users can simply register their own disposable address with them.
 
 ### Sub-Addressing
 
-Sub-addressing allows a user to specify "tag" in the local part of the email address (before the `@` sign), which will be ignored by the mail server. For example, if that `example.org` domain supports sub-addressing, then the following email addresses are equivalent:
+Sub-addressing allows a user to specify a _tag_ in the local part of the email address (before the `@` sign), which will be ignored by the mail server. For example, if that `example.org` domain supports sub-addressing, then the following email addresses are equivalent:
 
 - `user@example.org`
 - `user+site1@example.org`
@@ -226,6 +226,6 @@ Sub-addressing allows a user to specify "tag" in the local part of the email add
 
 Many mail providers (such as Microsoft Exchange/Office 365) do not support sub-addressing. The most notable provider who does is Gmail, although there are many others that also do.
 
-Some users will use a different tag for each website they register on, so that if they start receiving spam to one of the sub-addresses they can identify which website leaked or sold their email address.
+Some users will use a different _tag_ for each website they register on, so that if they start receiving spam to one of the sub-addresses they can identify which website leaked or sold their email address.
 
 Because it could allow users to register multiple accounts with a single email address, some sites may wish to block sub-addressing by stripping out everything between the `+` and `@` signs. This is not generally recommended, as it suggests that the website owner is either unaware of sub-addressing or wishes to prevent users from identifying them when they leak or sell email addresses. Additionally, it can be trivially bypassed by using [disposable email addresses](#disposable-email-addresses), or simply registering multiple email accounts with a trusted provider.
--- a/cheatsheets/Input_Validation_Cheat_Sheet.md
+++ b/cheatsheets/Input_Validation_Cheat_Sheet.md
@@ -185,6 +185,9 @@ As such, the best way to validate email addresses is to perform some basic initi
 - The email address does not contain dangerous characters (such as backticks, single or double quotes, or null bytes).
   - Exactly which characters are dangerous will depend on how the address is going to be used (echoed in page, inserted into database, etc).
 - The domain part contains only letters, numbers, hyphens (`-`) and periods (`.`).
+- The email address is a reasonable length:
+  - The local part (before the `@`) should be no more than 63 characters.
+  - The total length should be no more than 254 characters.
 
 ## Semantic Validation
 
--- a/cheatsheets/Input_Validation_Cheat_Sheet.md
+++ b/cheatsheets/Input_Validation_Cheat_Sheet.md
@@ -168,7 +168,7 @@ The upload feature should be using a whitelist approach to only allow specific f
 
 ## Syntactic Validation
 
-The format of email addresses is defined by [RFC 5321](https://tools.ietf.org/html/rfc5321#section-4.1.2), and if far more complicated than most people realise. As an example, the following are all considered to be valid email addresses:
+The format of email addresses is defined by [RFC 5321](https://tools.ietf.org/html/rfc5321#section-4.1.2), and is far more complicated than most people realise. As an example, the following are all considered to be valid email addresses:
 
 - `"><script>alert(1);</script>"@example.org`
 - `user+subaddress@example.org`
@@ -182,7 +182,7 @@ The biggest caveat on this is that although the RFC defines a very flexible form
 As such, the best way to validate email addresses is to perform some basic initial validation, and then pass the address to the mail server and catch the exception if it rejects it. This means that any the application can be confident that its mail server can send emails to any addresses it accepts. The initial validation could be as simple as:
 
 - The email address contains two parts, separated with an `@` symbol.
-- The email address does not contain dangerous characters (such as backticks, single or double quotes, or null bytes
+- The email address does not contain dangerous characters (such as backticks, single or double quotes, or null bytes).
 - The domain part contains only letters, numbers, hyphens (`-`) and periods (`.`).
 
 ## Semantic Validation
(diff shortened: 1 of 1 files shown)


cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,62 @@ The upload feature should be using a whitelist approach to only allow specific f
...
+- Time limited (e.g, expiring after eight hours).
 
-## Address Normalization
+After validating the ownership of the email address, the user should then be required to authenticate on the application through the usual mechanism.
 
-As the local-part of email addresses are, in fact - case sensitive, it is important to store and compare email addresses correctly. To normalise an email address input, you would convert the domain part ONLY to lowercase.
+### Disposable Email Addresses
 
-Unfortunately this does and will make input harder to normalise and correctly match to a users intent. It is reasonable to only accept one unique capitalisation of an otherwise identical address, however in this case it is critical to:
+In some cases, users will may not want to give their real email address when registering on the application, and will instead provide a disposable email address. These are publicly available addresses that do not require the user to authenticate, and are typically used to reduce the amount of spam received by users' primary email addresses.
```
ThunderSon: users ~will~ may not

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,62 @@ The upload feature should be using a whitelist approach to only allow specific f
 
 # Email Address Validation
 
-## Email Validation Basics
+## Syntactic Validation
 
-Many web applications do not treat email addresses correctly due to common misconceptions about what constitutes a valid address. Specifically, it is completely valid to have an mailbox address which:
+The format of email addresses is defined by [RFC 5321](https://tools.ietf.org/html/rfc5321#section-4.1.2), and if far more complicated than most people realise. As an example, the following are all considered to be valid email addresses:
```
ThunderSon: and `is` far from ... instead of `if`

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,62 @@ The upload feature should be using a whitelist approach to only allow specific f
...
 
-- Increased popularity of sub-addressing by providers such as Gmail (commonly using `+` as a token in the local-part to affect delivery)
--  New [gTLDs](https://en.wikipedia.org/wiki/Generic_top-level_domain) with long names (many regular expressions check the number and length of each label in the domain)
+As such, the best way to validate email addresses is to perform some basic initial validation, and then pass the address to the mail server and catch the exception if it rejects it. This means that any the application can be confident that its mail server can send emails to any addresses it accepts. The initial validation could be as simple as:
 
-Following [RFC 5321](https://tools.ietf.org/html/rfc5321), best practice for validating an email address would be to:
+- The email address contains two parts, separated with an `@` symbol.
+- The email address does not contain dangerous characters (such as backticks, single or double quotes, or null bytes
```
ThunderSon: Close parentheses. I'd say dangerous characters should be defined based on how the email is used. It needs to be context aware. You think we should add that point?

ThunderSon:
Syntatic and semantic sections are pretty much specific for the points being tackled. It's debatable to have clear sections that can be followed for every CS out there.

rbsec:
@ThunderSon I meant for each section of this CS (general input, files, email addresses, etc).

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,62 @@ The upload feature should be using a whitelist approach to only allow specific f
...
+- The domain part contains only letters, numbers, hyphens (`-`) and periods (`.`).
```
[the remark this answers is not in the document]
rbsec: The RFC says only letters, numbers, hyphens (and periods to separate labels):

https://tools.ietf.org/html/rfc1035 (section 2.3.1):
> The labels must follow the rules for ARPANET host names.  They must
> start with a letter, end with a letter or digit, and have as interior
> characters only letters, digits, and hyphen.  There are also some
> restrictions on the length.  Labels must be 63 characters or less.

You can get internationalised domains, but they should be represented as punycode (the `xn--..` format).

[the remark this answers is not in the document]
ThunderSon: Alright that works well in my opinion. I was thinking if it went behind the punycode format and was sent straight, should it then be blocked or should the mail library interpret it?

ThunderSon:
For this CS I can see this happening based on every section's best fit.

cheatsheets/Input_Validation_Cheat_Sheet.md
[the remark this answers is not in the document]
rbsec: From what I've read, support for internationalised addresses can be pretty patchy, so a lot of mail libraries (or downstream mail servers) will reject them.

cheatsheets/Input_Validation_Cheat_Sheet.md
rbsec: Added a note about this in 18a878d - is that what you were thinking?

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,66 @@ The upload feature should be using a whitelist approach to only allow specific f
...
 
-- Store the user-part as provided and verified by user verification.
-- Perform comparisons by `lowercase(provided)==lowercase(persisted)`.
+Blocking disposable email addresses is almost impossible, as there are a large number of websites offering these services, with new domains being created every day. There are a number of publicly available lists and commercial lists of known disposable domains, but these will always be incomplete.
+
+If these lists are used to block the use of disposable email addresses then the user should be presented with a message explaining why they are blocked (although they are likely to simply search for another disposable provider rather than giving their legitimate address).
+
+If it is imperative that disposable email addresses are blocked, then registrations should only be allowed from a specifically whitelisted email providers. However, if this includes public providers such as Google or Yahoo, users can simply register their own disposable address with them.
```
ThunderSon: - Let's avoid the usage of `imperative` as it might be heavy to non native speakers.
- from ~a~ specifically whitelisted email providers. I believe the `a` should be removed as we are talking about multiple providers.

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,66 @@ The upload feature should be using a whitelist approach to only allow specific f
...
+
+### Sub-Addressing
+
+Sub-addressing allows a user to specify "tag" in the local part of the email address (before the `@` sign), which will be ignored by the mail server. For example, if that `example.org` domain supports sub-addressing, then the following email addresses are equivalent:
```
ThunderSon:
```suggestion
Sub-addressing allows a user to specify a _tag_ in the local part of the email address (before the `@` sign), which will be ignored by the mail server. For example, if that `example.org` domain supports sub-addressing, then the following email addresses are equivalent:
```

cheatsheets/Input_Validation_Cheat_Sheet.md
```diff
@@ -166,42 +166,66 @@ The upload feature should be using a whitelist approach to only allow specific f
...
+
+- `user@example.org`
+- `user+site1@example.org`
+- `user+site2@example.org`
+
+Many mail providers (such as Microsoft Exchange/Office 365) do not support sub-addressing. The most notable provider who does is Gmail, although there are many others that also do.
+
+Some users will use a different tag for each website they register on, so that if they start receiving spam to one of the sub-addresses they can identify which website leaked or sold their email address.
```
ThunderSon: _tag_ instead of tag.

xanderificnl:
I'm unsure whether GitHub actually did save my review. I wasn't getting "any" feedback when I clicked any button. So I do apologize if it get saved several times, or if it didn't do anything.

To reiterate, my "review" was regarding the maximum length limits of email addresses. This was my comment:

---

I had just read RFC 3969 and then looked at OWASP recommendations. The note about the total length not exceeding 254 characters doesn't match RFC 3969 (in which the domain part has a max length of 255).

After some research, I found [this errata](https://www.rfc-editor.org/errata/eid1690) that explains that RFC 2821 restricts the maximum length of the MAIL and RCPT (smtp) commands to 254 characters. However, it appears that RFC 5321 (the successor of 2821) does increase limits, as per [this section](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1).

A reviewer might wish to consider updating this section to reflect [RFC 3969](https://datatracker.ietf.org/doc/html/rfc3696#section-3). To quote from the RFC:

> There is a length limit on email addresses.  That limit is a maximum of **64 characters (octets) in the "local part"** (before the "@") and a **maximum of 255 characters (octets) in the domain part** (after the "@") for a **total length of 320 characters**.  Systems that handle email should be prepared to process addresses which are that long, even though they are rarely encountered.

---

Again, I do apologize if you got pinged several times. I didn't receive any feedback from GitHub's UI, and it didn't seem to do anything.

PR #310, opened 2019-11-30
File Upload Rewrite

### LINKED ISSUE #304: Review File Upload CS
The [file upload](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Protect_FileUpload_Against_Malicious_File.md) CS has a few issues:

- Hasn't been updated for a while
- Doesn't really have very much useful information
- 90% of the CS is just Java code samples (are these still valid/working/correct?)
- Some of the samples are using a commercial library (#303)

The whole CS should be reviewed, and either rewritten or removed.

ThunderSon:
The File Upload CS has been updated with clear advice on:
- Threats relating to the attack vector
- Prevention mechanisms
- Best practices

Every commit handles a section, which allows the reviewer to go through the PR commit by commit.
If the structure needs modification or improvement, please let me know too.

This PR covers closes #304


cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+### Malware Upload
 
-- For Word/Excel/PowerPoint/Pdf documents: Detect when a document contains "code"/OLE package, if it's the case then block the upload process.
-- For Images document: Sanitize incoming image using re-writing approach and then disable/remove any "code" present (this approach also handle case in which the file sent is not an image).
+The attacker delivers a file for malicious intent, such as:
 
-Remarks:
+1. Call the file to execute code (_e.g._ [ImageTrick Exploit](https://imagetragick.com/))
+2. Use the file for phishing (_e.g._ careers form)
+3. Send ZIP bombs, or simply huge files in a way to fill the server storage which hinders the server's availability
```
rbsec: Also XML bombs (billion laughs)

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
 
-Based on this context, the goals here are:
+### Malware Upload
 
-- For Word/Excel/PowerPoint/Pdf documents: Detect when a document contains "code"/OLE package, if it's the case then block the upload process.
-- For Images document: Sanitize incoming image using re-writing approach and then disable/remove any "code" present (this approach also handle case in which the file sent is not an image).
+The attacker delivers a file for malicious intent, such as:
 
-Remarks:
+1. Call the file to execute code (_e.g._ [ImageTrick Exploit](https://imagetragick.com/))
```
rbsec: Also more general file parser exploits, including things like XXE (which we have a CS for).

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
-- Microsoft Office document: Word/Excel/PowerPoint using [VBA Macro](https://en.wikipedia.org/wiki/Visual_Basic_for_Applications) and [OLE package](https://en.wikipedia.org/wiki/Object_Linking_and_Embedding).
-- Adobe PDF document: Insert malicious code as attachment.
-- Images: Malicious code embedded into the file or use of binary file with image file extension.
+## File Upload Threats
 
-# Approaches
+In order to assess and know exactly what controls to implement, knowing what you're facing is essential to protect your assets. The following sections will hopefully showcase the risks accompanying the file upload functionality.
 
-Based on this context, the goals here are:
+### Malware Upload
```
rbsec: Perhaps "Malicious Files", since some of these aren't really malware.

rbsec: Also need to have a section about dealing with the the contents of files, which could be:

* Personal data (GDPR issues)
* Copyrighted (piracy)
* Illegal (CSE, stolen data, etc)
* Reputationaly damaging (especially if publicly available)
* Dangerous (macros, CSV formula injection, XXE payloads, etc)

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+3. Send ZIP bombs, or simply huge files in a way to fill the server storage which hinders the server's availability
+4. Overwrite an existing file on the system
+5. Client-side active content that could lead to all types of client-side attacks if the file is publicly retrievable
 
-- It's technically possible to perform sanitizing on Word/Excel/PowerPoint/PDF documents but we have chosen here the option to block them in order to avoid the risk of missing any evasion techniques and then let pass one evil document. The following [site](https://www.greyhathacker.net/?p=872) show how many way exists to embed Macro into a Microsoft Office documents.
-- The other reason why we have chosen the blocking way is that for Word/Excel/PowerPoint, changing document format (for example by saving any document to `DOCX/XSLX/PPTX/PPSX` formats in order to be sure that no Macro can be executed) can have impacts or cause issues on document structure/rendering depending on the API used.
+### Public File Retrieval
 
-# Cases
+If the file uploaded is publicly retrievable, additional threats can be addressed:
```
rbsec: See earlier point about illegal/inappropriate files

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+2. Initiate a DoS attack by requesting lots of files. Requests are small, yet responses are much larger
 
-The following codes are shared by the code snippets proposed into the rest of this article.
+## File Upload Protection
 
-Interfaces:
+### Extension Validation
 
-*DocumentDetector*
+Ensure that the validation occurs after decoding the file name, and that a proper filter is set in place (_e.g._ `\.png` vs `\.png\z`; etc.). Refer to the [Input Validation CS](Input_Validation_Cheat_Sheet.md) for more information.
```
rbsec: You often get bugs around this, so giving a few specific warnings here might be a good idea about:

* Double extensions (.jpg.php)
* Null bytes (.php%00.jpg)
* Bad regex (such as a regex for `.html$`, which could match executable formats like `.cshtml`).

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
-     *
-     * @param f File to made safe
-     *
-     * @return TRUE only if the specified file has been successfully made safe.
-     */
-    boolean madeSafe(File f);
-}
-```
+- image upload, allow one type that is agreed upon to fit the business requirement;
+- cv upload, allow `docx` and `pdf` extensions. Access files through a closed down machine.
```
rbsec: The `Access files...` point seems a bit out of place here

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
-            }
-        } catch (Exception e) {
-            safeState = false;
-            LOG.warn("Error during Image file processing !", e);
-        }
-
-        return safeState;
-    }
-}
+```python
```
rbsec: Not sure this really adds much - it's very simple and we don't want to make it easier for people to blacklist. I'd say to remove this.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
-            }
-        } catch (Exception e) {
-            safeState = false;
-            LOG.warn("Error during PowerPoint file analysis !", e);
-        }
-        return safeState;
-    }
-}
-```
+> This should be only used when the [whitelist approach](#whitelist-extensions) fails, and is in no way a replacement for it.
```
rbsec: Needs big warnings

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+```python
+# This code snippet only serves as an example, and should not be used
+blacklist = ["php", "php5"]  # can be replaced with a regex pattern, but that would be even weaker, such as \.php$
+for bad in blacklist:
+    if "file.php".endswith(bad):
+        // raise exception
 ```
 
-# Sources of the prototype
+### User Permissions
```
rbsec: Section seems a bit out of place - file type and content type are much more related. I'd either move this earlier or alter.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+Before any file upload service is accessed, proper validation should occur on two levels for the user uploading a file:
+
+- Authentication level
+  - The user should be a registered user, or an identifiable user, in order to set restrictions and limitations for their upload capabilities
+- Authorization level
+  - Make sure that the user's accesses allow them to upload a file, or multiple files for administrative cases.
+
+### Content-Type Validation
+
+Other than defining the extension of the uploaded file, its MIME-type can be checked for a quick protection against simple file upload attacks.
```
rbsec: Not really clear how you'd `check` it - I'm assuming matching the content type to the file type or whitelist?

Needs strong emphasis that this is trivially spoofed and should not be relied upon.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+
+This can be done preferrably in a whitelist approach; otherwise, this can be done in a blacklist approach.
+
+> This should not be used on its own, as bypassing it is pretty common and easy.
+
+### File Name Sanitization
```
rbsec: I think this section needs to be expanded a bit. There a few different issues here that should be split into (sub)sections:

- Validating the file name that the user provides when uploading.
  - Length
  - Dangerous characters (null bytes, directory traversal, colons, slashes, etc)
  - File extension (subsection?)
- What name (if any) the file should be stored under.
  - Never use the provided filename unless you have to
  - Avoid predictable names for user-facing files (incremented numbers, timestamps, hashes, etc)
  - Dangerous characters
  - Dangerous filenames (`CON`, `AUX`, `.htaccess`, etc)
- What name the file should be served under
  - Dangerous characters like CRLF (response splitting/header injection)
  - Dangerous extensions
  - Content-type disposition
  - Browsers doing stupid things like MIME sniffing

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+Multiple threats relate to the file-name, thus, it is essential to ensure that the file-name is stored in a secure fashion.
+
+This can be done by creating a **random string** as a file-name. If the file-name is required by the business needs, proper input validation should be done for client-side (_e.g._ active content) and back-end side (_e.g._ special files overwrite or creation) attack vectors. File-name length limits should be taken into consideration based on the system storing the files, as each system has its own file name length limit.
+
+### File Storage Location
+
+All file uploaded will be stored in a location on the server. This location must be chosen based on security and business requirements. The following points are set by security priority, and are inclusive:
+
+1. Store the files on a **different host**, which allow for complete segragation of duties between the application serving the user, and the host handling file uploads and their storage.
+2. Store the files in a **private directory**, where only administrative access is allowed.
```
rbsec: Not entirely sure what `private` means here? Is this outside of the webroot, or is it only accessible to admin/authenticated users?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+3. Store the files on a **public directory**, and set them in write permissions only.
```
rbsec: As above, is this within the webroot? I assume `write permissions only` means `block public access`?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+
```
rbsec: There's also the option to store the files in a (No)SQL database as BLOBs. This can resolve lots of filename/path related issues, but also makes it harder to AV scan then.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+### File Name Sanitization
+
+Multiple threats relate to the file-name, thus, it is essential to ensure that the file-name is stored in a secure fashion.
+
+This can be done by creating a **random string** as a file-name. If the file-name is required by the business needs, proper input validation should be done for client-side (_e.g._ active content) and back-end side (_e.g._ special files overwrite or creation) attack vectors. File-name length limits should be taken into consideration based on the system storing the files, as each system has its own file name length limit.
+
+### File Storage Location
+
+All file uploaded will be stored in a location on the server. This location must be chosen based on security and business requirements. The following points are set by security priority, and are inclusive:
```
rbsec: Files aren't always stored - they might be emailed to someone (such as a CV upload/contact form), or processed and parsed (such as a CSV upload).

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+
+Files should be stored in a way that ensures:
+
+- Allowed system users are the only ones capable of reading the files
+- Required modes only are set for the file
+  - If execution is required, scanning the file before running it is required as a security best practice, to ensure that no macros or hidden scripts are available.
+
+### Upload and Download Limits
+
+The application should set proper size limits for the upload service in order to protect the file storage capacity. The file size limit should be considered after file decompression is conducted, and by using secure methods to calculate zip files size.
```
rbsec: Maybe some guidance about how to securely calculate file sizes? Or just not decompressing files

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+
+It should set proper request limits as well for the download service if available to protect the server from DoS attacks.
```
rbsec: I'd be tempted to remove the stuff about downloading - this is a whole separate area with **lots** of potential issues. Perhaps a separate CS?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: That is an example. Want me to add more examples to it? This is not covering all possible attacks.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: So I was considering that part, then again, how can they cover it? I discussed the last point, the dangerous files.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Can definitely mention them, just thought of keeping it small.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Will add an emoji and maybe make them bold.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: They can be stored in DBs if they aren't that big and the databases can be properly scaled. I'll add a point to it

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Private as in not in the webroot.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: I think the 3rd point is not relevant much, as that will go into permissions.
My point was that users can upload files normally (so they only require write permissions), and people that require those files, read permissions.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: What would we be tackling specifically in there? To see if it requires a whole CS, or a section.

ThunderSon: Planning to rename this CS too, once we're done. Can be something to cover both.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: Perhaps it's just the phrase "call the file to execute code" that was confusing me - maybe this should be something like "exploit vulnerabilities in the file parser or processing"?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: It'll depend on exactly what the site is doing with uploaded files, but you could have stuff like:

* A process for users to report illegal/inappropriate files
* A process for copyright owners to report abuse
* Manual review of files (before they're made public)?
* Automated review (filename/keyword analysis, hash databases, image/video processing, etc)

Testing for Macros/CSV formula/XXE is going to be harder - it'll vary depending on the exact type of file

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: In the vast majority of cases I'd expect the same application to be reading and writing the files - so not sure how it'd work without read permissions?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: This is the kind of thing that a NoSQL database (essentially a key:value store for arbitrary data) could be perfect for.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: It's often an area that I see very bad vulns around when doing pentests. Areas such as:

* Directory traversal/arbitrary download
* Downloads from other servers (SSRF, `file://` URIs, etc)
* Filename tampering (e.g, having files downloaded with dangerous extensions such as .bat, CRLF injection, etc)
* Access control issues on a file download handler
* No access control for files in the webroot

Perhaps not a CS in its own right, but certainly a decent size section if it's in another one.

Maybe this could become a combined "file upload and download" or something?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
rbsec: I'd perhaps have it as something along the lines of.

"Blacklisting extensions is a bad idea and is very dangerous. Don't do it unless you have no other choice"

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+    - [File Name Sanitization](#file-name-sanitization)
+    - [File Storage Location](#file-storage-location)
+    - [System Permissions](#system-permissions)
+    - [Upload and Download Limits](#upload-and-download-limits)
+- [Java Code Snippets](#java-code-snippets)
 
-- Microsoft Office document: Word/Excel/PowerPoint using [VBA Macro](https://en.wikipedia.org/wiki/Visual_Basic_for_Applications) and [OLE package](https://en.wikipedia.org/wiki/Object_Linking_and_Embedding).
-- Adobe PDF document: Insert malicious code as attachment.
-- Images: Malicious code embedded into the file or use of binary file with image file extension.
+## File Upload Threats
```
rbsec: Would be worth a mention that file upload can be vulnerable to CSRF just like any other form - less common to see people exploiting it, but not one you'd want to forget.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
-# Introduction
+# File Upload Cheat Sheet
 
-This article propose a way to protect a file upload feature against submission of file containing malicious code.
+## Introduction
 
-# Context
+File upload is becoming more and more an essential part of any application, where the user is able to upload their photo, their CV, or a video showcasing a project they are working on. It is evident that applications should be able to handle file uploads based on the business requirement. In return, the application should be able to fend off bogus, and malicious files in a way to keep the application and the users safe.
```
nightwatchcyber: Should there be a reference here to the input validation cheetsheet, being that handling file uploads is a special case of input validation?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
nightwatchcyber: There are libraries like Apache Tika that can sniff the file content and compare to the extension

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,138 @@
...
+
+This can be done preferrably in a whitelist approach; otherwise, this can be done in a blacklist approach.
+
+> This should not be used on its own, as bypassing it is pretty common and easy.
+
+### File Name Sanitization
+
+Multiple threats relate to the file-name, thus, it is essential to ensure that the file-name is stored in a secure fashion.
+
+This can be done by creating a **random string** as a file-name. If the file-name is required by the business needs, proper input validation should be done for client-side (_e.g._ active content) and back-end side (_e.g._ special files overwrite or creation) attack vectors. File-name length limits should be taken into consideration based on the system storing the files, as each system has its own file name length limit.
```
nightwatchcyber: Using a randomly generated UUID / GUID maybe a good option here

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: @rbsec I have added a warning in this section, and in the introduction. I won't be adding anything to this. Content Type validation can be conducted fairly easily by almost all web frameworks.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Not necessarily. Some apps only allow file upload -> Write perms. If you give it read, then the users can check the files as well.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: For now, I will keep the small note. It can be tackled in another PR. As I said previously, this CS can handle both. We just need to change the name and make it firm in the ground.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Input validation is referenced across the document. The opposite should be done though properly 😄

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: MIME sniffing occurs on the loading of the file client-side, and not on the upload. I added MSDN and Wikipedia's information to this. I doubt I can cover what you wrote in a CS without repeating what is already mentioned in there.

ThunderSon:
@nightwatchcyber your re-review is highly appreciated 😄

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
mackowski: Should we create an issue for it?

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,153 @@
...
 
-Into web applications, when we expect upload of working documents from users, we can expose the application to submission of documents that we can categorize as *malicious*.
+In short, the following principles should be followed to reach a secure file upload implementation:
 
-We use the term "malicious" here to refer to documents that embed *malicious code* that will be executed when another user (admin, back office operator...) will open the document with the associated application reader.
+- **Whitelist allowed extensions. Only allow safe and critical extensions for business functionality**
+  - **Ensure that [input validation](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Input_Validation_Cheat_Sheet.md#validating-free-form-unicode-text) is applied before validating the extensions.**
+- **Allowed users only can upload files**
+- **Validate the file type, don't trust the [Content-Type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type)  as it can be spoofed**
```
mackowski: In the [File Upload Protection](#file-upload-protection) there no mention of validating the file type/content - you wrote about validating extensions and Content-Type but nothing about actual content of a file (we can recommend here libraries/frameworks that do that)

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,153 @@
...
 
-The most common file types used to transmit *malicious code* into file upload feature are the following:
+- [File Upload Threats](#file-upload-threats)
+    - [Malware Upload](#malware-upload)
+    - [Public File Retrieval](#public-file-retrieval)
+- [File Upload Protection](#file-upload-protection)
+    - [Extension Validation](#extension-validation)
+        - [Whitelist Extensions](#whitelist-extensions)
+        - [Blacklist Extensions](#blacklist-extensions)
+    - [User Permissions](#user-permissions)
```
mackowski: ToC is in the wrong order. For example, in the document, [User Permissions](#user-permissions) are between [File Storage Location](#file-storage-location) and [System Permissions](#system-permissions)

ThunderSon: Yeah, definitely forgot to update it. Will do

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
ThunderSon: Hmm, I'll add to it. I can't sustain that though in the document. It's very difficult. There is a whole wide of file types to cover, lots of languages/frameworks. I'd prefer just putting the note, and if it were to be done, they'd need a person with enough security expertise.

cheatsheets/Protect_FileUpload_Against_Malicious_File.md
```diff
@@ -1,597 +1,163 @@
...
+- **Validate the file type, don't trust the [Content-Type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) as it can be spoofed**
+- **Change the file name to something generated by the application**
+- **Set a file name length limit. Restrict the allowed characters if possible**
+  - **Setting a character limit could deny the attacker from sending null bytes, adding secondary extensions, and performing XSS attacks if the file name is reflected**
+- **Set a file size limit**
+- **Store the files on a different server. If that's not possible, store them in a private directory**
+  - **In the case of public access to the files, ensure that proper permissions are set, file names are safe, and encoded**
+- **Run the file through an antivirus or a sandbox if available to validate that it doesn't contain malicious or bogus data**
+- **Ensure the usage of secure and updated libraries for your needs**
+- **Add CSRF protection, as this is input being served from the user**
```
mackowski: You can add link to CSRF cheat sheet here.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
+# File Upload Cheat Sheet
+
+## Introduction
+
+File upload is becoming more and more an essential part of any application, where the user is able to upload their photo, their CV, or a video showcasing a project they are working on. It is evident that applications should be able to handle file uploads based on the business requirement. In return, the application should be able to fend off bogus, and malicious files in a way to keep the application and the users safe.
```
rbsec: `a more and more essential part`

The `It is evident...requirement` sentence doesn't really seem to say anything.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+In short, the following principles should be followed to reach a secure file upload implementation:
+
+- **Whitelist allowed extensions. Only allow safe and critical extensions for business functionality**
+  - **Ensure that [input validation](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Input_Validation_Cheat_Sheet.md#validating-free-form-unicode-text) is applied before validating the extensions.**
+- **Allowed users only can upload files**
```
rbsec: `Only allow authorised users to upload files`

rbsec: Also maybe move this down, so that all the file type/content points are together?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+- **Validate the file type, don't trust the [Content-Type header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) as it can be spoofed**
+- **Change the file name to something generated by the application**
+- **Set a file name length limit. Restrict the allowed characters if possible**
+  - **Setting a character limit could deny the attacker from sending null bytes, adding secondary extensions, and performing XSS attacks if the file name is reflected**
```
rbsec: This isn't really a recommendation - it's an explanation, but none of the other points have this.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+- **Set a file size limit**
+- **Store the files on a different server. If that's not possible, store them in a private directory**
```
rbsec: It's not clear what `private directory` means - is this `outside of the webroot`?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+  - **In the case of public access to the files, ensure that proper permissions are set, file names are safe, and encoded**
```
rbsec: Usually better to serve through a file download handler so you can do access control/etc

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+- **Run the file through an antivirus or a sandbox if available to validate that it doesn't contain malicious or bogus data**
```
rbsec: Remove `bogus data` - AV cannot test for this

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+- **Ensure the usage of secure and updated libraries for your needs**
```
rbsec: `Ensure that any libraries used are securely configured and kept up to date`

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+- **Add [CSRF protection](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md), as this is input being served from the user**
```
rbsec: I'm not really sure what `as this input is being served from the user means`. I think just `Protect the file upload from against CSRF` would be fine

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+## File Upload Threats
+
+In order to assess and know exactly what controls to implement, knowing what you're facing is essential to protect your assets. The following sections will hopefully showcase the risks accompanying the file upload functionality.
+
+### Malicious Files
+
+The attacker delivers a file for malicious intent, such as:
+
+1. Exploit vulnerabilities in the file parser or processing module (_e.g._ [ImageTrick Exploit](https://imagetragick.com/), [XXE](https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing))
+2. Use the file for phishing (_e.g._ careers form)
```
rbsec: I'm not entirely sure what you're talking about here. Would this be a CV with macros to try and exploit a user (covered in point 5), or a HTML page you upload to steal creds, or a CV with a link to a dodgy site?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+3. Send ZIP bombs, XML bombs (otherwise known as billion laughs attack), or simply huge files in a way to fill the server storage which hinders and damages the server's availability
+4. Overwrite an existing file on the system
+5. Client-side active content that could lead to all types of client-side attacks if the file is publicly retrievable
+6. Make the user upload a file without the user's consent, by crafting a Cross-Site Request Forgery attack (CSRF).
```
rbsec: This isn't really a type of malicious file - I'd remove this point

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Double extensions, _e.g._ `.jpg.php`, where it circumvents easily the regex `\.jpg`
+- Null bytes, _e.g._ `.php%00.jpg`, where `.jpg` gets truncated and `.php` becomes the new extension
+- Generic bad regex that isn't properly tested and well reviewed. Refrain from building your own logic unless you have enough knowledge on this topic.
+
+Refer to the [Input Validation CS](Input_Validation_Cheat_Sheet.md) to properly parse and process the extension.
+
+#### Whitelist Extensions
+
+Ensure the usage of *business-critical* extensions only, without allowing any type of *non-required* extensions. If the system requires:
```
rbsec: Add `For example` before `If the system requires`

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+- image upload, allow one type that is agreed upon to fit the business requirement;
+- cv upload, allow `docx` and `pdf` extensions.
+
+Based on the needs of the application, ensure the **least harmful** and the **lowest risk** file types to be used.
+
+#### Blacklist Extensions
+
+> Blacklisting extensions is a bad idea and is very dangerous. Don't do it unless you have no other choice!
```
rbsec: This syntax looks wrong to me - this should be used for quoting another document, and that's not what you're doing here. I'd just have this as a normal sentence (given there's only two sentences in this section).

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Content-Type Validation
+
+Other than defining the extension of the uploaded file, its MIME-type can be checked for a quick protection against simple file upload attacks.
+
+This can be done preferrably in a whitelist approach; otherwise, this can be done in a blacklist approach.
+
+> This should not be used on its own, as bypassing it is pretty common and easy.
+
+### Magic Number Validation
```
rbsec: `File Signature` would be better than `Magic Number`, which is a bit of a vague term

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+In order to perform this validation, specifying and identifying which patterns that could should be rejected are used in order to protect the service.
+
+### Content-Type Validation
+
+Other than defining the extension of the uploaded file, its MIME-type can be checked for a quick protection against simple file upload attacks.
+
+This can be done preferrably in a whitelist approach; otherwise, this can be done in a blacklist approach.
+
+> This should not be used on its own, as bypassing it is pretty common and easy.
```
rbsec: This should be the first sentence, and should be emphasised. Maybe something along the lines of,

`The Content-Type for uploaded files is provided by the user, and as such cannot be trusted, as it is trivial to spoof. Although it should not be relied upon for security, it provide a quick check to prevent users from unintentionally uploading files with the incorrect type.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Magic Number Validation
+
+In conjunction with [content-type validation](#content-type-validation), validating the file's magic numbers can be checked and verified against the expected file that should be received.
```
rbsec: `the file's signature`

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+> This should not be used on its own, as bypassing it is pretty common and easy.
+
+### File Name Sanitization
+
+Multiple threats relate to the file-name, thus, it is essential to ensure that the file-name is stored in a secure fashion.
```
rbsec: Sentence isn't really saying anything that's not in the next paragraph

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+File-names can endager the system in multiple ways, either by using non acceptable characters, or by using special and restricted filenames, which can be found under the following [MSDN guide](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming-conventions). For a wider overview on different filesystems and how they treat files, refer to [Wikipedia's Filename page](https://en.wikipedia.org/wiki/Filename).
```
rbsec: Make it clear that the MSDN guide is only applicable to Windows

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+In order to avoid the above mentioned threat, creating a **random string** as a file-name, _e.g_ generating a UUID/GUID, is essential. If the file-name is required by the business needs, proper input validation should be done for client-side (_e.g._ active content) and back-end side (_e.g._ special files overwrite or creation) attack vectors. File-name length limits should be taken into consideration based on the system storing the files, as each system has its own file name length limit.
```
rbsec: `such as` would read better than `e.g`

Not clear what an `active content` attack is in a filename. Do you mean things like XSS?

Some more practical guidance could be given here, such as:

- Implement a maximum length
- Restrict characters to alphanumeric, spaces, hyphens and periods (for example)
- Otherwise blacklist dangerous characters (non-printable, slashes, colons, etc) and patterns (two periods)

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### File Content Validation
+
+As mentioned in the [Public File Retrieval](#public-file-retrieval) section, file content can contain unwelcomed data and information.
```
rbsec: Replace `unwelcomed data` with `malicious, inappropriate or illegal contents` or something

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+The File Upload service should allow users to report illegal content, and copyright owners to report abuse.
+
+If there are enough resources, manual file review should be conducted in a sandboxed environment before releasing the files to the public.
+
+Adding some automation to the review could be helpful, which is a harsh process and well studied before its usage. Some services (_e.g._ Virus Total) provide APIs to scan files against well known malicious file hashes. Some frameworks can check and validate the raw content type and validating it against predefined file types, such as in [ASP.NET Drawing Library](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.imaging.imageformat)
```
rbsec: Warn about data leakage to third parties

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### File Storage Location
+
+The location where the files should be stored must be chosen based on security and business requirements. The following points are set by security priority, and are inclusive:
+
+1. Store the files on a **different host**, which allow for complete segragation of duties between the application serving the user, and the host handling file uploads and their storage.
+2. Store the files **outside the webroot**, where only administrative access is allowed.
+3. Store the files **inside the webroot**, and set them in write permissions only.
+   - If read access is required, setting proper controls is a must (_e.g._ internal IP, authorized user, etc.)
+4. Store the files to a **database**, most preferrably as Blobs in a NoSQL database. The database should be chosen based on the application's requirements.
```
rbsec: Why is this considered less secure than storing them in the webroot?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+Adding some automation to the review could be helpful, which is a harsh process and well studied before its usage. Some services (_e.g._ Virus Total) provide APIs to scan files against well known malicious file hashes. Some frameworks can check and validate the raw content type and validating it against predefined file types, such as in [ASP.NET Drawing Library](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.imaging.imageformat)
+
+### File Storage Location
+
+The location where the files should be stored must be chosen based on security and business requirements. The following points are set by security priority, and are inclusive:
+
+1. Store the files on a **different host**, which allow for complete segragation of duties between the application serving the user, and the host handling file uploads and their storage.
+2. Store the files **outside the webroot**, where only administrative access is allowed.
+3. Store the files **inside the webroot**, and set them in write permissions only.
```
rbsec: If there files can't be read, what's the point of uploading them? Do you mean something like `in a folder that is not publicly accessible`?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+   - If read access is required, setting proper controls is a must (_e.g._ internal IP, authorized user, etc.)
+4. Store the files to a **database**, most preferrably as Blobs in a NoSQL database. The database should be chosen based on the application's requirements.
+
+> Some files are emailed on point, or processed and dismissed, without being stored. It is essential to conduct the security measures discussed in this sheet before doing any actions on them.
```
rbsec: `are emailed or processed once they are uploaded, and are not stored on the server`.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### User Permissions
+
+Before any file upload service is accessed, proper validation should occur on two levels for the user uploading a file:
+
+- Authentication level
+  - The user should be a registered user, or an identifiable user, in order to set restrictions and limitations for their upload capabilities
+- Authorization level
+  - Make sure that the user's accesses allow them to upload a file, or multiple files for administrative cases.
```
rbsec: Write is the same style as the previous point - something like `The user should have appropriate permissions...`

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+### System Permissions
```
rbsec: Filesystem Permissions?

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+Files should be stored in a way that ensures:
```
rbsec: Talk about he principle of least privilege (i.e, only the minimum permissions required for the application to function are granted)

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Allowed system users are the only ones capable of reading the files
+- Required modes only are set for the file
+  - If execution is required, scanning the file before running it is required as a security best practice, to ensure that no macros or hidden scripts are available.
+
+### Upload and Download Limits
+
+The application should set proper size limits for the upload service in order to protect the file storage capacity. The file size limit should be considered after file decompression is conducted, and by using secure methods to calculate zip files size. For more on this, see how to [Safely extract files from ZipInputStream](https://wiki.sei.cmu.edu/confluence/display/java/IDS04-J.+Safely+extract+files+from+ZipInputStream), Java's input stream to handle ZIP files.
```
rbsec: Compression is only a factor if your application is going to extract the files

rbsec: Most of the examples given are for Microsoft/.NET, so it seems to to now just give a Java example. I'd either remove this, or replace it with a more link to a more general discussion of the issue.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -0,0 +1,163 @@
...
+
+It should set proper request limits as well for the download service if available to protect the server from DoS attacks.
+
+## Java Code Snippets
```
rbsec: If we're linking to this, are we confident that it is correct and relevant? Looks a bit out of place when we don't give snippets for any other language.

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: What do you mean by that?

rbsec: So having a `download.php?file=someid` is usually better than just having `/uploads/file.docx`, because you can do access control/validation/etc on the handler. Directly accessible files are much harder to do access control for.

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: Point 5 doesn't cover phishing attacks using macros. I will update point 5 to be clearer

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: After doing additional research, I'll modify this point. There are tons of thoughts onto this that we shouldn't actually dig into. I'll mainly just tell the readers to remember that storing files in a DB is possible.
It is more secure because you will require an SQL configuration issue to happen, which is harder to happen versus a permissions issue on the filesystem. It kind of relieves security issues, but adds burden on the DBA to properly handle the file storage and how to actually separate them from the rest of the DB tables.

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: We can't cover everything, yet it doesn't hurt to actually have a certain language safe solution. Developers can understand what's happening and actually replicate that behavior. There are not so many public discussions on file safety

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: No. This point sometimes happen when:
- User has only access to the webroot, thus file storage occurs on the webroot.
- The files need to be accessed by a more administrative user (_e.g._ wordpress admin, or drupal admin). Setting them outside the webroot removes them from the admin's reach.

cheatsheets/File_Upload_Cheat_Sheet.md
ThunderSon: @righettod do think that there are more well known best practices for implementing your practices? I can agree that the code written is not battle tested. It is a way to do it though, and shouldn't hurt, but to follow our best practice, your input is highly appreciated.

jmanico:
Does this protect vs a ZIP bomb where a 42k file unzips to many Petabytes?

I typically suggest keep AWAY from zip files during uploading....

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

ThunderSon:
Yes, it validates against ZIP bombs. It sets limits both to the file size, the number of files, and the size of the extracted files. It throws an exception once any of those go off. I am more interested in readers seeing some practices to handling files, more than tackling specifically ZIP files, which is the case in here. The referenced site is one of the most mature in code reviews. I doubt there's a better repository of best code practices than that one ... Would love to hear your thoughts on this. Only throwing banter `Know how to handle your files` is a bit too vague and actually is not such a good advice.

jmanico:
Images:  use image rewriting to verify
Office: use Apache POI or similar to validate
ZIP: what you are doing!

That’s a start...

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

cheatsheets/File_Upload_Cheat_Sheet.md
righettod: At the time, I have implemented it to address a client issue with office document / image / pdf and then I have proposed a CS about it to share the code in case in which it can help someone too. I have done that because I did not found a library doing the job.

The code provided do the job for the cases mentioned and is currently limited to this. It's up to you to decide to ignore it if you decide that is not relevant. I propose it as sample fixation in my web assessment for the same cases.

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -117,6 +117,12 @@ In order to avoid the above mentioned threat, creating a **random string** as a
 
 As mentioned in the [Public File Retrieval](#public-file-retrieval) section, file content can contain malicious, inappropriate, or illegal data.
 
+Based on the expected type, special file content validation can be applied:
+
+- For **images**, applying image rewriting techniques destroys any kind of malicious content injected in an image; this could be done through [randomization](https://security.stackexchange.com/a/8625/118367).
```
mackowski: Personally I do not like links to stackexchange/stackoverflow. I would prefer to describe `randomization` inside cheat sheet.
@rbsec @ThunderSon any thoughts about this?

ThunderSon: As far as I know, answers are never removed, they have a strong history of changes, and the answer provided good. If Robin agrees with you I'll definitely add them.

rbsec: I don't have  problem with links to stackexchange - and given this is a fairly minor point in the CS I'm not sure it's worth devoting a whole paragraph to trying to explain it.

PR #327, opened 2019-12-24
Create some style guidelines

rbsec:
Following some discussion on Slack, I've made some initial style/content guidelines. Seemed more sensible to add them to the existing Contribution guidelines than have them in a separate document.

I'm in favour of a fairly minimal style guide (that can grow as we progress) in order to increase the chances that people actually read it.

Are there any key things I'm missing? This doesn't have to be perfect - getting some initial guidelines in place that can grown over time is a probably a better approach.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,51 @@ Follow these steps:
     3. Submit your [Pull Request](https://help.github.com/articles/creating-a-pull-request/).
     4. Verify that the CI job [applied on your Pull Request](https://travis-ci.org/OWASP/CheatSheetSeries/pull_requests) do not fail!
 
+# Content Guidelines
+
+The intended audience of the cheat sheets is developers, _not_ security experts. As such, do not assume that the person reading the cheat sheet has a strong understanding of security topics. In depth or academic discussions are generally not appropriate in cheat sheets, and should be linked to as external references where appropriate.
+
+The purpose of the cheat sheets is to provide **useful, practical advice** that can be followed by developers. It is much better to give _good_ practices that can actually be followed than _best_ practices that are completely impractical
+
+The submitting changes in a PR, consider the following area:
+
+- The content should be useful to developers.
+- The contents should be factual and correct.
+- Statements should be supported by authoritative references where possible.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -19,6 +19,51 @@ Follow these steps:
...
+- Statements should be supported by authoritative references where possible.
+- Recommendations should be feasible for the majority of developers to implement.
+
+# Style Guidelines
+
+## Cheat Sheet Structure
+
+- Start with a H1 of the cheat sheet name
+- Generally the first two sections should be `Introduction` and `Contents`, followed by the body of the cheat sheet.
+- Headings should have a blank line after them.
```
ThunderSon: Part of the linter. Not a style guide per se.

CONTRIBUTING.md
```diff
@@ -19,6 +19,51 @@ Follow these steps:
     3. Submit your [Pull Request](https://help.github.com/articles/creating-a-pull-request/).
     4. Verify that the CI job [applied on your Pull Request](https://travis-ci.org/OWASP/CheatSheetSeries/pull_requests) do not fail!
 
+# Content Guidelines
+
+The intended audience of the cheat sheets is developers, _not_ security experts. As such, do not assume that the person reading the cheat sheet has a strong understanding of security topics. In depth or academic discussions are generally not appropriate in cheat sheets, and should be linked to as external references where appropriate.
+
+The purpose of the cheat sheets is to provide **useful, practical advice** that can be followed by developers. It is much better to give _good_ practices that can actually be followed than _best_ practices that are completely impractical
+
+The submitting changes in a PR, consider the following area:
```
ThunderSon: When instead of The?

CONTRIBUTING.md
rbsec: It is checked by the linter, but telling people to do this here should hopefully mean that they do it right the first time (otherwise they commit, linter fails the build, and they have to fix it in a second commit).

CONTRIBUTING.md
rbsec: Fixed in a3a3978.

PR #334, opened 2020-01-06
Grammar Cleanup

dhurlburtusa:

I checked the entire file.

There were many place where an independent clause was connected with a dependent clause using a comma and a conjunction (e.g. `and`, `so`, etc). This can be fixed either by removing the comma or changing the dependent clause into an independent clause. I chose to use the former solution in most cases.


mackowski:
Thanks @dhurlburtusa for PR!

PR #336, opened 2020-01-08
Fix #321

### LINKED ISSUE #321: Update: XML_External_Entity_Prevention_Cheat_Sheet.md
Hello,
I'm working at Sonarsource and we try to improve our products by raising awareness for developers when they are using libraries in such a way that could make their code vulnerable.

In the section C/C++, subsection concerning libxerces-c, about the use of `XercesDOMParser`, it is written that in order to prevent XXE, `setCreateEntityReferenceNodes` should be called with `false` argument.
Correct me if I'm wrong but it seems to me that creating entity reference nodes will prevent the entity to be expanded and therefore will prevent XXE, no?

Just replacing
`parser->setCreateEntityReferenceNodes(false);`
by
`parser->setCreateEntityReferenceNodes(true);`

amelie-renard-sonarsource:
This PR covers issue #321 .

### DIFF
--- a/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
@@ -56,8 +56,14 @@ Use of `XercesDOMParser` do this to prevent XXE:
 
 ``` cpp
 XercesDOMParser *parser = new XercesDOMParser;
-parser->setCreateEntityReferenceNodes(false);
+parser->setCreateEntityReferenceNodes(true);
 ```
+Creating reference nodes prevents XXE coming from DTDs, local file inclusions as well as outbound http requests.
+NB: `true` is the default value so it is not mandatory to call this method.
+
+To go a bit further,
+- `setLoadExternalDTD` only applies to DTDs so this method does not fully protect against XXE. In addition to this, the behavior set by this method is ignored if the validation scheme (`setValidationScheme`) is equal to `XercesDOMParser::Val_Auto` or `XercesDOMParser::Val_Always`.
+- It is possible to implement and set to the parser your own entity resolver. In this case, be careful with what you do to not become vulnerable to XXE and be sure to call `parser->setDisableDefaultEntityResolution(true)` so that the parser will not try to resolve the external entity on its own if your customized entity resolver did not succeed.
 
 Use of SAXParser, do this to prevent XXE:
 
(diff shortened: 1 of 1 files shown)


cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
```diff
@@ -56,9 +56,17 @@ Use of `XercesDOMParser` do this to prevent XXE:
...
+parser->setCreateEntityReferenceNodes(true);
 ```
 
+Creating reference nodes prevents XXE coming from DTDs, local file inclusions as well as outbound http requests.
+
+NB: `true` is the default value so it is not mandatory to call this method.
+
+To go a bit further,
+- `setLoadExternalDTD` only applies to DTDs so this method does not fully protect against XXE. In addition to this, the behavior set by this method is ignored if the validation scheme (`setValidationScheme`) is equal to `XercesDOMParser::Val_Auto` or `XercesDOMParser::Val_Always`.
+- It is possible to implement and set to the parser your own entity resolver. In this case, be careful with what you do to not become vulnerable to XXE and be sure to call `parser->setDisableDefaultEntityResolution(true)` so that the parser will not try to resolve the external entity on its own if your customized entity resolver did not succeed.
```
rbsec: I think that `setDisableDefaultEntityResolution(true)` should probably be part of the main recommendation. Without it you still make outbound HTTP requests from an `<!ENTITY` tag, which is not something that should be allowed by default.

cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
```diff
@@ -56,9 +56,17 @@ Use of `XercesDOMParser` do this to prevent XXE:
 
 ``` cpp
 XercesDOMParser *parser = new XercesDOMParser;
-parser->setCreateEntityReferenceNodes(false);
+parser->setCreateEntityReferenceNodes(true);
 ```
 
+Creating reference nodes prevents XXE coming from DTDs, local file inclusions as well as outbound http requests.
```
rbsec: Having just this does allow outbound HTTP - you need `setDisableDefaultEntityResolution(true)` as well to block it.

ThunderSon:
Hi @amelie-renard-sonarsource ! Any updates from your side on this?

amelie-renard-sonarsource:
Hi @ThunderSon and @rbsec,
Sorry for my late reply.
The thing is that I am not really comfortable to do these changes because, as I said in the issue #321, I did not succeed in using the method `setDisableDefaultEntityResolution` (and I was under the impression that this method was related to the usere-definition of an EntityResolver....)
Unfortunately I am on sick leave for a while (thus my late reply) and won't be able to investigate more...

rbsec:
@amelie-renard-sonarsource I hope you have a speedy recovery.

I'm going to close this PR for now so it's not hanging open and create a new PR with this change, and also the `setDisableDefaultEntityResolution(true)` which seems prevents outbound HTTP in my testing. I'll try and get the other guys (or anyone else who's available) to to do some testing to make sure that this is a complete fix - but even if it's not it'll be better than the current recommendation.

Thanks for raising this issue - it's clearly been wrong in the CS for a while, so is definitely one that needs addressing.

PR #337, opened 2020-01-08
Fix Grammar, Readability, and Other Changes

dhurlburtusa:

I checked the entire document.

Most of this document looks like it was written by a non-native English speaker. Some issues were minor grammar issues, but some sentences just didn't make sense. I did my best to interpret what the author meant. So, please review for accuracy.

Some other changes include fixing spelling errors and formatting including indentation and title-case in headings.

### DIFF
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
@@ -65,7 +65,7 @@ If your application does not need to be fully stateless, you can consider using 
 
 # Issues
 
-## `none` Hashing Algorithm
+## None Hashing Algorithm
 
 ### Symptom
 
@@ -101,14 +101,14 @@ This attack occurs when a token has been intercepted/stolen by an attacker and t
 
 ### How to Prevent
 
-A way to prevent it is to add an "user context" in the token. An user context will be composed of the following information:
+A way to prevent it is to add a "user context" in the token. A user context will be composed of the following information:
 
 - A random string that will be generated during the authentication phase. It will be sent to the client as an hardened cookie (flags: [HttpOnly + Secure](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies) + [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) + [cookie prefixes](https://googlechrome.github.io/samples/cookie-prefixes/)).
 - A SHA256 hash of the random string will be stored in the token (instead of the raw value) in order to prevent any XSS issues allowing the attacker to read the random string value and setting the expected cookie.
 
-IP address must not be used because there are some legitimate situations in which the IP address can change during the same session. For example, when an user accesses an application through their mobile device and the mobile operator changes during the exchange, then the IP address may (often) change. Moreover, using the IP address can potentially cause issues with [European GDPR](http://www.eugdpr.org/) compliancy.
+IP addresses should not be used because there are some legitimate situations in which the IP address can change during the same session. For example, when an user accesses an application through their mobile device and the mobile operator changes during the exchange, then the IP address may (often) change. Moreover, using the IP address can potentially cause issues with [European GDPR](http://www.eugdpr.org/) compliancy.
 
-During token validation, if the received token does not contain the right context, e.g., it is replayed, then it must be rejected.
+During token validation, if the received token does not contain the right context (for example, if it has been replayed), then it must be rejected.
 
 ### Implementation example
 
@@ -309,7 +309,7 @@ public class TokenRevoker {
 
 ### Symptom
 
-This attack occurs when an attacker has access to a token (or a set of tokens) and extracts information stored in it (JWT token information are base64 encoded at the basis) in order to obtain information about the system. Information can be for example the security roles, login format...
+This attack occurs when an attacker has access to a token (or a set of tokens) and extracts information stored in it (the contents of JWT tokens are base64 encoded, but is not encrypted by default) in order to obtain information about the system. Information can be for example the security roles, login format...
 
 ### How to Prevent
 
@@ -464,7 +464,6 @@ This occurs when an application stores the token in a manner exhibiting the foll
 ### How to Prevent
 
 1.  Store the token using the browser *sessionStorage* container.
-1.  Avoid storing token in a cookie. Use a hardened cookie if you must.
 1.  Add it as a *Bearer* HTTP `Authentication` header with JavaScript when calling services.
 1.  Add [fingerprint](JSON_Web_Token_Cheat_Sheet_for_Java.md#token-sidejacking) information to the token.
 
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
@@ -4,13 +4,13 @@ Many applications use **JSON Web Tokens** (JWT) to allow the client to indicate 
 
 From [JWT.IO](https://jwt.io/introduction):
 
-*JSON Web Token (JWT) is an open standard (RFC 7519) that defines a compact and self-contained way for securely transmitting information between parties as a JSON object. This information can be verified and trusted because it is digitally signed. JWTs can be signed using a secret (with the HMAC algorithm) or a public/private key pair using RSA.*
+> JSON Web Token (JWT) is an open standard (RFC 7519) that defines a compact and self-contained way for securely transmitting information between parties as a JSON object. This information can be verified and trusted because it is digitally signed. JWTs can be signed using a secret (with the HMAC algorithm) or a public/private key pair using RSA.
 
-JSON Web Token is used to carry information related to the identity and characteristics (claims) of a client. This "container" is signed by the server in order to avoid that a client tamper it in order to change, for example, the identity or any characteristics (example: change the role from simple user to admin or change the client login).
+JSON Web Token is used to carry information related to the identity and characteristics (claims) of a client. This information is signed by the server in order for it to detect whether it was tampered with after sending it to the client. This will prevent an attacker from changing the identity or any characteristics (for example, changing the role from simple user to admin or change the client login).
 
-This token is created during authentication (is provided in case of successful authentication) and is verified by the server before any processing. It is used by an application to allow a client to present a token representing his "identity card" (container with all information about him) to server and allow the server to verify the validity and integrity of the token in a secure way, all of this in a stateless and portable approach (portable in the way that client and server technologies can be different including also the transport channel even if HTTP is the most often used).
+This token is created during authentication (is provided in case of successful authentication) and is verified by the server before any processing. It is used by an application to allow a client to present a token representing the user's "identity card" to the server and allow the server to verify the validity and integrity of the token in a secure way, all of this in a stateless and portable approach (portable in the way that client and server technologies can be different including also the transport channel even if HTTP is the most often used).
 
-# Token structure
+# Token Structure
 
 Token structure example taken from [JWT.IO](https://jwt.io/#debugger):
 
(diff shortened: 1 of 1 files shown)


cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -55,29 +55,29 @@ The tips presented in this article are part of a Java project that was created t
...
 
 Even if a JWT token is "easy" to use and allow to expose services (mostly REST style) in a stateless way, it's not the solution that fits for all applications because it comes with some caveats, like for example the question of the storage of the token (tackled in this cheatsheet) and others...
 
-If your application does not need to be fully stateless, you can consider using traditional session system provided by all web frameworks and follow the advice from the dedicated [cheatsheet](Session_Management_Cheat_Sheet.md). However, for stateless applications, when well implemented, it's a good candidate.
+If your application does not need to be fully stateless, you can consider using traditional session system provided by all web frameworks and follow the advice from the dedicated [session management cheat sheet](Session_Management_Cheat_Sheet.md). However, for stateless applications, when well implemented, it's a good candidate.
 
 # Issues
 
-## NONE hashing algorithm
+## `none` Hashing Algorithm
```
rbsec: This looks very strange in a title - I think it'd be better as either the original NONE, or None.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -55,29 +55,29 @@ The tips presented in this article are part of a Java project that was created t
 
 You can find the Java project [here](https://github.com/righettod/poc-jwt), it uses the official [JWT library](https://jwt.io/#libraries).
 
-In the rest of the article, the term **token** refer to the **JSON Web Tokens** (JWT).
+In the rest of the article, the term **token** refers to the **JSON Web Tokens** (JWT).
 
-# Consideration about using JWT
+# Consideration about Using JWT
```
rbsec: "About" should probably be capitalised for consistency

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -93,26 +93,26 @@ JWTVerifier verifier = JWT.require(Algorithm.HMAC256(keyHMAC)).build();
...
 ### Symptom
 
-This attack occur when a token has been intercepted/stolen by a attacker and this one use it to gain access to the system using targeted user identity.
+This attack occurs when a token has been intercepted/stolen by an attacker and they use it to gain access to the system using targeted user identity.
 
-### How to prevent
+### How to Prevent
 
-A way to protect is to add "user context" in the token. User context will be composed by the following information:
+A way to prevent it is to add an "user context" in the token. An user context will be composed of the following information:
```
rbsec: This should be `a "user context"` rather than `an "user context"`. Same for the next sentence, should be `a user context`

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -93,26 +93,26 @@ JWTVerifier verifier = JWT.require(Algorithm.HMAC256(keyHMAC)).build();
...
+- A SHA256 hash of the random string will be stored in the token (instead of the raw value) in order to prevent any XSS issues allowing the attacker to read the random string value and setting the expected cookie.
 
-IP address will not be used because there some situation in which IP address can change during the same session like for example when a user access an application through his mobile and he change of mobile operator during the exchange then he change legitimately (often) is IP address. Moreover, using IP address can potentially cause issue at [European GDPR](http://www.eugdpr.org/) compliance level.
+IP address must not be used because there are some legitimate situations in which the IP address can change during the same session. For example, when an user accesses an application through their mobile device and the mobile operator changes during the exchange, then the IP address may (often) change. Moreover, using the IP address can potentially cause issues with [European GDPR](http://www.eugdpr.org/) compliancy.
```
rbsec: `IP Addresses` rather than `IP Address`.

I think this should be softened a bit to `should not be used` - there are times where it's a completely valid thing to do (such as an internal application).

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -93,26 +93,26 @@ JWTVerifier verifier = JWT.require(Algorithm.HMAC256(keyHMAC)).build();
...
 
-During token validation, if the received token do not contains the right context so, it is replayed and then it must be rejected.
+During token validation, if the received token does not contain the right context, e.g., it is replayed, then it must be rejected.
```
rbsec: This sentence is better than it was, but still feels a bit clunky. How about:

`...the right context (for example if it has been replayed), then it must be rejected.`

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -305,19 +305,19 @@ public class TokenRevoker {
  }
 ```
 
-## Token information disclosure
+## Token Information Disclosure
 
 ### Symptom
 
-This attack occur when a attacker access to a token (or a set of tokens) and extract information stored into it (JWT token information are base64 encoded at the basis) in order to obtains information about the system. Information can be for example the security roles, login format...
+This attack occurs when an attacker has access to a token (or a set of tokens) and extracts information stored in it (JWT token information are base64 encoded at the basis) in order to obtain information about the system. Information can be for example the security roles, login format...
```
rbsec: I'm not sure what `at the basis` is meant to mean here. I think the brackets should be something like:

`(the contents of JWT tokens is base64 encoded, but is not encrypted by default)`

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -461,90 +461,92 @@ This occurs when an application stores the token in a manner exhibiting the foll
...
-             sessionStorage.removeItem("token");
-         }
-     })
-     .fail(function(jqXHR, textStatus, error){
-             ...
-         sessionStorage.removeItem("token");
-     });
- }
+/* Handle request for JWT token and local storage*/
+function authenticate() {
```
rbsec: Why was this function renamed?

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
```diff
@@ -461,90 +461,92 @@ This occurs when an application stores the token in a manner exhibiting the foll
 - Retrieved even if the browser is restarted (Use of browser *localStorage* container).
 - Retrieved in case of [XSS](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) issue (Cookie accessible to JavaScript code or Token stored in browser local/session storage).
 
-### How to prevent
+### How to Prevent
 
 1.  Store the token using the browser *sessionStorage* container.
-2.  Add it as a *Bearer* with JavaScript when calling services.
-3.  Add [fingerprint](JSON_Web_Token_Cheat_Sheet_for_Java.md#token-sidejacking) information to the token.
+1.  Avoid storing token in a cookie. Use a hardened cookie if you must.
```
rbsec: Where has this recommendation come from? Most of the changes are stylistic, but this seems to be giving a new recommendation, and not necessarily one that I'd agree with. A properly secured cookie has security benefits over sessionStorage (such as the fact it can be HttpOnly) - so I think this point should probably discussed as a separate issue.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
dhurlburtusa: "about" is a preposition. In the rules I learned for title casing, prepositions should be lowercase. See https://capitalizemytitle.com/ for list of rules for title casing.

Do you still think it should be capitalized? I have no strong preference either way.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
dhurlburtusa: Only because the original `getToken` implies that a token (or a Promise that resolves to a token) is going to be returned. The call to `$.post` is making a request to the `authenticate` service.

I think the new name, `authenticate`, works better with what the example code is doing.

Please let me know if you still think we should revert it back or use a different name.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
dhurlburtusa: I added it because I thought storing tokens in a cookie aids CSRF.
I removed it.
I'd be interested in getting a better understanding why it is safe to store the token in a cookie.
BTW, I am not a security expert (working toward it, hence the reason I am reading the cheat sheets).

dhurlburtusa:
@rbsec Thanks for reviewing my changes. It is very much appreciated. I made several updates based on your review. I replied to some of the comments. Please let me know how to proceed.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
kingthorin: Just to add my two cents. Leaving it lower seems to be 'technically' correct based on the tool I normally use. https://capitalizemytitle.com/#Chicago

It does look out of place, I think due to the length of the word.

As long as the style is consistently adhered to I think that's the main thing. https://grammar.yourdictionary.com/capitalization/rules-for-capitalization-in-titles.html seems to confirm that for Chicago style prepositions should be lower.

rbsec: @kingthorin I think I had exactly the same thought process as you - it looks _off_ based on the length of the word. If it's technically correct (according to some guide - and we don't have a consistent rule about which styles we use) then that's fine.

We've all got better things to do that argue about obscure grammar rules, so happy for this to stay as it is.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
rbsec: The danger with storing a JWT (or session ID, for that matter) in a cookie is that it _can_ lead to CSRF (as cookies are normally sent automatically). However, if you mark the cookie as [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) (which browsers will start doing automatically in the coming months), then it won't be included in cross-domain POST requests, so CSRF won't work unless you can do it from inside the same domain.

The benefit of having it in a cookie is you can mark is as Secure so it's only sent over HTTPS, and HttpOnly so it's not accessible from JavaScript (so can't be stolen with XSS) - there's no equivalent protection for SessionStorage.

The most obvious way to exploit that cookie would be use to XSS in the domain to trigger a request with the cookie (and also to obtain and bypass the CSRF token) - but if you can do this you could trivially read a cookie from SessionStorage anyway.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
rbsec: Ok, that seems reasonable, and the function isn't referenced anywhere else, so it's fine.

cheatsheets/JSON_Web_Token_Cheat_Sheet_for_Java.md
dhurlburtusa: @rbsec Thanks for the explanation. Much appreciated.

dhurlburtusa:
@rbsec You're welcome! And thank you OWASP team for this project.

mackowski:
Thanks @dhurlburtusa for PR!

ThunderSon:
@dhurlburtusa Would you be able to send me a twitter message on my twitter account [`@7hunderson`](https://twitter.com/7hunderSon)? I think we could help each other bond the developer in you with the security in us 😄

dhurlburtusa:
@ThunderSon I don't have a Twitter account. Is there some other way we can connect? Is there a Slack workspace or something else?

ThunderSon:
Definitely! Please, join using this [invitation](https://join.slack.com/t/owasp/shared_invite/enQtNjExMTc3MTg0MzU4LWQ2Nzg3NGJiZGQ2MjRmNzkzN2Q4YzU1MWYyZTdjYjA2ZTA5M2RkNzE2ZjdkNzI5ZThhOWY5MjljYWZmYmY4ZjM), then proceed to join the [project's channel](https://owasp.slack.com/archives/C073YNUQG) 😄

PR #339, opened 2020-01-10
Updates to Ruby on Rails Cheatsheet 

ankane:
Hey, thanks for this great project. This PR includes a number of updates to the Ruby on Rails cheatsheet:

- Fixed command injection examples - `system` is not capitalized, and `exec` doesn't require `Kernel` prefix, as it's required in all objects by default
- Updated indentation in some examples (some used 2 spaces, others 4 - Ruby convention is 2)
- Removed Rails 2.x examples (has been EOL for a long time)
- Changed URLs to https where possible

### DIFF
--- a/cheatsheets/Ruby_on_Rails_Cheatsheet.md
+++ b/cheatsheets/Ruby_on_Rails_Cheatsheet.md
@@ -1,6 +1,6 @@
 # Introduction
 
-This *Cheatsheet* intends to provide quick basic Ruby on Rails security tips for developers. It complements, augments or emphasizes points brought up in the [rails security guide](http://guides.rubyonrails.org/security.html) from rails core. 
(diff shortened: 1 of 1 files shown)


ankane:
Yeah, I can't recall the last time I saw a reference to Rails 2.x (5 or more years ago, if not longer). Apps should be on Rails 5.2 or Rails 6.0, which are the only two versions that receive security updates, but some apps are likely on Rails 4.2 still. The Rails 3.x pattern (`Project.where(...)`) is still used in the most recent version of Rails.

ThunderSon:
@ankane Are you part of OWASP Slack? If not, would you like to join us? We could benefit from your development experience, and maybe give back here and there with our security knowledge? 😄 
If you're not part of slack and you'd like to merge efforts, please message me on my twitter account [`@7hunderSon`](https://twitter.com/7hunderSon)

Issue #340, opened 2020-01-11
Update the CI/CD Flow


ThunderSon:
**What is wrong?**
We are dependent on 3 CI/CD flows. Travis, Circle, and GH actions.

**What should be done?**
Merge the required tasks and set them in GH actions.

- [x] Old issues + PRs migration to GH-actions.
- [x] Website publishing migration to GH-actions.
- [x] TravisCI clean up. GH-actions markdown lint already set-up.
- [ ] markdown link check GH-actions
- [ ] markdown lint GH-actions
- [ ] gh-pages implemention in GH-actions

rejahrehim:
Let me work on this Issue

mackowski:
@rejahrehim is there any more work to do?

mackowski:
I see that this was addressed some time ago

PR #341, opened 2020-01-14
Migrate cron (send old issues and PR details to slack) to GitHub actions

rejahrehim:
This PR is part of  issue #340

- [x] The CI build of your PR pass

@ThunderSon , Could you please add a new secret `SLACK_WEBHOOK` with the incoming webhook to the current slack channel to recive the notifications

### DIFF
--- a/.github/workflows/identify-old-issues-and-pr.yml
+++ b/.github/workflows/identify-old-issues-and-pr.yml
@@ -2,7 +2,7 @@ name: Identify old issues and PR
 
 on:
   schedule:
-    - cron: "* * * * *"
+    - cron: "*/5 * * * *"
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/identify-old-issues-and-pr.yml
```diff
@@ -0,0 +1,29 @@
+name: Identify old issues and PR
```
ThunderSon: I don't think that we still need the python script. I found a stale github actions, but I'd prefer not implementing it yet. There are issues I don't want closed. What would you recommend?

ThunderSon:
Webhook added. To anyone interested in how the hook link was created, you can check the following [incoming webhooks integration](https://slack.com/apps/A0F7XDUAZ-incoming-webhooks)

.github/workflows/identify-old-issues-and-pr.yml
rejahrehim: Yup, We can migrate the python scripts in the next round. I prefer one at a time.

ThunderSon: This shall be tackled in a different PR

PR #354, opened 2020-02-05
Added Kubescan to Rule 9

KellyGriffin:

- [ ] All the markdown files do not raise any validation policy violation, see the [policy](https://github.com/OWASP/CheatSheetSeries#editor--validation-policy).
- [ ] All the markdown files follow these [format rules](https://github.com/OWASP/CheatSheetSeries#conversion-rules).

### DIFF
--- a/cheatsheets/Docker_Security_Cheat_Sheet.md
+++ b/cheatsheets/Docker_Security_Cheat_Sheet.md
@@ -230,6 +230,7 @@ To detect misconfigurations in Kubernetes:
  - [kubeaudit](https://github.com/Shopify/kubeaudit)
  - [kubesec.io](https://kubesec.io/)
  - [kube-bench](https://github.com/aquasecurity/kube-bench)
+ - [kubescan](https://github.com/octarinesec/kube-scan)
 
 To detect misconfigurations in Docker:
  - [inspec.io](https://www.inspec.io/docs/reference/resources/docker/)


rbsec:
Hi @KellyGriffin - thanks for the PR.

This tool looks to be fairly new and still under quite active development - so we need to consider whether it's mature and stable enough to be recommending in this CS. It's not a tool that I've used personally. @ThunderSon or @mackowski - have either of you come across this?

mackowski:
I would love to have a comparison between tools. I am not against new tools but we should have a reason when we recommend more than two tools. Why is it better, how it is different from other tool etc.
Something simillar to what [trivy](https://github.com/aquasecurity/trivy#comparison-with-other-scanners) have on their github.

@rbsec what do you think?

@KellyGriffin are you able to create such comparison?

rbsec:
I wouldn't be against a brief summary - but you could probably write an entire article comparing these types of tools, which isn't really what that cheat sheets about.

KellyGriffin:
@mackowski @rbsec essentially the reason I added Kube-scan on there, whilst agree its still very new, has a different use case and outcome.  Kube-scan focuses on showcasing Kubernetes Common Configuration Scoring System which is very similar to CVSS which gives ratings, etc.

mackowski:
@KellyGriffin we decided that currently we cannot recommend kube-scan because other tools are more matured and battle proved. 

This cheatsheet is for Docker containers and not for k8s. kuba-scan adds a value (e.g. Kubernetes Common Configuration Scoring System) and a lot of checks against k8s configuration and it might be a good fit for k8s CS (after such CS will be created).

PR #374, opened 2020-03-17
Forgot Password CS Rewrite

### LINKED ISSUE #159: Cheat sheet update/refactor proposal: Forgot Password Cheat Sheet
### 1. What is missing or needs to be updated in the current iteration of the cheat sheet?

On step 2, the first paragraph preconize a solution which is vulnerable to user enumeration and thus unfortunately not secured:

"After the form on Step 1 is submitted, the application verifies that each piece of data is correct for the given username. If anything is incorrect, or **if the username is not recognized, the second page displays a generic error message such as Sorry, invalid data**."

### 2. How do you intend to resolve this with an update/refactoring?

This issue implies to rethink the process described in this cheat sheet, as I do not see a way to just rephrase it.

ThunderSon:
This PR closes #159 .

The forgot password CS focused heavily on the security questions.

In this rewrite, the CS focuses on:
- multiple methods that could be used for the service
- how to implement each method
- common tasks to be conducted for all methods of forgot password


cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,70 @@
...
 
-- The transport mechanism used to send the password (email, SMS, etc) may not be encrypted.
-- The password may be stored in the users' mailbox or device for an indefinite period.
-- The password may be accessible to IT staff or other users who have access to the mailbox or device.
-- The password may be logged by third parties such as mail security scanning platforms.
+> OTPs can be sent through other channels as well, such as emails and SMSs. Various attacks and weaknesses have been identified in such technologies that it is preferrable not to use them for OTPs.
 
-Of course, by all means, once a user's password has been reset, the randomly-generated token should no longer be valid.
+### URL Tokens
```
rbsec: Needs some explanation about properties of tokens:

- Randomly generated (not hash of email or something stupid)
- Linked to specific user in DB
- Time limited
- Single use

rbsec: Also, I've seen some times using JWT for password reset. It provides a nice stateless way to have expiration dates without a backend DB, but has all the usual risks of JWT.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,70 @@
...
+After sending the URL token to the user through a side-channel, _e.g._ through email, the user will:
 
-The reset must be performed before any other operations can be performed by the user.
+1. Access the URL with the attached token.
+2. Set the token in a cookie (recommended), or the client-storage mechanism that your application uses.
+3. Remove the token from the URL and redirect the user to the password reset service. This ensures that protection against [referer leakage](https://portswigger.net/kb/issues/00500400_cross-domain-referer-leakage).
+4. Let the user create a new password and confirm it. Ensure that the password policy is applied.
 
-## Step 5) Logging
+### Security Questions
```
rbsec: Make it clear that these shouldn't be used in isolation

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,70 @@
...
+## Methods Implementation
 
-Because users' security questions / answers generally contains much less entropy than a well-chosen password (how many likely answers are there to the typical *What's your favorite sports team?* or *In what city where you born?* security questions anyway?), make sure you limit the number of guesses attempted and if some threshold is exceeded for that user (say 3 to 5), lock out the user's account for some reasonable duration (say at least 5 minutes) and then challenge the user with some form of challenge token per standard multi-factor workflow (see \#3, below) to mitigate attempts by hackers to guess the questions and reset the user's password. It is not unreasonable to think that a user's email account may have already been compromised, so tokens that do not involve email, such as SMS or a mobile soft-token, are best.
+In order to implement the forgot password service, the developer needs to choose one of the proposed [methods](#methods).
 
-## Step 3) Send a Token Over a Side-Channel
+### One Time Password
```
rbsec: Maybe just link to MFA CS rather than long explanation.

Also need to be clear how this is used - are you saying that you implement MFA, but only use it for password resets? That sounds...sketchy

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,70 @@
-# Introduction
+# Forgot Password Cheat Sheet
```
rbsec: Would it make sense to try and cover forgotten/lost OTP tokens as well?

ThunderSon:
The file name will be changed once everything is accepted and agreed upon.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+# Forgot Password and MFA Cheat Sheet
 
-This article provides a simple model to follow when implementing a *forgot password* web application feature.
+## Introduction
 
-# The Problem
+In order to implement a proper user management system, almost all systems integrate a **Forgot Password/MFA?** service, which allows the user to request a reset to their password or MFA whenever they forget any of them, or if their account ever gets breached.
 
-There is no industry standard for implementing a **Forgot Password** feature. The result is that you see applications forcing users to jump through myriad hoops involving emails, special URLs, temporary passwords, personal security questions, and so on. In the end you have to reset it to a new value.
+As much as is this functionality looks straight-forward and easy to implement, the details of its implementation makes it a sweet spot for security attacks, such as the renowned [user enumeration attack](https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account.html).
```
bigshebang, line 7: There's an extra word at the beginning: "is". This should instead read

> As much as this functionality...

jmanico:
The industry standard is for forgot password is to involve a multi factor challenge for high risk websites. It’s part of banking and NIST standards. Especially NIST SP800-63.

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+    + [URL Tokens](#url-tokens)
+    + [Security Questions](#security-questions)
+    + [Backup Codes](#backup-codes)
+  * [Operational Tasks](#operational-tasks)
 
-At a minimum, you should have collected some data that will allow you to send the password reset information to some out-of-band side-channel, such as a (possibly different) email address or an SMS text number, etc. to be used in Step 3.
+## Methods
 
-## Step 2) Verify Security Questions
+In order to allow a user to request a password or MFA reset, you will need to have some way to identify the user, or a mean to reach out to them through a side-channel.
```
bigshebang, line 31: "mean" here should be "means". As in the phrase "[a means to an end](https://www.merriam-webster.com/dictionary/a%20means%20to%20an%20end)".

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+This can be done through any of the following methods:
 
-If all submitted data is correct, Step 2 should display at least two of the user's pre-established personal security questions, along with input fields for the answers. It's important that the answer fields are part of a single HTML form.
+- One Time Password (OTP).
+- URL tokens.
+- Security Questions.
+- Offline backup codes.
 
-Do not provide a drop-down list for the user to select the questions he wants to answer. Avoid sending the username as a parameter (hidden or otherwise) when the form on this page is submitted. The username should be stored in the server-side session where it can be retrieved as needed.
+> These methods can be used together, and it is recommended most of the times. Make sure that the user always have a mean to recover their account.
```
bigshebang, line 40: Is this meant to be a quote? It's also got a couple small issues. I'd rephrase as the below:

> These can be used together and many times it is recommended to do so. No matter what you must ensure that a user always has a way to recover their account.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
 
-## Step 3) Send a Token Over a Side-Channel
+In order to implement the forgot password service, the developer needs to choose one of the proposed [methods](#methods).
 
-After step 2, lock out the user's account immediately. Then SMS or utilize some other multi-factor token challenge with a randomly-generated code having 8 or more characters. 
+### One Time Password
 
-This introduces an *out-of-band* communication channel and adds defense-in-depth as it is another barrier for a hacker to overcome. If the bad guy has somehow managed to successfully get past steps 1 and 2, he is unlikely to have compromised the side-channel. It is also a good idea to have the random code which your system generates to only have a limited validity period, say no more than 20 minutes or so. That way if the user doesn't get around to checking their email and their email account is later compromised, the random token used to reset the password would no longer be valid if the user never reset their password and the *reset password* token was discovered by an attacker. 
+OTP is the best method in order to implement a secure forgot password service that triggers as a 2FA functionality.
```
bigshebang, line 48: I think it's worth giving a quick overview of the OTP process so it's completely obvious with no room for misinterpretation. I'd add something like this:

> A one time password is generated for the user requesting the reset, then sent to them using an out of band channel. The user can then login with this password. Upon login with a one time password they would be required to change their password before being able to take any other actions. Also the OTP that was given must no longer allow a successful login since the user just used the password once successfully.

Maybe it makes sense to have a separate section about requiring them to reset the password right away instead of having it under the OTP method only.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
-- The password may be logged by third parties such as mail security scanning platforms.
+For a better description of OTP generation, refer to the [MFA CS](Multifactor_Authentication_Cheat_Sheet.md#something-you-have), where the pros and cons of every implementation are provided.
 
-Of course, by all means, once a user's password has been reset, the randomly-generated token should no longer be valid.
+One implementation can be found over for [Authy](https://www.twilio.com/docs/authy/tutorials/two-factor-authentication-python-flask). If you don't want to rely on applications (such as Authy, Google/Microsoft Authenticator, etc.), you can generate QR codes and send them from the server for the user to scan using their device. The following python library, [pyotp](https://github.com/pyauth/pyotp), helps the developer implement any of the methods discussed in this section.
 
-## Step 4) Allow user to change password in the existing session
+> OTPs can be sent through other channels as well, such as emails and SMSs. Various attacks and weaknesses have been identified in SMS that it is preferrable not to use them for OTPs.
```
bigshebang, line 56: Here are some interesting resources related to SMS for MFA that might make sense to add here as additional reading or just citations for proof. I'm sure there are even better resources lurking out there that I'm not aware of, I just grabbed the first few from Google.

* https://www.schneier.com/blog/archives/2016/08/nist_is_no_long.html
* https://auth0.com/blog/why-sms-multi-factor-still-matters/
* https://www.theverge.com/2017/9/18/16328172/sms-two-factor-authentication-hack-password-bitcoin
* https://blog.sucuri.net/2020/01/why-2fa-sms-is-a-bad-idea.html

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
 
-Step 4 requires input of the code sent in step 3 in the existing session where the challenge questions were answered in step 2, and allows the user to reset his password. Display a simple HTML form with one input field for the code, one for the new password, and one to confirm the new password. Verify the correct code is provided and be sure to enforce all password complexity requirements that exist in other areas of the application. 
+### URL Tokens
```
bigshebang, line 58: This is essentially a one time password just placed in the URL. An implementation of this could just be an actual OTP added to a query string in a URL. Does it make sense to group these methods together and mention that generation and use can be a little different? Or it might be better keeping these separated, I'm just raising the similarity.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+  * [Methods](#methods)
+  * [Methods Implementation](#methods-implementation)
+    + [One Time Password](#one-time-password)
+    + [URL Tokens](#url-tokens)
+    + [Security Questions](#security-questions)
+    + [Backup Codes](#backup-codes)
+  * [Operational Tasks](#operational-tasks)
 
-At a minimum, you should have collected some data that will allow you to send the password reset information to some out-of-band side-channel, such as a (possibly different) email address or an SMS text number, etc. to be used in Step 3.
+## Methods
```
bigshebang, line 29: What about knowledge-based auth as another method? It's separate from security questions but similar. Usually financial institutions use this and depending on the knowledge they look for it can be a decent control. A combination of any of these types of things: social security number, full address, account number, most recent transaction and what it included, what year the account was opened, [LexisNexis](https://risk.lexisnexis.com/products/instantid-q-and-a) [KBA](http://lexisnexis.com/risk/downloads/idm/role-of-knowledge-based-authentication-in-identity-proofing.pdf) type questions (the ones car insurance companies ask you), etc. Google's and Amazon's password reset processes also have some interesting questions they ask that are not so easy for people to find through OSINT.

If added we can mention the weaknesses associated with the method. But this method combined with OTP sent to email would be pretty solid for MFA reset.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+
+- Minimum length of 8 digits, 12 for improved security.
+- Randomly generated and unique (CSPRNG use is recommended).
+- Should be stored in a secure fashion, refer to the [Password Storage CS](Password_Storage_Cheat_Sheet.md) for secure practices.
+  - If the service allows users to view the backup codes, the codes should be [securely stored](Cryptographic_Storage_Cheat_Sheet.md) and access should only happen in an authenticated session after asking for a user identifier (password, email token, etc.).
+- One time use.
+- Linked to the user.
+- A user should have multiple recovery codes at any given time to ensure that one of them works (most services provide the user with 10 backup codes).
+
+## Operational Tasks
```
bigshebang, line 97: I think these are all great items! But some are vague and could be misinterpreted or misunderstood. Some also would do well to mention why they are necessary. For example: "Send the user an email that their password has been reset so that if they did not reset it they will be notified about it and be alerted to their account being compromised".

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
+
+- Ask the user if they want to invaldiate all of the sessions, or invalidate the sessions by default without prompting the user.
```
bigshebang, line 99: Misspelled "invalidate" the first time.

bigshebang:
Is the intention that any of these methods alone would be sufficient for a password or MFA reset? Because I think MFA reset should have a higher bar to reset than a password from a logged out state. If an attacker compromises a user's email address for site A they should not be able to just trigger an MFA reset/removal and password reset to own the user's account on site A. The way this document reads now would make this possible.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
 
-There is no industry standard for implementing a **Forgot Password** feature. The result is that you see applications forcing users to jump through myriad hoops involving emails, special URLs, temporary passwords, personal security questions, and so on. In the end you have to reset it to a new value.
+As much as is this functionality looks straight-forward and easy to implement, the details of its implementation makes it a sweet spot for security attacks, such as the renowned [user enumeration attack](https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account.html).
 
-# Steps
+The following short guidelines can be used as a quick reference to protect the forgot password or MFA service:
 
-## Step 1) Gather Identity Data or Security Questions
+- **Unified user message, whether the user exists or not.**
+- **Ensure that the time taken for the user response message is uniform.**
```
bigshebang, line 12: I find it weird that these 2 items are not mentioned in the rest of the document but the third item is. This part reads like it's a TLDR of the whole document but that's not how it actually works currently. IMO this "quick reference" should be essentially a TLDR of the main ideas from the rest of the document. These 2 maybe belong under the Operational Tasks section at the bottom.

Being a person with appsec experience I understand exactly what these mean but a developer or new appsec person looking at their first or second cheat sheet might not. What's something short that can be added to help with that? Maybe a link to or mention of user enumeration for item one and timing attacks for item two.

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: This CS is not about MFA. We should link to this CS https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Multifactor_Authentication_Cheat_Sheet.md#sms-messages-and-phone-calls and improve MFA CS one if needed

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: Knowledge-based auth AKA System Defined Security Questions is already mentioned in this CS: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Choosing_and_Using_Security_Questions_Cheat_Sheet.md#types-of-security-questions
we should link to that CS and do not re-write the same topic in multiple CSs

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: This is covered in MFA CS.

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: I agree on adding more context by mentioning why they are necessary. But @bigshebang which point can be misinterpreted or misunderstood? Do you have any ideas how to make it better?

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
...
 
-- The transport mechanism used to send the password (email, SMS, etc) may not be encrypted.
-- The password may be stored in the users' mailbox or device for an indefinite period.
-- The password may be accessible to IT staff or other users who have access to the mailbox or device.
-- The password may be logged by third parties such as mail security scanning platforms.
+For a better description of OTP generation, refer to the [MFA CS](Multifactor_Authentication_Cheat_Sheet.md#something-you-have), where the pros and cons of every implementation are provided.
 
-Of course, by all means, once a user's password has been reset, the randomly-generated token should no longer be valid.
+One implementation can be found over for [Authy](https://www.twilio.com/docs/authy/tutorials/two-factor-authentication-python-flask). If you don't want to rely on applications (such as Authy, Google/Microsoft Authenticator, etc.), you can generate QR codes and send them from the server for the user to scan using their device. The following python library, [pyotp](https://github.com/pyauth/pyotp), helps the developer implement any of the methods discussed in this section.
```
calabacito, line 54: Since we are mentioning examples of OTP applications, for the sake of the reader: is it a crazy idea to mention QR implementations such as Web Whatsapp? 
Most of us will known what are we talking about but somehow I feel it is worth mentioning since the document is already giving other examples.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,104 @@
-# Introduction
+# Forgot Password and MFA Cheat Sheet
```
calabacito, line 1: Since it is the title, shouldn't be Multi-factor Authentication instead of MFA?

cheatsheets/Forgot_Password_Cheat_Sheet.md
bigshebang: I don't see anything about it in the "Something You Have Section" of the MFA CS and a quick CTRL-F of the whole CS didn't bring anything back related to this. Maybe you could share what content specifically you're thinking of? I also think this document should stand on its own and give the basic high level idea and then can reference other cheatsheets or resources. This section on its own is not completely clear on how to implement OTP or even what OTP is. Given that the [intended audience](https://github.com/OWASP/CheatSheetSeries/blob/master/CONTRIBUTING.md#content-guidelines) is developers who may not have heard of OTP, I think it's worthwhile to explain at a high level what OTP is.

cheatsheets/Forgot_Password_Cheat_Sheet.md
bigshebang: I understand this is not about MFA, but these articles discuss the security of using SMS for sending a sensitive value which is exactly what this cheat sheet is talking about in this section. Sending MFA codes or password reset values via SMS is the same concept applied in 2 different areas. If anything it would probably make the most sense to create a CS on out of band secure channels and include this there.

I don't see the harm in adding an in-line link to the existing sentence which is already saying that SMS is not preferable. I just think it would be nice to include something that explains why. In-line link seems like an elegant way to do that without stepping on the toes of another CS. In a perfect world the link would be to a CS about secure out of band channels.

cheatsheets/Forgot_Password_Cheat_Sheet.md
bigshebang: Ah yes, that seems like the perfect place to have these details. I guess my only issue here then would be the name. "security questions" seems too specific; I think "Knowledge-based auth" captures the entire method best. But I'm probably splitting hairs here so feel free to ignore.

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Nice suggestion :+1:

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Quotes are used as Notes. They step out of the context of the thing being written. The sentence is improved as well, thank you

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Yet OTP is not anything specific to MFA. This is authentication if it comes down to it. I can write it in this document for now, and then make sure the new refreshed Authentication CS reflects that. Would everyone agree on this?

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: I think this is on me for being lazy (for now) and not adding an inline link. Thanks for the discussion. Secure out of band channels ... I think out of band doesn't belong to any one CS. Under what generalization could it fit, if we were to discuss it?

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: QR codes are used more into creating the MFA link, and not for resetting the MFA link. This exists in the [MFA CS](https://cheatsheetseries.owasp.org/cheatsheets/Multifactor_Authentication_Cheat_Sheet.html#software-totp-tokens). I've never seen a MFA reset using QR codes. If there is such an implementation, we can discuss it, definitely.

ThunderSon:
- Removed the MFA parts for now. This needs more thinking to properly implement. I'll do that tomorrow. It made the CS a little bit weak.
- Fleshed out how the service should behave in order to provide enough security to the user.
- Consolidated security best practices for tokens and codes in one section.
- Tweaks here and there based on the reviews.

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Please, open up an issue so that we can have proper discussions about this. :smile:

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Should be resolved with the new structure

cheatsheets/Forgot_Password_Cheat_Sheet.md
calabacito: My bad. Now that I am re-reading this paragraph I think why I miss-understood and got lost suggesting the above. 
 
Trying to make it easier to understand, can I suggest a little re-write of it?  

> If you don't want to rely on applications (such as Authy, Google/Microsoft Authenticator, etc.), there are libraries like [pyotp](https://github.com/pyauth/pyotp), which helps the developer implement on their own any of the methods discussed in this section.

The example of the QR scanning distracted me from the important point I think. 
Maybe it is just me! hah.

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: Yes good idea

cheatsheets/Forgot_Password_Cheat_Sheet.md
mackowski: hmm Authentication CS? Since you send tokens out of band for authentication the user, right? No matter if this is part of MFA, forgot password etc main usage is to authenticate user.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,131 @@
-# Introduction
+# Forgot Password Cheat Sheet
 
-This article provides a simple model to follow when implementing a *forgot password* web application feature.
+## Introduction
 
-# The Problem
+In order to implement a proper user management system, almost all systems integrate a **Forgot Password?** service, which allows the user to request a reset to their password whenever they forget any of them, or if their account ever gets breached.
```
mackowski, line 5: '?' after 'Forgot Password' is unnecessary

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -15,16 +15,20 @@ The following short guidelines can be used as a quick reference to protect the f
 
 ## Content
 
-- [Forgot Password and MFA Cheat Sheet](#forgot-password-and-mfa-cheat-sheet)
```
mackowski: Personally I do not like this change because we have separate MFA CS and here we should only content relevant to Forgot Password and cross-reference MFA CS.

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Sure, sounds more concise without too much details :smile:

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: I'm still considering the focus properly.

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Not really sure what at what you're pointing. I am guessing OTPs? It's a method. I did cross reference it, I had to discuss it though. It's barely a 2 liner description.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+- Unified user message, whether the user exists or not.
+- Unified time for the user response. A quick exit method should not be used if a user doesn't exist and an email is not sent.
+- Rate-limiting (*e.g.* Captchas, blocking IPs for a period of time, etc.) to protect the application against brute-force attacks.
+- Employ proper [Input Validation](Input_Validation_Cheat_Sheet.md) on the sent field.
```
rbsec, line 46: Not entirely sure what you mean - is this "validate the username or email address"?

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
 
-## Step 3) Send a Token Over a Side-Channel
+### Code or Token Activation
 
-After step 2, lock out the user's account immediately. Then SMS or utilize some other multi-factor token challenge with a randomly-generated code having 8 or more characters. 
+After a user uses the service to reset their password, a code or token is provided to the user following one of the [methods](#methods) discussed later on in this document. In order to securely handle that token or code once the user inserts it, the same security practices employed for the [forgot password request](#forgot-password-request) should be used with one additional measure that should be used when sending the code or token in the querystring:
```
rbsec, line 50: The second sentence here is very hard to read - I'm not entirely sure what it's saying.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
-- The password may be stored in the users' mailbox or device for an indefinite period.
-- The password may be accessible to IT staff or other users who have access to the mailbox or device.
-- The password may be logged by third parties such as mail security scanning platforms.
+Once the user is validated through the provided token or code, they should reset their password to a new secure one. In order to secure this last step, the measures should be taken:
 
-Of course, by all means, once a user's password has been reset, the randomly-generated token should no longer be valid.
+- Validate that a secure password policy is in place.
+- The user should confirm the password they set by writing it twice.
+- Update and store the password following [secure practices](Password_Storage_Cheat_Sheet.md).
+- Send the user an email informing them that their password has been reset.
```
rbsec, line 61: Worth noting that this email should not contain the actual password?

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+This can be done through any of the following methods:
 
-The reset must be performed before any other operations can be performed by the user.
+- One Time Password (OTP).
+- URL tokens.
+- Security Questions.
+- Offline backup codes.
 
-## Step 5) Logging
+> These methods can be used together and many times it is recommended to do so. No matter what you must ensure that a user always has a way to recover their account.
```
rbsec, line 76: Are you quoting this from somewhere? If so, reference it. If not, remove the `> ` quote marker

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
 
-# Other Considerations
+It is essential to employ security practices for the reset codes and tokens that will be used in the methods.
 
-- Whenever a successful password reset occurs, all other sessions should be invalidated. Note the current session is already authenticated and does not require a login prompt.
-- Strength of questions used for reset should vary based on the nature of the credential. Administrator credentials should have a higher requirement.
-- The ideal implementation should rotate the questions asked in order to avoid automation.
+**Common secure practices**:
+
+- Randomly generated with a [CSPRNG](https://en.wikipedia.org/wiki/Cryptographically_secure_pseudorandom_number_generator) (*e.g.* [secrets](https://docs.python.org/3/library/secrets.html) library in Python).
```
rbsec, line 84: We have a section on secure randomness in another CS (Cryptographic storage?). Link to that

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+
+**Tokens secure practices**:
+
+- Long enough to avoid brute-force attacks (16 characters should be the minimum used).
+
+**Codes secure practices**:
+
+- Minimum length of 8 digits, 12 for improved security.
+- If the service allows users to view the backup codes, the codes should be [securely stored](Cryptographic_Storage_Cheat_Sheet.md) and access should only happen in an authenticated session after asking for a user identifier (password, email token, etc.).
+- A user should have multiple recovery codes at any given time to ensure that one of them works (most services provide the user with 10 backup codes).
```
rbsec, line 98: There also needs to be a way to invalidate all current codes

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
-Of course, by all means, once a user's password has been reset, the randomly-generated token should no longer be valid.
+- Validate that a secure password policy is in place.
+- The user should confirm the password they set by writing it twice.
+- Update and store the password following [secure practices](Password_Storage_Cheat_Sheet.md).
+- Send the user an email informing them that their password has been reset.
+- Ask the user to re-login. Don't auto-login users on password reset!
+- Ask the user if they want to invalidate all of the sessions, or invalidate the sessions by default without prompting the user.
 
-## Step 4) Allow user to change password in the existing session
+## Methods
```
rbsec, line 65: This is one the same level as "Methods Implementation" and "Tokens and Codes Secure Practices", which doesn't really make sense. This part of the CS doesn't flow very well.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+
+## Methods Implementation
+
+In order to implement the forgot password service, the developer needs to choose one of the proposed [methods](#methods).
+
+### One Time Password
+
+One Time Password (OTP) is the best method in order to implement a secure forgot password service that triggers as a 2FA functionality.
+
+The two most famous methods are Time-OTP ([TOTP](https://tools.ietf.org/html/rfc6238)), or HMAC-OTP ([HOTP](https://tools.ietf.org/html/rfc4226)). The main difference is in the counter, where TOTP focuses on the Unix time, and HOTP has a counter that gets incremented on every user call to generate the OTP.
```
rbsec, line 108: "well-known" would be better than "famous"

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+
+For a better description of OTP generation, refer to the [MFA CS](Multifactor_Authentication_Cheat_Sheet.md#something-you-have), where the pros and cons of every implementation are provided.
+
+One implementation can be found over for [Authy](https://www.twilio.com/docs/authy/tutorials/two-factor-authentication-python-flask). If you don't want to rely on applications (such as Authy, Google/Microsoft Authenticator, etc.), there are libraries like pyotp, which helps the developer implement on their own any of the methods discussed in this section.
+
+> OTPs can be sent through other channels as well, such as emails and SMSs. [Various attacks and weaknesses](https://en.wikipedia.org/wiki/SIM_swap_scam) have been identified in SMS that it is [preferrable not to use them for OTPs](https://auth0.com/blog/why-sms-multi-factor-still-matters/).
```
rbsec, line 114: Again, why is this a quote?

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+1. Access the URL with the attached token.
+2. Set the token in a cookie (recommended), or the client-storage mechanism that your application uses.
+3. Remove the token from the URL and redirect the user to the password reset service. This ensures that protection against [referer leakage](https://portswigger.net/kb/issues/00500400_cross-domain-referer-leakage).
+4. Let the user create a new password and confirm it. Ensure that the password policy is applied.
+
+### Security Questions
+
+> This method should not be used as the sole method to reset a password, and should be used in conjunction with other methods.
+
+Kindly refer to the [Security Questions Cheat Sheet](Choosing_and_Using_Security_Questions_Cheat_Sheet.md) for further guidance.
```
rbsec, line 133: Rephrase to something along the lines of

Security questions should not be used as the sole mechanism for resetting passwords due to PROBLEMS. However, if they are used as a factor then ensure that secure questions are chosen, as discussed in the Security Questions CS.

cheatsheets/Forgot_Password_Cheat_Sheet.md
```diff
@@ -1,60 +1,137 @@
...
+
+### Backup Codes
+
+Backup codes should be provided to the user upon registering where the user should store them offline in a secure place (password managers). Some companies that implement this method are [Google](https://support.google.com/accounts/answer/1187538), [GitHub](https://help.github.com/en/github/authenticating-to-github/recovering-your-account-if-you-lose-your-2fa-credentials), and [Auth0](https://auth0.com/docs/mfa/guides/reset-user-mfa#recovery-codes).
```
rbsec, line 137: This doesn't really tell me anything about implementing them - the implementation details seem to be in the previous section?

jmanico:
I ❤️ all of these changes.

--
Jim Manico
@Manicode

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: It's basically telling the backend to clean the incoming data to avoid input injection attacks. So if you're expecting an email, you should be locking down the input field to that and employing input validation. I'll try to reword it.

jmanico:
Injection is not solved by validation. You need query parameterization or escaping. 

Even valid email addresses can cause SQLi like Jim’or’1’!=‘@manicode.com

Thanks.
--
Jim Manico
@Manicode

rbsec:
That reminds me, it wouldn't hurt to put a warning about `Host` header injection somewhere in the email section. I've found that a couple of times in forgotten password functions, and it can be pretty bad if that header is used in the link sent to the user.

rbsec:
I don't mind whether it's included in this one, or added to the injection one and linked to from here - but I've only really seen it (exploitable) in forgotten password functions, so I think it needs at least a reference here.

jmanico:
I get your point and stand corrected.

•hat tip•

--
Jim Manico

cheatsheets/Forgot_Password_Cheat_Sheet.md
ThunderSon: Using quotes to look like a note. I will update this to follow notes that are provided by MkDocs. I'll do that after verifying the content and being happy with it.

ThunderSon:
I deleted my fork to work on a clean slate a couple of weeks ago. A new PR has been created.

PR #382, opened 2020-04-02
Point to the correct Angular docs

domfarolino:
This cheat sheet pointed to the wrong angular docs. This was discussed/found in https://chromium-review.googlesource.com/c/chromium/src/+/2134569/2/services/network/cross_origin_read_blocking_explainer.md


ThunderSon:
A small note to this PR: It wasn't pointing to the wrong link, more so to the whole article, instead of the section, which is confusing in the sentence provided by the sheet.

Issue #200, opened 2019-09-01
Migration of the CI/CD infrastructure to Python > 3.5


righettod:
* Migrate all the CI/CD pipeline descriptors in order to use a latest version of python (use a common version to TravisCI and CircleCI).
* Review all the Python scripts used to manage the project (folder **scripts**) in order to use the lastest feature of python and enhance them.
* Find an alternative to GitBook V1 system to generate the website because [V2 do not propose toolchain anymore](https://docs.gitbook.com/v2-changes/important-differences#cli-toolchain).

sj66phone:
#201 Add link to GitHub on cheatsheets to encourage contribution

nicoduj:
Hi, concerning gitbook, what about switching to Mkdocs ? I used it on this repo to have an internal website with an extraction of those cheat sheets focused on our work, and it does not involve too much work I think. I can try to do it and make a pull request if you are interested.

And by the way ,thank you for this repo, it is very valuable !

ThunderSon:
@nicoduj Hi! Forgive me, this somewhat slipped me? If you still have it, would you provide with a screenshot of what you have? :smile:

nicoduj:
Hi @ThunderSon , I have forked the repo and made the changes to it. 

Here are some screenshots with [Material](https://squidfunk.github.io/mkdocs-material/) theme: 

**Desktop**
<img width="1317" alt="Capture d’écran 2020-05-08 à 15 48 19" src="https://user-images.githubusercontent.com/19813688/81426110-07f78400-9159-11ea-8940-fdd3cf9b83de.png">
<img width="1264" alt="Capture d’écran 2020-05-08 à 15 48 46" src="https://user-images.githubusercontent.com/19813688/81426116-0a59de00-9159-11ea-8a6e-bef7e633c6a9.png">

**Mobile**
![IMG_3927](https://user-images.githubusercontent.com/19813688/81426195-2493bc00-9159-11ea-87f1-d45a27633a1f.PNG)
![IMG_3928](https://user-images.githubusercontent.com/19813688/81426204-26f61600-9159-11ea-8893-9a7ef52a7c22.PNG)
![IMG_3929](https://user-images.githubusercontent.com/19813688/81426209-29587000-9159-11ea-9dad-555b945dc32f.PNG)


And here with [readthedocs](https://mkdocs.readthedocs.io/en/stable/user-guide/styling-your-docs/#readthedocs) theme:

**Desktop**
[...]

Issue #386, opened 2020-05-04
OWASP Google Analytics


ThunderSon:
As per OWASP's request, we are going to add the google analytics to the cheatsheets website.
The following script needs to be appended on every page:
```javascript
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-4531126-15"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'UA-4531126-15');
</script>
```
@rejahrehim would you want me to assign this to you?

rejahrehim:
Yes, Assign me

rbsec:
@ThunderSon I presume you'll have access to the analytics data?

ThunderSon:
@hblankenship I usually reach out to Harold and grab information when need be. Harold is there any dashboard we could access once it is set?

ThunderSon:
Closed in #387

PR #387, opened 2020-05-08
Migration to mkdocs

nicoduj:
PR for migration to mkdocs instead of deprecated gitbook v1.

Configured for use material theme, comments to allow readthedocs theme in yml configuration file.

Please note that some cheat sheets needs modification for best results in TOC generation, since they have more than one ```<h1>``` (equivalent to markdown ```#``` ) in them, like this [one](https://nicoduj.github.io/CheatSheetSeries/cheatsheets/Abuse_Case_Cheat_Sheet/)

(custom 404 file was added for "demonstration" purpose (and a bit of fun), can be ignored or removed)

This PR is related to issue #200 where you can find screenshots and links to live preview


ThunderSon:
Once this is accepted, we'll have to go over the sheets to make sure that lists are using `-` and not `*` (which should be easy with regex), and a small review for h1 headers will be done. We'll need to see precisely what's breaking or changing. Once I get my full feedback I'll pick between the themes and then dig into where it's breaking. (The needle so far is tending a bit more towards material)

.github/workflows/build-and-deploy-website.yml
```diff
@@ -58,9 +58,17 @@ jobs:
       run: |
         rm -rf assets 1>/dev/null 2>&1
         rm -rf cheatsheets 1>/dev/null 2>&1
-        rm -rf gitbook 1>/dev/null 2>&1
-        rm index.html 1>/dev/null 2>&1
-        rm search_index.json 1>/dev/null 2>&1
+        rm -rf search 1>/dev/null 2>&1
```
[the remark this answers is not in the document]
nicoduj: Agree, don't like this very much also. I think what we only want to keep there is CNAME and robots.txt, right ?

nicoduj:
> I didn't properly get the $date stuff in the generation file, if you could give me a hint on it.

the date styuff is in cas ewe use a template in mkdocs that use it : [Markdown style] metadata(https://www.mkdocs.org/user-guide/writing-your-docs/#multimarkdown-style-meta-data) which I think is not used by both themes we are talking about.

It is correctly added into the Markdown before mkdoc generation : 

Capture d’écran 2020-05-09 à 16 47 48

but not in the html result with material. We can remove this if you prefer.

nicoduj:
> Once this is accepted, we'll have to go over the sheets to make sure that lists are using `-` and not `*` (which should be easy with regex), and a small review for h1 headers will be done. We'll need to see precisely what's breaking or changing. 

I have pushed also changes to `*` to `-`

Concerning h1 headers, I prefer you tell me what to do, since in material the title is extracted automatically from the name of the fil in my proposal, but we could switch to a h1 inserted or asked instead (if we do both there will be I think twice the title in material, not in readthedocs)

.github/workflows/build-and-deploy-website.yml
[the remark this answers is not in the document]
nicoduj: I have used extglob and the followinfg syntax for the moment t simplify this, oncsidering only CNAME (which I removed from my fork) and robots.txt are the only files to keep :
```bash
shopt -s extglob
rm -rdfv !("CNAME"|"robots.txt")
```

Going to a full github action to publish, and thus limiting the script file would be better I agree but it involves more changes (and tests) I think, especially since we need due to the repo structure to do some "operation" before running mkdocs. Step by Step like you said :smile:

ThunderSon:
@nicoduj I did the updates required over all the sheets. Once the other PR is accepted, we can merge it with this one and push forward.

ThunderSon:
I'd prefer removing the additional unused content, yet thank you for doing that.

nicoduj:
> I'd prefer removing the additional unused content, yet thank you for doing that.

So we remove the date in the script right ?

mkdocs.yml
```diff
@@ -0,0 +1,31 @@
+site_name: OWASP Cheat Sheet Series
+site_description: Website with the collection of all the cheat sheets of the project.
+docs_dir: cheatsheets/
+plugins:
+    - search:
+        prebuild_index: true
+#theme:
```
rejahrehim, line 7: google_analytics: ['UA-4531126-15', 'auto']

rejahrehim, line 7: Can we add google analytics here ? 
`google_analytics: ['UA-4531126-15', 'auto']`

This will solve #386

nicoduj: can be done at the top level, not in the plugin part, but yes in my opinion.

ThunderSon:
@nicoduj I merged the big PR. Are you able to merge and move this forward?
I opened an issue with the markdown link check repo to see if we can filter based on response status, so no worries about that for now.
Let me know if I can help you in any way.

mkdocs.yml
ThunderSon: Can you please inject it as well? That'd be a great help.

ThunderSon:
Once you merge with your master, I'll add the front matter to every page. We need to add the redirection to handle the move from `.html` links to the current one. I'll do this.

nicoduj:
Great  news, will try to find some time in the next days to work on this and will let you know . If I sum up what to do 
- merge the master into my fork
- remove date in markdown comment
- update markdown title comment to remove the cheat sheet suffix
- add ga settings from @rejahrehim

What theme will be used by the way ?

ThunderSon:
Updated the comment. Yep, what you said as well, made it a bit nicer to track :)

nicoduj:
@ThunderSon I have made the changes but there are some problems I'd like to discuss : the title which in markdown still have the "Cheat Sheet" (# level) and some file names that are not well formatted (no _Cheat_sheet, duplicate for C_Based-ToolChain, and Ruby_on_Rails_CheatSheet. 
I think it would be better to normalise all this. 

You can send me the link for Slack, will be easier like you said

ThunderSon:
For the first point, that's alright. That's the title. The filename shouldn't have it because you're in the cheatsheets section already :)
Hop in, follow this section in the readme: https://github.com/OWASP/CheatSheetSeries#chat-with-us

ThunderSon:
What's left here is the redirection to happen for some of the renamed CSs (mainly Java ones, RoR). Required change.
The full naming update will happen in another PR with redirection to handle 404s.

Considering adding front-matter to the current changed CSs. Just thinking if this should happen on build time, or added in the CS.

ThunderSon:
@nicoduj kindly merge from master the latest changes

ThunderSon:
I believe we're at the stage of reviewing overall how everything is working. The theme move is big enough. We can look at plugins, extensions, and styling in the next iteration :smile:

ThunderSon:
I went over your last couple of commits, fixed some merge issues. I think this should be good now.

PR #391, opened 2020-05-15
Clarify maximum password length guidelines in authentication CS. Fixe…

rbsec:
Align the recommendations here with the password storage CS.

### DIFF
--- a/cheatsheets/Authentication_Cheat_Sheet.md
+++ b/cheatsheets/Authentication_Cheat_Sheet.md
@@ -27,7 +27,7 @@ A key concern when using passwords for authentication is password strength. A "s
     - **Minimum** length of the passwords should be **enforced** by the application. Passwords **shorter than 8 characters** are considered to be weak ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)).
     - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. A common maximum length is 64 characters due to limitations in certain hashing algorithms, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.md#maximum-password-lengths). It is important to set a maximum password length to prevent [long password Denial of Service attacks](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).
 
-- Do not silently truncate passwords. If the user enters a password that is longer than that maximum permitted length then return an error message explaining to them what the problem is.
+- Do not silently truncate passwords. The [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.md#maximum-password-lengths) provides further guidance on how to handle passwords that are longer than the maximum length.
 
 - Allow usage of **all** characters including unicode and whitespace. There should be no password composition rules limiting the type of characters permitted.
 
--- a/cheatsheets/Authentication_Cheat_Sheet.md
+++ b/cheatsheets/Authentication_Cheat_Sheet.md
@@ -25,12 +25,9 @@ A key concern when using passwords for authentication is password strength. A "s
 
 - Password Length
     - **Minimum** length of the passwords should be **enforced** by the application. Passwords **shorter than 8 characters** are considered to be weak ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)).
-    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. Typical maximum length is 128 characters. It is important to set a maximum password length to prevent [long password Denial of Service attacks](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).
-    - Some password hashes such as Bcrypt truncate the input, so a shorter maximum length may be required, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.md#maximum-password-lengths).
+    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. A common maximum length is 64 characters due to limitations in certain hashing algorithms, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.html#maximum-password-lengths). It is important to set a maximum password length to prevent [long password Denial of Service attacks](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).
 
-      When selecting a maximum password length, consider whether the hashing algorithm to be used has any limitations because some [have a maximum password length](https://security.stackexchange.com/questions/39849/does-bcrypt-have-a-maximum-password-length/39851#39851).
-
-- Do not truncate passwords. Make sure that every character the user types in is actually included in the password.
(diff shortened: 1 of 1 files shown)


cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -25,12 +25,9 @@ A key concern when using passwords for authentication is password strength. A "s
...
-    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. Typical maximum length is 128 characters. It is important to set a maximum password length to prevent [long password Denial of Service attacks](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).
-    - Some password hashes such as Bcrypt truncate the input, so a shorter maximum length may be required, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.md#maximum-password-lengths).
+    - **Maximum** password length should not be set **too low**, as it will prevent users from creating passphrases. A common maximum length is 64 characters due to limitations in certain hashing algorithms, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.html#maximum-password-lengths). It is important to set a maximum password length to prevent [long password Denial of Service attacks](https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/).
```
ThunderSon, line 28: The link for the password storage is `html` and not `md`

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -25,12 +25,9 @@ A key concern when using passwords for authentication is password strength. A "s
...
 
-      When selecting a maximum password length, consider whether the hashing algorithm to be used has any limitations because some [have a maximum password length](https://security.stackexchange.com/questions/39849/does-bcrypt-have-a-maximum-password-length/39851#39851).
-
-- Do not truncate passwords. Make sure that every character the user types in is actually included in the password.
+- Do not silently truncate passwords. If the user enters a password that is longer than that maximum permitted length then return an error message explaining to them what the problem is.
```
ThunderSon, line 30: Don't you think pre-hashing could be referenced in this point from pass storage CS?
Maybe create 2 bullets under `Do not silently truncate password` one for each option.

rbsec: The sections linked the the bullet point above, so it seems a little unnecessary to link again here. Even if you're pre-hashing you still need to have some kind maximum length (although probably a very high one), so I think the point still stands.

rbsec:
@ThunderSon the link check is failing, but it looks more like a bug with the code that any broken links:

```
> cheatsheetseries@1.0.1 link-check /home/runner/work/CheatSheetSeries/CheatSheetSeries
> find  cheatsheets -name \*.md -exec markdown-link-check -c markdown-link-check-config.json 1> log 2> err {} \; && if [ -e err ] && grep -q  "ERROR:" err ; then exit 113  ; else echo -e "All good"; fi

npm ERR! code ELIFECYCLE
npm ERR! errno 113
npm ERR! cheatsheetseries@1.0.1 link-check: `find  cheatsheets -name \*.md -exec markdown-link-check -c markdown-link-check-config.json 1> log 2> err {} \; && if [ -e err ] && grep -q  "ERROR:" err ; then exit 113  ; else echo -e "All good"; fi`
npm ERR! Exit status 113
npm ERR! 
npm ERR! Failed at the cheatsheetseries@1.0.1 link-check script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2020-05-16T17_23_35_208Z-debug.log
##[error]Process completed with exit code 113.
```

ThunderSon:
You are reading the wrong stage. Check the one after:
```
 FILE:cheatsheets/Authentication_Cheat_Sheet.md
[✖] Password_Storage_Cheat_Sheet.html#maximum-password-lengths → Status: 400


FILE:cheatsheets/Pinning_Cheat_Sheet.md
[✖] http://www.rsa.com/rsalabs/node.asp?id=2125 → Status: 403
[✖] http://www.rsa.com/rsalabs/node.asp?id=2128 → Status: 403
```

cheatsheets/Authentication_Cheat_Sheet.md
ThunderSon: Then you should be removing everything after `Do not silently truncate passwords`, as the rest is part of the maximum length. I am just following the logic of what's written.

rbsec:
@ThunderSon I see - so the link check fails, and then the "show broken links" succeeds with the list of broken links. Slightly confusing, but fair enough.

I fixed the markdown link (the other two are in a different CS). I re-ran the checks and now the "setup action" and "show broken links" checks are both failing...

cheatsheets/Authentication_Cheat_Sheet.md
rbsec: I think that telling people to present an error to the user here is valid - there's no conflict between that and the previous point. @mackowski do you have a strong feeling either way?

ThunderSon: The idea is the following:
In the Maximum length recommendation, it is instructed to either block the user from setting a length longer than the max allowed, or use some kind of a pre-hash.
In the Do Not Truncate, which happens if the user goes above the max length, they have the same 2 options ...
A cleaner thought could be to say:
> Do not truncate passwords. Refer to the maximum length measures to be taken.
Something similar to that.

I am not saying that presenting an error is not valid!

mackowski: I think that simply:
`Do not silently truncate passwords. See details in [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.html#maximum-password-lengths) how to handle scenario when user enters a password that is longer than that maximum permitted length`
I think that there is nothing bad in linking here again to that CS. In that place it is well described so anything we put here will be only a simplification.

rbsec: Implement a slight variation on the wording that @mackowski suggested.

cheatsheets/Authentication_Cheat_Sheet.md
```diff
@@ -278,7 +275,7 @@ Web applications should at least not make password managers job more difficult t
 
 - Use standard HTML forms for username and password input with appropriate `type` attributes.
   - Avoid plugin-based login pages (such as Flash or Silverlight).
-- Allow long passwords (at least 64 characters).
+- Implement a reasonable maximum password length, such as 64 characters, as discussed in the [Password Storage Cheat Sheet](Password_Storage_Cheat_Sheet.html#maximum-password-lengths).
```
ThunderSon, line 278: Forgot this html instance to md

rbsec: Fixed

PR #395, opened 2020-05-27
Chore/lint 4 spaces indent

ThunderSon:
Mkdocs use python markdown, which in turn implements 4 tabs for indent levels.

Updated the linter and the docs accordingly.


ThunderSon:
Closing this, will re-open a cleaner PR without the branched out commits.

PR #407, opened 2020-06-05
Removed typo space causing rendering problems

ncallister:
This pull request simply removes a leading space on a code block terminator which I believe is causing the problem in the rendering of the dotnet security sheet just below [here](https://cheatsheetseries.owasp.org/cheatsheets/DotNet_Security_Cheat_Sheet.html#os-injection).

Note that the space does not appear to cause problems when rendering with other markdown engines such as github.

As this is a simple typo I didn't see the need to create a ticket, if you wish me to do so please let me know.


ThunderSon:
@ncallister turns out that issue was on more places. Fixed in a new PR. Thanks for spotting this :)

PR #411, opened 2020-06-06
Fix up formating in CSP CS

rbsec:
This CS had some strange formatting where some code examples were given as backticked code inside blockquotes, rather than just using code blocks.


rbsec:
@ThunderSon the linter seems to be flagging code fenced code blocks as an error when there's no language specified, even though this [isn't required](https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#syntax-highlighting) (and often isn't desirable). Can we remove this check?

ThunderSon:
Sure, I'll add a rule to the linter in a bit

Issue #421, opened 2020-06-15
New CS: GraphQL Security Cheat Sheet


ThunderSon:
After various discussion on Slack, it is important to create a GraphQL Cheat Sheet in order to guide developers to writing secure GraphQL endpoints.

@bigshebang to provide us with a draft on work already done from their side.

@PauloASilva FYI. This should be moving properly with the API security project's line as well.

Let's see this effort through. After Luke's draft, we can modify the template and see what's missing and could be improved :smile:

mackowski:
Awesome! ;-)

ErezYalon:
@bigshebang, is there an ETA for the draft? We (together with @PauloASilva) are starting to work on the subject as well.

bigshebang:
@ErezYalon the ETA is now :D https://github.com/OWASP/CheatSheetSeries/pull/434

ThunderSon:
What @bigshebang presented is an effort already made on creating a GraphQL knowledge base. Next up is growing that to cover the defenses spectrum and tidying it up to have a sheet feeling.

ErezYalon:
We are a small team working on several API Security projects.
We will start our GraphQL research soon and add our results to the sheet. Most will probably come from @PauloASilva.

PauloASilva:
Hi guys,
Sorry for the late reply.

My initial comments and thoughts on how to improve the GraphQL Cheat Sheet proposal were [left in the PR](https://github.com/OWASP/CheatSheetSeries/pull/434#pullrequestreview-442852077).

On an initial stage I think we can work on our general suggestion: make the CS more actionable, but only after hearing your thoughts about our comments.

Cheers,
Paulo A. Silva

ThunderSon:
Thanks @PauloASilva ! We didn't provide our full feedback as well. Thanks for tackling this!
I'll add some more comments and try to get some improvements in hopefully in the weekend.

mackowski:
I will try to review it over the weekend

jmanico:
What is the stats? Can I help here?

nikitastupin:
Hi there! I can describe some common vulnerabilities  related to GraphQL APIs that I've found during my bug bounty journey. Also methodology of researching GraphQL APIs from bug hunter perspective.

However as far as I understand this cheatsheet intended for developers. From your perspective, is that kind of information valuable to the cheatsheet?

mackowski:
[...]

PR #434, opened 2020-06-24
New CS: GraphQL

bigshebang:
This PR covers issue #421

This is a new PR with a different branch in my fork


cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+- Queries (data fetching)
+- Mutation (write) queries
+- [Introspection](https://graphql.org/learn/introspection/) system/query
+
+Each of these need to have permission checks for most GraphQL setups.
+
+#### Query Access
+
+As part of a GraphQL API there will be various data fields that can be returned. One thing to consider is if you want different levels of access around these fields. For example, you may only want certain consumers to be able to fetch certain data fields rather than allowing all consumers to be able to retrieve all available fields. This can be done by adding a check in the code to ensure that the requestor should be able to read a field they are trying to fetch. This may be best implemented with simple [role-based access control](https://auth0.com/docs/authorization/concepts/rbac) ([RBAC](https://en.wikipedia.org/wiki/Role-based_access_control)): by creating roles which have access to certain fields and then attaching the correct roles to individual consumers/users. ABAC or other access control methods could also work.
```
mackowski, line 85: I think that we should link to our RBAC cheatsheet https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Access_Control_Cheat_Sheet.md#role-based-access-control-rbac and if we are missing something in that CS let improve it :)

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+> The additional mitigation technique below probably belongs in the IDOR prevention cheat sheet but isn't there so I'll leave it here for now.
+
+One alternative protection would be creating an abstraction of the ID that is usually mapped to the requester's session. For example, instead of the user requesting a picture via its direct ID, they would request picture 2 in their account and the backend would use that abstract ID with the user's ID to derive the actual direct ID of the picture. This prevents the user from being able to provide their own direct identifier for an object and ensures that the user can only access objects belonging to them. Another abstraction mechanism could involve using different IDs for the frontend vs the backend with some unpredictable translation that happens on the backend where the user does not have any influence.
+
+## Other Resources
```
mackowski, line 107: Awesome resources - can we add some kind of categories to them?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Mutation Access
+
+GraphQL supports mutation, or manipulation of data, in addition to its most common use case of data fetching. If your service implements/allows mutation then there may need to be access controls put in place to restrict which consumers, if any, can modify data through the API. Setups requiring mutation access control include APIs where only read access is intended or where only certain parties should be able to modify certain fields. This should be implemented similarly to Query access: use [RBAC](https://auth0.com/docs/authorization/concepts/rbac) and have the code only allow certain roles to perform mutation on approved data fields.
```
PauloASilva, line 89: I do agree with the general idea, but we need to make this more actionable.

* Use [Interfaces] and [Unions] to create hierarchical data types.
  Then, based on requester's profile/permissions, return the most appropriate data type.
* Use Mutation [Resolvers] to update data, and perform access control validation, may be passing some middleware
* Always validate whether requester is authorized to execute the requested mutation ([RBAC]).
  Access Control validation may be performed passing some middleware to the mutation function.

[Interfaces]: https://graphql.org/learn/schema/#interfaces
[Unions]: https://graphql.org/learn/schema/#union-types
[Resolvers]: https://graphql.org/learn/execution/#root-fields-resolvers
[RBAC]: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Access_Control_Cheat_Sheet.md#role-based-access-control-rbac

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Introspection
```
PauloASilva, line 91: [GraphiQL] is quite often found on production systems.
We should also add some recommendation to do not deploy such tools in production environments, nor make them publicly accessible on other environments (e.g. QA, staging, dev).

May be this should has it's own category e.g. "Security Misconfigurations", but it came up to my mind while reviewing the Introspection section, because it's GraphiQL first query.

[GraphiQL]: https://github.com/graphql/graphiql

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+Many implementations of GraphQL have Introspection enabled by default and leave it accessible to any incoming requests without requiring authentication. This is usually problematic because introspection allows the requester to learn all about supported schema and queries (see a [real-world example](https://hackerone.com/reports/291531) abusing this). Introspection might be how the API owner wants to educate consumers about how to use the API. However, the preferred way to educate consumers about a service is through a separate documentation channel such as a wiki, Git Readme, or readthedocs.
+
+The safest and usually easiest approach is to just disable introspection system-wide. See [this page](https://lab.wallarm.com/why-and-how-to-disable-introspection-query-for-graphql-apis/) or consult your GraphQL implementation's documentation to learn how to disable introspection altogether. If your implementation does not natively support disabling introspection or if you would like to allow some consumers/roles to have this access you can build a filter in your service to only allow approved consumers to access the introspection system.
+
+### IDOR Protection
```
PauloASilva, line 97: In OWASP API Security Top 10 2019 we decided to split IDOR back in two different categories:
* [API1:2019 Broken Object Level Authorization]
* [API5:2019 Broken Function Level Authorization]

Insecure Direct Object Reference isn't a very accurate name: there's nothing _insecure_ about using object's unique identifier as long as proper access control validations are performed. Creating some identifiers indirection won't solve the real issue unless the translation mechanism performs the required access control validations.

How are these recommendations different from "Query Access" ones?

I think we should remove this section and add the IDOR CS as a reference to the Query Access section.

[API1:2019 Broken Object Level Authorization]: https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md
[API5:2019 Broken Function Level Authorization]: https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa5-broken-function-level-authorization.md

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Resource Management
+
+> I assume we don't want the "internal vs external" API talk in here?
+
+For a GraphQL API that receives input more or less directly from an external party, it can be difficult to limit the possibilities a malicious user can use to exhaust the API's resources to cause slow downs or indefinite outages. It is easier to protect an API that is only used by other internal services. Because of this, it may be better to only expose your API internally. See the table below for which of these controls should be added to each type of GraphQL API.
+
+#### Depth Limiting
+
+This is limiting the depth of incoming queries so that the user cannot supply unnecessarily deep queries that could consume a lot of resources. See [this](https://www.apollographql.com/blog/securing-your-graphql-api-from-malicious-queries-16130a324a6b) and [this blog post](https://www.howtographql.com/advanced/4-security/) as well as [this Github repo](https://github.com/stems/graphql-depth-limit) (old, potentially unsupported) for more details on implementing this control. I'm not sure if there is a good/trusted open source project that does this.
```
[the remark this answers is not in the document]
bigshebang: Yeah from my quick searching I couldn't really find anything legit or trusted. We can def delete this. It seems like the code to do this is pretty simple though; would be cool to see an official OWASP project tackle this!

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Process Validation
+
+When using user input, even if sanitized and/or validated it should not be used for certain purposes that would give a user control over data flow. For example, do not make an HTTP/resource request to a host that the user supplies (unless there is a business need).
+
+### Resource Management
+
+> I assume we don't want the "internal vs external" API talk in here?
+
+For a GraphQL API that receives input more or less directly from an external party, it can be difficult to limit the possibilities a malicious user can use to exhaust the API's resources to cause slow downs or indefinite outages. It is easier to protect an API that is only used by other internal services. Because of this, it may be better to only expose your API internally. See the table below for which of these controls should be added to each type of GraphQL API.
```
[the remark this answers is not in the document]
bigshebang: This doc was originally made for my company which has APIs that are internal-only and externally-facing. So I was trying to basically say if the API is not exposed externally we don't have to be as careful with preventing DoS given the low likelihood of exploitation and significant cost to add protections against DoS. I don't know if we want to add this line of thinking to a CS; it's basically recommending to accept the risk because it's low enough. Seems out of place in a CS but is valuable info to engineering teams.

There are a ton of things that can be done to prevent DoS, for sure. We could link to the [DoS CS](https://cheatsheetseries.owasp.org/cheatsheets/Denial_of_Service_Cheat_Sheet.html), but it seems in need of work.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I hadn't heard of GraphiQL. Great call out! I think we should add a general note about not providing utilities like this in production and specifically name GraphiQL.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Yeah we can definitely categorize these so they're more easily consumable. I usually don't see sections like this in other sheets.

I'm thinking it might be cleaner to not include links in this section that were mentioned/linked previously in the doc, but I could go either way. Any thoughts on that?

PauloASilva: I would try to fit them along the CS where they make sense: I am not sure someone will go through a long list of links to read one by one. The remaining articles may be left as "Other Resources" as you have now.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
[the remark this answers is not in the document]
PauloASilva: Ok, now I see your point, nevertheless I still think it is risky to say something like that.

* "Internal-only" is something hard to define and someone may lower defenses based on a wrong premise.
* If internal-only APIs interact with external services, then those services are potential threat agents.
* Sometimes threat agents belong to the organization.
* Compromised systems may allow access to internal-only APIs.

I don't think that we should advice accepting even low enough risks.
This is something engineering teams should discuss and define with the business: what's the risk it is willing to accept.

Let's see what other think about this subject.

Regarding the work needed in the DoS CS, may be we can open a new issue to put it in the agenda.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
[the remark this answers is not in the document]
mackowski: If you feel that it is worth to have a short code snippet here we can think about it. 
We try to not have longer code samples because they are hard to maintain but short snippet or pseudocode is a good idea sometimes.

mackowski:
@bigshebang do you need help with resolving these comments? 
@PauloASilva I saw that you have a lot of good ideas here, if you want please suggest changes (you can use suggest change feature from GitHub for it)

mackowski:
@bigshebang are you still around? this is amazing work that just need to be finished!

bigshebang:
@mackowski I am still around! I'm hoping to come back to this in the next week or two

mackowski:
> @mackowski I am still around! I'm hoping to come back to this in the next week or two

Awesome! Thanks for reply!

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,128 @@
...
+    - [SQL](https://portswigger.net/web-security/sql-injection) [injection](https://owasp.org/www-community/attacks/SQL_Injection)
+- [DoS](https://owasp.org/www-community/attacks/Denial_of_Service) ([Denial of Service](https://www.cloudflare.com/learning/ddos/glossary/denial-of-service/))
+- [IDOR](https://portswigger.net/web-security/access-control/idor)
+- Abuse of improper/excessive access
+
+## Best Practices and Recommendations
+
+### Input Validation
+
+Adding strict input validation can help prevent against SSRF, SQL injection, and DoS. The main design for GraphQL is that an identifier is given and the backend has a number of fetchers making HTTP, DB, or other calls. This means that user input will be included in HTTP requests, DB queries, or other requests/calls and there is opportunity for injection that could lead to SSRF or SQL injection. Some pages that may be helpful here are OWASP Cheat Sheets for generic [Injection Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Injection_Prevention_Cheat_Sheet.html) or [Java Injection Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Injection_Prevention_Cheat_Sheet_in_Java.html).
```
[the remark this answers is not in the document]
bigshebang: I very much agree with this feedback! But as I work on this rewrite, I have a problem calling out "ORM injection" as an issue. I have less experience with ORMs so maybe I'm off base here, but it seems to me that when a developer uses an ORM properly it will parameterize their queries and prevent injection entirely. Where devs would get into trouble is using functionality of an ORM that allows for non-parameterized queries like raw query execution/manipulation or similar.

So I think calling it "ORM injection" isn't really accurate; it's actually just plain old SQL injection but using an ORM, which is just a different driver/connection than is traditionally used. Instead I think the advice to devs should be to use parameterized queries with their ORM and not to use their ORM to mess with raw queries.

Any thoughts on that? We don't necessarily need to agree on terminology of whether ORM injection is its own attack, but we should agree on the recommendation to devs using ORMs.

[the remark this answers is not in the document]
PauloASilva: I see your point.

The ORM Injection is a little different from a plain SQL Injection, since the attacker will target the ORM itself or the SQL generated by the ORM, instead of the DBMS.

A developer may choose an ORM that offers parameterized queries as well as a fluent querying API to programmatically build queries ([example](https://www.jooq.org/#examples)). Although the parameterized queries API is safe, attackers may trick the ORM to generate unsafe SQL.

> Any thoughts on that? We don't necessarily need to agree on terminology of whether ORM injection is its own attack, but we should agree on the recommendation to devs using ORMs.

I am OK recommending ORMs/ODMs as long as we add something like: "ORMs/ODMs may introduce other issues such as [ORM Injection](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.7-Testing_for_ORM_Injection): always audit your dependencies."

bigshebang:
New changes are up! Please give your feedback :)

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+- To properly handle unicode input, use a [single internal character encoding](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#validating-free-form-unicode-text)
+- Gracefully [reject invalid input](https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html), being careful not to reveal excessive information about how the API and its validation works.
+
+#### Injection Prevention
+
+When handling input meant to be passed to another interpreter (e.g. SQL/NoSQL/ORM, OS, LDAP, XML):
+
+- Always prefer safe tools with support for parameterized statements
+    - Ensure that you follow the documentation so you are properly using the tool
+    - For SQL, using ORMs/ORDs is a good option but they must be used properly to avoid [ORM injection](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.7-Testing_for_ORM_Injection).
```
bigshebang, line 47: Is there anything better within OWASP that we can link to? The audience for WTSG isn't developers like this CS is, so another resource would work better here if it exists. I couldn't find anything.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+      album {           # Depth: 3
+        ...             # Depth: ...
+        album {id: N}   # Depth: N
+      }
+    }
+  }
+}
+```
+
+> We should have an example of someone taking advantage of no amount limiting and requesting N of an object. I'm not sure if a query can ask for N of an object directly or if the query has to use aliases and make N separate requests within the query for the same object.
```
bigshebang, line 98: Would love some help from anyone with more GraphQL experience on this.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I removed any links that were already included earlier in the doc and split the remaining links into categories as well prettified the links.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+- [Whitelist allowed characters](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#whitelisting-vs-blacklisting) - don't use a blacklist
+    - The stricter the whitelist the better. A lot of times a good starting point is only allowing alphanumeric, non-unicode characters because it will disallow many attacks.
+- To properly handle unicode input, use a [single internal character encoding](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#validating-free-form-unicode-text)
+- Gracefully [reject invalid input](https://cheatsheetseries.owasp.org/cheatsheets/Error_Handling_Cheat_Sheet.html), being careful not to reveal excessive information about how the API and its validation works.
+
+#### Injection Prevention
+
+When handling input meant to be passed to another interpreter (e.g. SQL/NoSQL/ORM, OS, LDAP, XML):
+
+- Always prefer safe tools with support for parameterized statements
```
PauloASilva, line 45: We won't be able to define what "safe tools" are, but "safe APIs" is used by other OWASP documents (e.g. [A1:2017 - Injection](https://owasp.org/www-project-top-ten/OWASP_Top_Ten_2017/Top_10-2017_A1-Injection))

```suggestion
- Always choose libraries/modules/packages offering safe APIs, such as parameterized statements.
```

cheatsheets_draft/GraphQL_Cheat_Sheet.md
PauloASilva, line 47: ODMs (Object Document Mapping) are also a good option to implement the data access layer when using NoSQL databases. I think we can make this recommendation generic.

```suggestion
    - Using ORMs and ODMs are a good option but they must be used properly to avoid flaws such as [ORM injection](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.7-Testing_for_ORM_Injection).
```

PauloASilva: I didn't find anything better then the WTSG.

We could link the [CAPEC-109: Object Relational Mapping Injection](https://capec.mitre.org/data/definitions/109.html) but I like the WTSG better since it includes examples. I am OK linking the WTSG.

bigshebang: Okay we'll stick with the WSTG doc then.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+
+#### Batching Attack
+
+[GraphQL batching attack](https://lab.wallarm.com/graphql-batching-attack/)?
+
+> This is an interesting attack that we should add. It seems like Amount Limiting will prevent it but I am not sure if that is the case. Needs some investigation. This may also belong in a different section since it's not DoS, it's brute force.
```
bigshebang, line 175: Also would love some help from a more experienced GraphQL person or somebody who has a test setup handy. If not I will eventually spend some time to mess with this myself and do further research.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Yeah I wanted something more generic than "APIs" because I think that term is too heavily used to reference HTTP services and didn't want to be confusing. But your new suggestion works great! I'll add that in.

PauloASilva (approved):
We can use "safe programming interfaces" instead of "safe tools" or "safe APIs".

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+app.use('/graphql', graphqlHTTP({
+  schema: MySessionAwareGraphQLSchema,
++ validationRules: [NoIntrospection]
+  graphiql: process.env.NODE_ENV === 'development',
+}));
+```
+
+### IDOR Protection
+
+In general, proper access controls will prevent any IDOR attacks since IDOR merely represents 2 separate authorization issues: [Broken Object Level Authorization](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md) and [Broken Function Level Authorization](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa5-broken-function-level-authorization.md). See the [IDOR prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html) for a comprehensive rundown of preventing IDOR.
```
PauloASilva, line 227: I am not sure whether this is clear enough, but I'll wait for others' feedback.

"Classic" IDOR is in fact both BOLA and BFLA, but considering the GraphQL scope I would say that BFLA is an access control issue specific to GraphQL mutations (functions).

May be we can add something about BFLA in the "Mutation Access (Data Manipulation)" section.

PauloASilva (changes requested):
A full document review.
We're getting closer and closer to a final draft: great work @bigshebang!

PauloASilva (approved):

cheatsheets_draft/GraphQL_Cheat_Sheet.md
inonshk: I would argue that "IDOR merely represents 2 separate authorization issues".

When you have a case of BFLA: for example, an admin function, "export_all_users", that doesn't implement access control, so regular users can access it - I don't think anybody would call it IDOR.
On top of that, the protection against IDOR/BOLA, is very different than protection against BFLA.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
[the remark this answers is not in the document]
inonshk: Just to add one more point: when exploiting SSRF, we don't really break the syntax of any command/query/code.
Hence, I wouldn't say that SSRF is a type of injection

mackowski:
@nikitastupin do you want to review this updated PR?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Yeah I agree @inonshk, BFLA should not be called out as IDOR. I added this section based on feedback without thinking too much about it. If a user can access a function they shouldn't have access to that just should not be called IDOR in my mind.

However I do think BOLA is a traditional IDOR. The whole reason I initially added IDOR to this CS is because GraphQL usually involves the caller providing object IDs directly to access them. But I am thinking I will remove this IDOR section. Then I will add a sub bullet point to the first bullet point under the "Access Control" section that says "This will prevent [IDOR](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html)/[BOLA](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md) issues." I could also add a new sentence to the "Mutation Access (Data Manipulation)" section and mention that doing proper checks for access to mutation would prevent [BFLA](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa5-broken-function-level-authorization.md), but I don't think it's necessary to add.

Thoughts on that?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+- Some default configurations (Introspection, GraphiQL) should be disabled/changed before releasing an API to production
+
+## Common Attacks
+
+- [Injection](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa8-injection.md) - this usually includes but is not limited to:
+    - [SQL](https://owasp.org/www-community/attacks/SQL_Injection) and [NoSQL](https://www.netsparker.com/blog/web-security/what-is-nosql-injection/) injection
+    - [OS Command injection](https://owasp.org/www-community/attacks/Command_Injection)
+    - [SSRF](https://portswigger.net/web-security/ssrf) and [CRLF](https://owasp.org/www-community/vulnerabilities/CRLF_Injection) [injection](https://www.acunetix.com/websitesecurity/crlf-injection/)/[Request](https://portswigger.net/web-security/request-smuggling) [Smuggling](https://www.pentestpartners.com/security-blog/http-request-smuggling-a-how-to/)
+- [DoS](https://owasp.org/www-community/attacks/Denial_of_Service) ([Denial of Service](https://www.cloudflare.com/learning/ddos/glossary/denial-of-service/))
+- [IDOR](https://portswigger.net/web-security/access-control/idor)
```
nikitastupin, line 19: How about moving IDOR to Broken Authorization section or removing? I think IDOR issues are pretty much covered by BOLA (Broken Object Level Authorization) you mentioned below 😃

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
++ validationRules: [NoIntrospection]
+  graphiql: process.env.NODE_ENV === 'development',
+}));
+```
+
+### IDOR Protection
+
+In general, proper access controls will prevent any IDOR attacks since IDOR merely represents 2 separate authorization issues: [Broken Object Level Authorization](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md) and [Broken Function Level Authorization](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa5-broken-function-level-authorization.md). See the [IDOR prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html) for a comprehensive rundown of preventing IDOR.
+
+> New section from nikitastupin. Not something I can personally verify with my limited GraphQL experience, so would be great if someone else can check this out.
```
nikitastupin, line 229: You can check GitHub's GraphQL schema https://api.github.com/graphql, they have such a field.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+
+Sometimes there are `node` or `nodes` or both fields in a query object. They can be used to access objects directly by `ID` which often introduces authorization vulnerabilities. You can check whether your schema has these fields by running this on the command line (assuming that `schema.json` contains your GraphQL schema): `cat schema.json | jq ".data.__schema.types[] | select(.name==\"Query\") | .fields[] | .name" | grep node`. To prevent such vulnerabilities either remove these fields or apply proper authorization checks when objects accessed directly by `ID` as discussed earlier in the "Access Control" section.
+
+## Other Resources
+
+### Tools
+
+- [InQL Scanner](https://github.com/doyensec/inql) - Security scanner for GraphQL. Particularly useful for generating queries and mutations automatically from given schema and them feeding them to scanner.
```
nikitastupin, line 237: InQL isn't really a scanner but rather a tool that helps to explore particular GraphQL API capabilities

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+
+Or you could get somewhat complex with throttling and implement it in your code (non-trivial). See "Throttling" [here](https://www.howtographql.com/advanced/4-security/) for more about GraphQL-specific rate limiting.
+
+#### Server-side Batching and Caching
+
+To increase efficiency of a GraphQL API and reduce its resource consumption, [the batching and caching technique](https://graphql.org/learn/best-practices/#server-side-batching-caching) can be used to prevent making duplicate requests for pieces of data within a small time frame. Facebook's [DataLoader](https://github.com/facebook/dataloader) tool is one way to implement this.
+
+#### OS/Container Resource Management
+
+> Should we remove this section since it's not specific to GraphQL?
```
nikitastupin, line 167: Yeah, let's keep it for another cheat sheet? 😄

nikitastupin (commented):
This version of cheat sheet is quite good. I've left some comments but they aren't blockers and may be fixed later. I suggest to introduce this cheat sheet now and make changes to it later on step by step 😃

cheatsheets_draft/GraphQL_Cheat_Sheet.md
PauloASilva: Although I do agree this is not GraphQL specific, I think we should keep the recommendation. In the OWASP API Security Top 10 2019 this issue comes 4th.

You may argue that Query Limiting partially addresses the issue, but from my experience, quite often it is not done properly and in such cases Resources Limiting may prevent DoS.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,257 @@
...
+
+Not properly limiting the amount of resources your API can use (e.g. CPU or memory), may compromise your API responsiveness and availability, leaving it vulnerable to DoS attacks. This can be done at the operating system level with blank and blank. However, containerization platforms tend to make this task much [easier](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts): see how to limit [memory](https://docs.docker.com/config/containers/resource_constraints/#memory), [CPU](https://docs.docker.com/config/containers/resource_constraints/#cpu), [number of restarts](https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart), [file descriptors, and processes](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit) using Docker.
```
PauloASilva, line 169:
```suggestion
Not properly limiting the amount of resources your API can use (e.g. CPU or memory), may compromise your API responsiveness and availability, leaving it vulnerable to DoS attacks. This can be done at the operating system level using, for example, [Control groups (cgroups)](https://en.wikipedia.org/wiki/Cgroups). However, containerization platforms tend to make this task much [easier](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts): see how to limit [memory](https://docs.docker.com/config/containers/resource_constraints/#memory), [CPU](https://docs.docker.com/config/containers/resource_constraints/#cpu), [number of restarts](https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart), [file descriptors, and processes](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit) using Docker.
```

PauloASilva (changes requested):
Add reference to Control Groups (cgroups) as an example of how to limit process resources at OS level.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: In the Readme the first title is "InQL Scanner" so I was using that. I think it's more succinct than "Security testing tool" and conveys about the same meaning. I prefer to leave it as is. Any strong arguments/feelings against?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: If I do end up removing the IDOR section I would probably leave this section out too. Maybe I could add the below note to the "Access Control" section in place of this section:

> Be careful with the `node` or `nodes` fields in query objects because they can be used to access objects directly by `ID`. Part of proper authorization checks includes ensuring that the requester has access to the objects they are requesting by ID. The requester supplying the ID of an object is not sufficient to prove they have access to it.

Maybe I will also include the command to check if the schema has these fields.

Thoughts on that?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I do think this recommendation is important to have. Ideally this topic would be covered well in the [Web Service Security CS](https://cheatsheetseries.owasp.org/cheatsheets/Web_Service_Security_Cheat_Sheet.html#availability), since that can be thought of as a parent to this CS (and the REST Security CS). @PauloASilva maybe that would be a task for you being on the API security team? :) Or I could get a PR up to add something small to that CS. Then in this CS I can just link to that content.

I think I will remove the Docker specific items here since they are ripped straight out of the Docker Security CS. I will probably change that part to the below blurb instead:

> However, containerization platforms tend to make this task much easier. See the resource limiting section in the [Docker Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts) for how to prevent DoS when using containers.

I will leave in the OS part about cgroups, ulimits, and LXC for now.

PauloASilva: > Ideally this topic would be covered well in the Web Service Security CS, since that can be thought of as a parent to this CS (and the REST Security CS).

I see your point. Unfortunately the Web Service Security CS is not as practice as this one, which I like most.
I'll open a PR to add a "Resources Limiting" section under "Availability".

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -166,7 +166,7 @@ To increase efficiency of a GraphQL API and reduce its resource consumption, [th
...
+Not properly limiting the amount of resources your API can use (e.g. CPU or memory), may compromise your API responsiveness and availability, leaving it vulnerable to DoS attacks. This can be done at the operating system level with a combination of [Control Groups(cgroups)](https://en.wikipedia.org/wiki/Cgroups), [User Limits (ulimits)](https://linuxhint.com/linux_ulimit_command/), and [Linux Containers (LXC)](https://linuxcontainers.org/lxc/security/). However, containerization platforms tend to make this task much [easier](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts): see how to limit [memory](https://docs.docker.com/config/containers/resource_constraints/#memory), [CPU](https://docs.docker.com/config/containers/resource_constraints/#cpu), [number of restarts](https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart), [file descriptors, and processes](https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit) using Docker.
```
PauloASilva, line 169: I do agree removing the Docker specific items.

```suggestion
Not properly limiting the amount of resources your API can use (e.g. CPU or memory), may compromise your API responsiveness and availability, leaving it vulnerable to DoS attacks. This can be done at the operating system level with a combination of [Control Groups(cgroups)](https://en.wikipedia.org/wiki/Cgroups), [User Limits (ulimits)](https://linuxhint.com/linux_ulimit_command/), and [Linux Containers (LXC)](https://linuxcontainers.org/lxc/security/). However, containerization platforms tend to make this task much [easier](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts).
```

PauloASilva (changes requested):

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: This section is basically just a condensed version of the major sections below. I'll change this to reflect whatever changes are made in the major sections.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Gonna resolve this since I've implemented the feedback from this thread.

mackowski:
Hey @inonshk @nikitastupin, are you still around? @bigshebang answered some of your comments after the review can you answer back and continue the discussion or resolve the conversation?
I would like to merge this PR at some pont, there is a lot of good stuff. Please review it.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I'm gonna move forward with this change since I haven't heard anything.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Gonna resolve since I haven't heard back. I'm confident in sticking with the wording from the official repo of the tool.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I'm trying to find some notes in the documentation about how to disable the node or nodes functionality, but it seems like the docs say all GraphQL APIs must support this: https://graphql.org/learn/global-object-identification/. Do you have any documentation or examples that show disabling the node or nodes field?

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,269 @@
...
+
+The safest and usually easiest approach is to just disable introspection and GraphiQL system-wide. See [this page](https://lab.wallarm.com/why-and-how-to-disable-introspection-query-for-graphql-apis/) or consult your GraphQL implementation's documentation to learn how to disable introspection altogether. If your implementation does not natively support disabling introspection or if you would like to allow some consumers/roles to have this access you can build a filter in your service to only allow approved consumers to access the introspection system.
+
+> This blurb is from nikitastupin. Would be great if we knew the name of this "hint" feature and link to documentation to disable it. I couldn't find anything from a quick Google.
+
+Keep in mind that even if introspection is disabled you can still guess fields by brute forcing them. Furthermore, some GraphQL implementations give you a hint when a field name you provide is similar to an existing field (e.g. you provide `usr` and the response may ask if you meant the valid `user` field instead). You should consider disabling this feature.
```
bigshebang, line 224: I looked at this a bit more and it seems like this hint feature is just standard/built-in for GraphQL. I've been looking at how to actually disable this and am not sure how just yet. Gonna keep doing more research but welcome any help. So far these couple pages might have an answer on this: 

* https://www.apollographql.com/docs/apollo-server/data/errors/
* https://www.howtographql.com/graphql-java/7-error-handling/
* https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/
* https://www.apollographql.com/docs/react/data/error-handling/

bigshebang:
I took another look at some stuff and made a change with IDOR. Now there are just 3 issues I want to iron out and this will be ready to go!

cheatsheets_draft/GraphQL_Cheat_Sheet.md
PauloASilva: GraphQL Apollo server field name hints are provided by [GraphQL.js].
It seems that there's no simple and easy way to disable this feature (graphql/graphql-js#2247) other than using some middleware to prevent that information to leak.

Since this is implementation-dependent, I think we're good to go with a general recommendation since it will create the required awareness how the feature can be abused: [shapeshifter] is just one tool that [should be able to do it][1].

[shapeshifter]: https://github.com/szski/shapeshifter
[1]: https://www.youtube.com/watch?v=NPDp7GHmMa0&t=2580

[GraphQL.js]: https://github.com/graphql/graphql-js

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -191,12 +190,21 @@ Not properly limiting the amount of resources your API can use (e.g. CPU or memo
...
 - Disable [GraphiQL](https://github.com/graphql/graphiql) and other similar schema exploration tools in production or publicly accessible environments.
 
+#### General Data Access
+
+It's commonplace for GraphQL requests to include one or more direct IDs of objects in order to fetch or modify them. For example, a request for a certain picture may include the ID that is actually the primary key in the database for that picture. As with any request, the server must verify that the caller has access to the object they are requesting. But sometimes developers make the mistake of assuming that possession of the object's ID means the caller should have access. Failure to verify the requester's access in this case is called [Broken Object Level Authentication](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md), which is a subset of a weakness called [IDOR](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html).
```
PauloASilva, line 202: We can no say that BOLA is a subset of IDOR.

IDOR **is not equal to** BOLA + BFLA: we do not call IDOR when an unauthorized user is example to access `api.somehost.com/admin/users/export` (this is BFLA).

```suggestion
It's commonplace for GraphQL requests to include one or more direct IDs of objects in order to fetch or modify them. For example, a request for a certain picture may include the ID that is actually the primary key in the database for that picture. As with any request, the server must verify that the caller has access to the object they are requesting. But sometimes developers make the mistake of assuming that possession of the object's ID means the caller should have access. Failure to verify the requester's access in this case is called [Broken Object Level Authentication](https://github.com/OWASP/API-Security/blob/master/2019/en/src/0xa1-broken-object-level-authorization.md), also known as [IDOR](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html).
```

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -191,12 +190,21 @@ Not properly limiting the amount of resources your API can use (e.g. CPU or memo
...
+
+> New section from nikitastupin. Not something I can personally verify with my limited GraphQL experience, so would be great if someone else can check this out.
+
+It's possible for a GraphQL API to support access to objects using their ID even if that is not intended. Sometimes there are `node` or `nodes` or both fields in a query object, and these can be used to access objects directly by `ID`. You can check whether your schema has these fields by running this on the command line (assuming that `schema.json` contains your GraphQL schema): `cat schema.json | jq ".data.__schema.types[] | select(.name==\"Query\") | .fields[] | .name" | grep node`. You can remove these fields from the schema to disable the functionality. Or as stated above you can apply proper authorization checks to verify the caller has access to the object they are requesting.
```
PauloASilva, line 206: "apply proper authorization checks" should not come as an "or".

```suggestion
It's possible for a GraphQL API to support access to objects using their ID even if that is not intended. Sometimes there are `node` or `nodes` or both fields in a query object, and these can be used to access objects directly by `ID`. You can check whether your schema has these fields by running this on the command line (assuming that `schema.json` contains your GraphQL schema): `cat schema.json | jq ".data.__schema.types[] | select(.name==\"Query\") | .fields[] | .name" | grep node`. Removing these fields from the schema should disable the functionality, but you should always apply proper authorization checks to verify the caller has access to the object they are requesting.
```

PauloASilva (changes requested):

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Great catch! Your suggestion is way better.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: I would like to actually test this out to confirm whether or not it works before adding a firm recommendation. Again, any help is appreciated :) If not I will get to it at some point.

We could also publish the CS without this point and have it as an issue to add later.

bigshebang: Ah I didn't see your message until now Paulo. I will add the information you gave which will resolve this issue.

PauloASilva: > I would like to actually test this out to confirm whether or not it works before adding a firm recommendation. Again, any help is appreciated :) If not I will get to it at some point.
> 
> We could also publish the CS without this point and have it as an issue to add later.

I did test shapeshifter. Although it provides the feature demonstrated in the video, it only works if introspection is also enabled, what is not required to take advantage of GraphQL server field name hints.

I am OK with your changes regarding this specific topic: I think it brings enough awareness and guidance regarding what should be enabled/disabled in a production environment.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,302 @@
...
+
+```javascript
+app.use('/graphql', graphqlHTTP({
+  schema: MySessionAwareGraphQLSchema,
++ validationRules: [NoIntrospection]
+  graphiql: process.env.NODE_ENV === 'development',
+}));
+```
+
+### Batching Attacks
```
bigshebang, line 223: This is the only section that is brand new. Would love a look from you @PauloASilva

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -232,6 +220,42 @@ app.use('/graphql', graphqlHTTP({
...
+    name
+  }
+  second:droid(id: "2001") {
+    name
+  }
+  third:droid(id: "2002") {
+    name
+  }
+}
+```
```
PauloASilva, line 239: This is not the most common syntax I have found: arrays of GraphQL queries are more common since they are easier to manage by the client

```
[
  {
    query: < query 0 >,
    variables: < variables for query 0 >,
  },
  {
    query: < query 1 >,
    variables: < variables for query 1 >,
  },
  {
    query: < query n >
    variables: < variables for query n >,
  }
]
```

Apollo server provides [comprehensive documentation on Query batching](https://www.apollographql.com/blog/query-batching-in-apollo-63acfd859862/).

PauloASilva (changes requested):

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Okay I will add this as another example of batching queries. I'll also add your Apollo server link.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -232,6 +220,42 @@ app.use('/graphql', graphqlHTTP({
...
+
+In this case it could be used to enumerate every possible `droid` object that is stored on the server in very few network requests as opposed to a standard REST API where the requester would need to submit a different network request for every different `droid` ID they want to request. This type of attack can lead to the following issues:
+
+- Enumerating every instance of an object on the server, such as users, emails, and user IDs.
+- Brute forcing passwords, 2 factor authentication codes (OTPs), session tokens, or other sensitive values.
+- WAFs, RASPs, IDS/IPS, SIEMs, or other security tooling will likely not detect these attacks since they only appear to be one single request rather than an a massive amount of network traffic.
+- This attack will likely bypass existing rate limits in tools like Nginx or other proxies/gateways since they rely on looking at the raw number of requests.
```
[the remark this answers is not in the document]
bigshebang: I added this in my own commit with slightly different wording

PauloASilva:
I am happy with the overall result/content.
I think it is time to get it merged :rocket: 

Cheers,
Paulo A. Silva

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+***Disable Introspection & GraphiQL - JavaScript***
+
+```javascript
+app.use('/graphql', graphqlHTTP({
+  schema: MySessionAwareGraphQLSchema,
++ validationRules: [NoIntrospection]
+  graphiql: process.env.NODE_ENV === 'development',
+}));
+```
```
ThunderSon, line 221: Why is introspection discussed in mutation access? This has been discussed elsewhere (if not it should be in a different section)

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Batching Attacks
```
ThunderSon, line 223: This section can be trimmed down and split between here and the WSTG scenario being written.
This is not a blocker for now over here (IMO, I don't have a big say here).
Once the test scenario is better in the WSTG, this CS can be cleaned up a bit from the attacks and use references across, and mainly focus on mitigations.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+- DoS attacks - A high number of queries or object requests in a single network call could cause a database to hang or exhaust other available resources (e.g. memory, CPU, downstream services).
+- Enumerating every instance of an object on the server, such as users, emails, and user IDs.
+- Brute forcing passwords, 2 factor authentication codes (OTPs), session tokens, or other sensitive values.
+- WAFs, RASPs, IDS/IPS, SIEMs, or other security tooling will likely not detect these attacks since they only appear to be one single request rather than an a massive amount of network traffic.
+- This attack will likely bypass existing rate limits in tools like Nginx or other proxies/gateways since they rely on looking at the raw number of requests.
+
+#### Mitigating Batching Attacks
+
+In order to mitigate this type of attack you should put limits on incoming requests at the code level so that they can be applied per request. There are 2 main options:
```
ThunderSon, line 270: There are more than 2 main options now after the latest update :)

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+- Prevent batching for sensitive objects
+- Add object request rate limiting in code
+- Limit the number of queries that can run at one time
+
+One option is to prevent batching for sensitive objects that you don't want to be brute forced, such as usernames, emails, passwords, OTPs, session tokens, etc. This way an attacker is forced to attack the API like a REST API and make a different network call per object instance. This is not supported natively so it will require a custom solution. However once this control is put in place other standard controls will function normally to help prevent any brute forcing.
+
+Another option is to create a code-level rate limit on how many objects that callers can request. This means the backend would track how many different object instances the caller has requested, so that they will be blocked after requesting too many objects even if they batch the object requests in a single network call. This replicates a network-level rate limit that a WAF or other tool would do.
```
ThunderSon, line 278: I believe this is the easiest one between the 3 options (and effective at the same time), and thus should be the first recommendation.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+Limiting the number of operations that can be batched and run at once, is another option to mitigate GraphQL batching attacks leading to Denial of Service (DoS).
```
ThunderSon, line 280:
```suggestion
Limiting the number of operations that can be batched and run at once is another option to mitigate GraphQL batching attacks leading to DoS. This is not a silver bullet though and should be used in conjunction with other methods.
```

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Secure Configurations
+
+By default, most GraphQL implementations have some insecure default configurations which should be changed:
+
+- Disable/restrict introspection + GraphiQL
```
ThunderSon, line 286:
```suggestion
- Disable or restrict based on your needs Introspection and GraphiQL which should only be used for development purposes.
```

ThunderSon (changes requested):
Other than the review done inline, just 2 more things:
- Cheatsheets links shouldn't be to the cheatsheetseries website. They should just call the other file (I did an example in my suggestions). The engine will resolve it.
- Referencing WSTG shouldn't be done on the "latest" branch. It should either use a version (v41 for example), or "stable".

Well done overall, I am just adding some CS magic touchup, not much. You guys did a fantastic job. As a v1 this is a job well done, it can be refined in next versions as it's better adopted across other guides.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
+***JavaScript Timeout Example***
+
+Code snippet from [this SO answer](https://stackoverflow.com/a/53277955/1200388):
+
+```javascript
+request.incrementResolverCount =  function () {
+    var runTime = Date.now() - startTime;
+    if (runTime > 10) {  // a timeout of 10 seconds
+      if (request.logTimeoutError) {
+        logger('ERROR', 'Request ' + request.uuid + ' query execution timeout');
```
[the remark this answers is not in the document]
bigshebang: Yeah probably, I'm no javascript expert and I did not personally test any of the non-GraphQL code snippets. But most are from established/trusted sources.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,325 @@
...
... to DoS attacks. Some limiting can be done at the operating system level with a combination of [Control Groups(cgroups)](https://en.wikipedia.org/wiki/Cgroups), [User Limits (ulimits)](https://linuxhint.com/linux_ulimit_command/), and [Linux Containers (LXC)](https://linuxcontainers.org/lxc/security/) when using Linux. There are also some options that [may work](https://docs.microsoft.com/en-us/windows/win32/procthread/processor-groups) on [Windows](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/ff384148(v=ws.10)), both [native](https://blogs.iis.net/thomad/put-the-brakes-on-your-application-pools-cpu-rate-limits-in-windows-7) and [third](https://superuser.com/questions/214566/are-there-solutions-that-can-limit-the-cpu-usage-of-a-process) [party](https://github.com/lowleveldesign/process-governor). However, containerization platforms tend to make this task much easier. See the resource limiting section in the [Docker Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-7-limit-resources-memory-cpu-file-descriptors-processes-restarts) for how to prevent DoS when using containers.
```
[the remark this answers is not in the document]
bigshebang: Yeah I hastily grabbed the first windows resources I could find, but yeah those OSes are deprecated now so I'll remove.

I'll also break this paragraph out to be more digestible.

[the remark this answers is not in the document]
bigshebang: Actually I just removed the Windows section completely. Most of the links are old and I have never had any experience with them. We can add Windows recommendations at another time.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Great catch. These are under the wrong section so I'll move them.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Makes sense. I'll leave as is for now then.

cheatsheets_draft/GraphQL_Cheat_Sheet.md
bigshebang: Works for me!

bigshebang:
@ThunderSon Thanks for the great feedback, I've gone through it all and incorporated everything. Please give it another look and give me the final okay.

ThunderSon (dismissed):
Amazing job! :smile:

cheatsheets_draft/GraphQL_Cheat_Sheet.md
```diff
@@ -0,0 +1,333 @@
...
+At the application level, timeouts can be added for queries and resolver functions. This option is usually more effective since the query/resolution can be stopped once the timeout is reached. GraphQL does not natively support query timeouts so custom code is required. See [this blog post](https://medium.com/workflowgen/graphql-query-timeout-and-complexity-management-fab4d7315d8d) for more about using timeouts with GraphQL or the two examples below.
+
+***JavaScript Timeout Example***
+
+Code snippet from [this SO answer](https://stackoverflow.com/a/53277955/1200388):
+
+```javascript
+request.incrementResolverCount =  function () {
+    var runTime = Date.now() - startTime;
+    if (runTime > 10) {  // a timeout of 10 seconds
```
PauloASilva, line 126: Since you're using [`Date.now()`][1], you'll getting a timestamp in milliseconds, meaning that the difference between `Date.now()` and `startTime` should also be in milliseconds.

```suggestion
    var runTime = Date.now() - startTime;
    if (runTime > 10000) {  // a timeout of 10 seconds
```


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

ThunderSon: @bigshebang you almost DoSed the service :yum:

mackowski (approved):
I love it! Thank you all for this work! 
If no one have anything against I will merge it tomorrow :) Any other issues can be addressed via separate issues/PR.

PR #436, opened 2020-07-02
Add recommendation to disable xp_dirtree and CLR execution

rbsec:
Redis "broken" link is a false positive.


rbsec:
@ThunderSon we don't have specific commands to make the other changes, so if we're going to them here it might look a bit odd.

PR #441, opened 2020-07-11
Update Transport_Layer_Protection_Cheat_Sheet.md

tabarra:
This PR covers issue #282.

### DIFF
--- a/cheatsheets/Transport_Layer_Protection_Cheat_Sheet.md
+++ b/cheatsheets/Transport_Layer_Protection_Cheat_Sheet.md
@@ -182,7 +182,7 @@ HTTP Strict Transport Security (HSTS) instructs the user's browser to always req
 
 HTTP Public Key Pinning (HPKP) is used to associate a specific certificate with a domain name, in order to prevent an attacker performing a man in the middle attack with a different (but trusted) certificate. See the [Certificate and Public Key Pinning article](https://owasp.org/www-community/controls/Certificate_and_Public_Key_Pinning#http-pinning) for further information in implementing HPKP.
 
-HPKP is controversial, and that while it is currently supported in Firefox and some other browsers, it has been removed from [Chrome](https://www.chromestatus.com/feature/5903385005916160), and is considered deprecated by [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Public-Key-Pins). As such, careful consideration of the risks should be made before implementing it.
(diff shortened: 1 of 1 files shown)


rbsec:
Hi @tabarra.

When I previously updated this CS HPKP still had some support, although that seems to no longer be the case. Your proposed changes say that HPKP isn't supported by any major browser, and has a load of problems with it that make it dangerous.

@jmanico (or anyone else) - are there any circumstances in where you would recommended HPKP in 2020, or any specific cases where it's still useful?

If it's something that's not useful (or even is harmful) in the majority of cases, and doesn't have any real benefits then we should just remove it.

tabarra:
As I can see, this *might* still useful for "Internal CAs" (i.e. company's internal CA used only for intranet stuff) since it *might* not possible to use Certificate Transparency.
  
Also, HPKP Suicide and Ransom PKP is probably not that big of a deal if it the server and client are in the same organization, so they can communicate with each other and solve the issue.  
But, for the reason above, it is a bit redundant since inside an organization it is way easier for the server and consumer systems already having another mean of key pinning that doesn't rely on HPKP at all.

ThunderSon:
Interesting scenario. If I have an Internal CA, I'll be setting the root certificate of that CA across my systems, since the CA is trusted.

From my point of view, this thing no longer has any value as most browsers have disabled it, so there's nothing to apply the header on the client side.

rbsec:
I hadn't thought about internal CAs - that's an interesting use case (although not one I've ever seen).

I suppose that you could either force all your clients to stick to an older browser that supports it (or re-enable it with about:config or chrome://flags?) and use it internally that way - but this feels like it's too much of an edge case to include in a cheat sheet.

I think we should just remove the whole section - any objections to that?

jmanico:
Agree we should fully remove it since it’s a depreciated standard....

--
Jim Manico
@Manicode

tabarra:
Perhaps I should go a step further and change `Consider Using Public Key Pinning` to `Notes On Using Public Key Pinning`, where I make it clear it is not recommended for most cases?  
Since you guys are more in the line of "remove it" instead of just adding a note like I did.  
Let me know what you think.

jmanico:
Public Key Pinning is still important in intra-server, Mobile and non web browser situations. I think we should remove the HPKP response header but keep the pinning section.

--
Jim Manico
@Manicode

tabarra:
any updates on the subject?

rbsec:
Hi,

Based on the discussion here, I've proposed some updated text in #447.

tabarra:
Solved in #447

PR #447, opened 2020-07-19
Update recommendation for public key pinning. Fixes #441

rbsec:
As per the discussion in #441, a proposed update to the recommendations about HPKP.


ThunderSon:
@tabarra does this answer your issue?

tabarra:
I think its fine.  
Thanks guys!

Issue #445, opened 2020-07-15
Update: Clarify guidance around security headers for REST APIs


dotnetnate:
## What is missing or needs to be updated?
The referenced cheat sheet requires that REST APIs return x-frame-options and set the values to deny. This is not meaningful or useful in the context of a REST API and seems to serve nothing more than to cause extra work and contention when having to explain this to security teams blindly reading vulnerability scan reports that are based off of OWASP documentation without understanding the context. The only possibly feasible manner in which this becomes relevant is if an API is returning HTML content for direct rendering in the browser. 

## How should this be resolved?
The statement to include x-frame-options header should be removed or make it optional, and only for APIs that return HTML, which a scanner should be able to detect via regex.

rbsec:
This is an interesting area - thanks for raising it. I think that the [Security Headers](https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html#security-headers) section in that cheat sheet could do with a rewrite, because this is an question that I've run into a few times.

There are some security headers that are important and should be sent by almost every REST API (such as HSTS). Others like CSP or X-Frame-Options are less useful, although they may still have some benefits depending on exactly what the API returns (which isn't always known).

Not all client will support them (browsers support HSTS, but not many other clients will), but I still think there's some benefit to setting them unless you can guarantee which clients will be used to access the API.

My view would be that you should set the same security headers on an API endpoint as you would on a website (especially if it doesn't have a dedicated domain) - because these headers are something that should be there by default, and while they might not always provide a security benefit, they'll never make your security worse.

If you can guarantee that your clients will never support them, or that there are no circumstances where they would provide a security benefit (such as if your API will never return HTML, including future versions) then that would be reasonable grounds to decide to disable them - but I think that the justification should be that way round ("why should I disable this security feature", rather than "why should I enable it").

It also stops people/scanning tools complaining, and although that's not a good reason to do something, in the real world it can end up being a reasonable justification for it.

I'm not aware of many authoritative sources for this (although as usual there's some discussion on [Security Stack Exchange](https://security.stackexchange.com/questions/147554/security-headers-for-a-web-api)) - so it might be a nice one for us to provide some clear guidance on. There's also a [few](https://observatory.mozilla.org/faq/) different [places](https://infosec.mozilla.org/guidelines/web_security#content-security-policy) where Mozilla are recommending a CSP for API endpoints.

@ThunderSon @mackowski @jmanico what are your views on this?

[...]

PR #448, opened 2020-07-20
Update guidance on security headers for REST APIs

rbsec:
As per the discussion in #445.

### DIFF
--- a/cheatsheets/REST_Security_Cheat_Sheet.md
+++ b/cheatsheets/REST_Security_Cheat_Sheet.md
@@ -149,9 +149,9 @@ The headers below are only intended to provide additional security when response
 
 | Header | Rationale |
 |--------|-----------|
-| `Content-Security-Policy` | The majority of CSP functionality only affects pages rendered as HTML. |
-| `Feature-Policy` | Feature policies only affect pages rendered as HTML. |
-| `Referred-Policy` | Non-HTML responses should not trigger additional requests. |
+| `Content-Security-Policy: default-src 'none'` | The majority of CSP functionality only affects pages rendered as HTML. |
+| `Feature-Policy: 'none'` | Feature policies only affect pages rendered as HTML. |
+| `Referred-Policy: no-referrer` | Non-HTML responses should not trigger additional requests. |
 
 ## CORS
 
--- a/cheatsheets/REST_Security_Cheat_Sheet.md
+++ b/cheatsheets/REST_Security_Cheat_Sheet.md
@@ -139,6 +139,7 @@ The following headers should be included in all API responses:
 
 | Header | Rationale |
 |--------|-----------|
+| `Cache-Control: no-store` | Prevent sensitive information from being cached. |
 | `Content-Security-Policy: frame-ancestors 'none'` | To protect against drag-and-drop style clickjacking attacks. |
 | `Content-Type` | To specify the content type of the response. This should be `application/json` for JSON responses. |
 | `Strict-Transport-Security` | To require connections over HTTPS and to protect against spoofed certificates. |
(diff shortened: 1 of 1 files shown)


cheatsheets/REST_Security_Cheat_Sheet.md
```diff
@@ -131,11 +131,27 @@ Services including script code (e.g. JavaScript) in their responses must be espe
...
-## Security headers
+## Security Headers
 
-To make sure the content of a given resources is interpreted correctly by the browser, the server should always send the `Content-Type` header with the correct content type, and preferably the `Content-Type` header should include a charset. The server should also send the `X-Content-Type-Options: nosniff` [security header](https://owasp.org/www-project-secure-headers/) to make sure the browser does not try to detect a different `Content-Type` than what is actually sent (can lead to XSS).
+There are a number of [security related headers](https://owasp.org/www-project-secure-headers/) that can be returned in the HTTP responses to instruct browsers to act in specific ways. However, some of these headers are intended to be used with HTML responses, and as such may provide little or no security benefits on an API that does not return HTML.
 
-Additionally the server should send the `X-Frame-Options: deny` [security header](https://owasp.org/www-project-secure-headers/) to protect against drag'n drop clickjacking attacks in older browsers.
+The following headers should be included in all API responses:
+
+| Header | Rationale |
```
ThunderSon, line 140: Part of these should be the `Cache-Control`, as it's important to be private without caching for user data.

cheatsheets/REST_Security_Cheat_Sheet.md
```diff
@@ -131,11 +131,27 @@ Services including script code (e.g. JavaScript) in their responses must be espe
...
+|--------|-----------|
+| `Content-Security-Policy: frame-ancestors 'none'` | To protect against drag-and-drop style clickjacking attacks. |
```
ThunderSon, line 142: This is not required if HTML is not returned, or any "functionality" that could be triggered, so framing protection should be part of the second table, no?

rbsec: There's some interesting variants on clickjacking where you get to the user to select text and then drag it into a text box (that you control) so you can read the contents. That could still potentially be exploitable against an endpoint that returns JSON/text:

https://www.w3.org/Security/wiki/Clickjacking_Threats#Drag_and_drop_attacks

cheatsheets/REST_Security_Cheat_Sheet.md
rbsec: Good call, added in 84acb9f4.

ThunderSon:
@dotnetnate does this answer your issue? :)

mackowski:
@rbsec please do not merge it yet. I would like to review it to check if we are not missing any case.

cheatsheets/REST_Security_Cheat_Sheet.md
```diff
@@ -131,11 +131,28 @@ Services including script code (e.g. JavaScript) in their responses must be espe
...
+| `X-Content-Type-Options: nosniff` | To prevent browsers from performing MIME sniffing, and inappropriately interpreting responses as HTML. |
+| `X-Frame-Options: DENY` | To protect against drag-and-drop style clickjacking attacks. |
+
+The headers below are only intended to provide additional security when responses are rendered as HTML. As such, if the API will __never__ return HTML in responses, then these headers may not be necessary. However, if there is any uncertainty about the function of the headers, or the types of information that the API returns (or may return in future), then it is recommended to include them as part of a defence-in-depth approach.
+
+| Header | Rationale |
+|--------|-----------|
+| `Content-Security-Policy: default-src 'none'` | The majority of CSP functionality only affects pages rendered as HTML. |
+| `Feature-Policy: 'none'` | Feature policies only affect pages rendered as HTML. |
+| `Referred-Policy: no-referrer` | Non-HTML responses should not trigger additional requests. |
```
koenrh, line 155:
```suggestion
| `Referrer-Policy: no-referrer` | Non-HTML responses should not trigger additional requests. |
```

PR #451, opened 2020-07-23
Update explanations around pre-hashing passwords. Fixes #450

### LINKED ISSUE #450: Update: Password Storage Cheat Sheet

The pre-hashing section contains two weird things:

### Magically obtaining an intermediate value

The context is:

> An alternative [to limiting the password length is] pre-hash[ing] the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resultant hash with a more secure algorithm such as Bcrypt (i.e, `bcrypt(sha256($password))`).

Sounds good. The next paragraph contains this:

> If an attacker is able to obtain password hashes from two different sources, one of which is storing passwords [...] as plain `sha256($password)`

How? That's like saying you might be able to get the 949th round of a 1000-round pbkdf2 operation. Sure, if you can magically obtain an intermediate value, you've got an issue. Nobody ever proposed doing one hash, storing the result, and then doing another hash, and storing that result too. And if you use different hashing schemes in different systems, of course the weaker system is, y'know, weaker.

I don't really care about this as I don't see this leading to vulnerabilities, it just seems superfluous so we might as well change it while we're at it.

### "Applying SHA-256 is potentially weaker than not applying SHA-256 because you limit the keyspace to 2^256"

... is basically what this conveys:

> Pre-hashing with SHA-256 also means that the keyspace for an attacker to brute-force the hashes is `2^256`, rather than `2^420` for passwords capped at 64 characters (although both of these are big enough to make no practical difference).

The remark at the end, "this makes no difference", already makes me consider removal for conciseness. But it also means that if the text before is taken standalone, it might be taken to mean something it does not, i.e. that it makes a difference (because why else write it?). I discovered this cheat sheet page because someone wrote this as a contra-argument for client-side hashing, saying basically:

> [Client-side hashing protects the user's password better because the server never even sees it, but for MD5,] an attacker now knows that every password on the server side is exactly 32 chars long, ranging from 0000000000000000000000000000000 - fffffffffffffffffffffffffffffff. With this knowledge an attacker only needs to brute force 2^128 possible combinations, which could be less compared to a password system that accepts input with arbitrarey length and a lot more different characters. See the OWASP Cheat Sheet on Pre Hashing for password storage.

Given that 2^128 is well beyond brute forcing capabilities of anyone, it really should not be taken as a downside that the "password" (as the server sees it) is now 32 hex chars. This text, as it was written on the Security StackExchange in a top-voted post (it was edited in after most of the upvotes), might have dissuaded developers from adding client-side hashing to client-server software (in which the server can't simply tweak the javascript to circumvent the hashing) for the wrong reason: it was based on this notion of a hash being potentially weaker than the original password. If that were true, we should find a stronger/longer hash function. Furthermore, the logical result of this conclusion would be to use base64 rather than hexadecimals (to have "more different characters") and to randomly truncated the output (so it's not "exactly 32 chars long"). It might severely weaken the password storage if someone took this at face value.

I find this part of the cheat sheet confusing at best or somewhat risky at worst and would propose for it to be removed.

Suggested text for the section `#### Pre-Hashing Passwords`:

> An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resultant hash with a more secure algorithm such as Bcrypt (i.e, `bcrypt(sha256($password))`). While this approach solves the problem of arbitrary length user inputs to slower hashing algorithms, one should ensure that the output from the first hashing algorithm is safely encoded as hexadecimal or base64, since some hashing algorithms such as Bcrypt can behave in undesirable ways if the [input contains null bytes](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html). Care should also be taken that the intermediate result is not stored in a temporary file, sent between systems, or could otherwise in any way become exposed to an attacker.

> As such, the preferred option should generally be to limit the maximum password length. Pre-hashing of passwords should only be performed where there is a specific requirement to do so, and appropriate steps have been taking to mitigate the issues discussed above.

Effectively:

- The paragraph "*Finally, when using pre-hashing ensure that the output ...*" was made to fit onto the first paragraph by writing "*..., one should ensure that the output from the first hashing algorithm [etc.]*"
- The paragraph "*Pre-hashing with SHA-256 also means ...*" (the keyspace / "no practical difference" thing) was removed since it makes no difference and can even be misleading.
- The paragraph "*If an attacker is able to obtain password hashes from two different sources...*" was replaced by the sentence "*Care should also be taken ... exposed to an attacker.*"
- The final paragraph is unchanged, I included it because this completes the subsection. It's worded quite firmly compared to the risks mentioned, but since I kind of agree that less crypto soup and simpler code (i.e. simply limiting the input length to something reasonable like 64 characters rather than "pre-hashing") is better, it might as well be left as-is. We could also add the reason for this conclusion/recommendation (it being less error-prone), but that would be a slightly bigger change than what I am proposing now and the current text isn't wrong or unfit.

Alternatively, the paragraph "*Pre-hashing with SHA-256 also means ...*" (the keyspace / "no practical difference" thing) could simply be dropped as the only change. I prefer the above version for the aforementioned reasons, but the minimum change I am looking for is dropping the superfluous text that I noticed was misleading to the person that read it.

---

Let me know what approach sounds best to you and I can prepare a pull request.

rbsec:
Based on discussion in #450 - some of the explanations around pre-hashing were unclear.

I removed the discussion about the keyspace limits from pre-hashing (which was being interpreted by someone as an argument against password hashing), and added an explanation in the maximum length section about why doesn't meaningfully reduce security.

@lgommans what do you think of these changes?

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -115,17 +115,17 @@ Additionally, due to how computationally expensive modern hashing functions are,
 
 In order to protect against both of these issues, a maximum password length should be enforced. This should be 64 characters for Bcrypt (due to limitations in the algorithm and implementations), and between 64 and 128 characters for other algorithms.
 
+Although implementing a maximum password length does reduce the possible keyspace for passwords, a limit of 64 characters still leaves a key space of at least `2^420`, which is completely infeasible for an attacker to break. As such, it does not represent a meaningful reduction in security.
+
 #### Pre-Hashing Passwords
 
 An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resultant hash with a more secure algorithm such as Bcrypt (i.e, `bcrypt(sha256($password))`). While this approach solves the problem of arbitrary length user inputs to slower hashing algorithms, it also introduces some vulnerabilities that could allow attackers to crack hashes more easily.
 
-If an attacker is able to obtain password hashes from two different sources, one of which is storing passwords with `bcrypt(sha256($password))` and the other of which is storing them as plain `sha256($password)`, and attacker can use uncracked SHA-256 hashes from the second site as candidate passwords to try and crack the hashes from the first (more secure) site. If passwords are re-used between the two sites, this can effectively allow the attacker to strip off the Bcrypt layer, and to crack the much easier SHA-256 passwords.
+If an attacker is able to obtain password hashes from two different compromised sites, one of which is storing passwords with `bcrypt(sha256($password))` and the other of which is storing them as plain `sha256($password)`, and attacker can use uncracked SHA-256 hashes from the second site as candidate passwords to try and crack the hashes from the first (more secure) site. If passwords are re-used between the two sites, this can effectively allow the attacker to strip off the Bcrypt layer, and then to focus their efforts on cracking the much easier SHA-256 hashes.
 
-Pre-hashing with SHA-256 also means that the keyspace for an attacker to brute-force the hashes is `2^256`, rather than `2^420` for passwords capped at 64 characters (although both of these are big enough to make no practical difference).
+When using pre-hashing ensure that the output for the first hashing algorithm is safely encoded as hexadecimal or base64, as some hashing algorithms such as Bcrypt can behave in undesirable ways if the [input contains null bytes](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html).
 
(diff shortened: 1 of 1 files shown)


lgommans:
@rbsec Thanks, I like these changes. Quoting from your reply in #450:

> This would allow them to determine when password are re-used between the two sites (even if they can't crack either hash), and then they're left with a list of SHA-1 hashes to focus on - effectively stripping away the layer of Bcrypt.

This is an angle I had not considered; that's a good point. I think this pull request clarifies the point, but now I've been [cursed with knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) so we'd need a new guinea pig to make sure it makes sense to those not already in the know :-). But let's just go ahead for now.

The new version of the "420 bits key space" paragraph also sounds good to me. I don't think this could still confuse anyone into thinking hashing is a bad idea due to reducing entropy.

LGTM!

PR #463, opened 2020-08-05
Fix typos plus markdown in CSRF cheatsheet

mohawk2:

Done.

This PR is motivated by https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html having a `\` after "Generate the token", which is obviously unintended, and appears to be caused by GFM making it look as intended, but not your Markdown -> HTML converter.

By the way, your PR template links to https://github.com/OWASP/CheatSheetSeries#conversion-rules but that does not seem to exist now.

> Thank you again for your contribution :smiley:

I hope it's useful!


mohawk2:
That was quick!

PR #478, opened 2020-08-31
Update Authorization CS draft

KellyMarchewa:
This PR covers issue #474

Updated one of the headers slightly per feedback in the linked issue. 
Thanks and looking forward to any discussions/feedback for improvement.

### DIFF
--- a/cheatsheets_draft/Authorization_Cheat_Sheet.md
+++ b/cheatsheets_draft/Authorization_Cheat_Sheet.md
@@ -2,33 +2,47 @@
 
 ## Introduction
 
-Briefly introduce reader to the topic and cheatsheet.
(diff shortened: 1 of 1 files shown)


mackowski:
@KellyMarchewa so far it looks awesome for me.
Please review the current Authorization CS and think what is not covered by your draft and either add it or make a comment why do you want to exclude particular topic from this CS. 
I am looking forward for next commits!

KellyMarchewa:
Thanks for the feedback @mackowski . To my understanding, this CS was supposed to be replacement/merging for both the current IDOR and Access Control CSs. Regarding the concepts covered in those sheets:

- Per an earlier discussion, it seemed to have been agreed in-depth discussions of MAC and DAC were not necessarily relevant for the scope of the CS. RBAC will necessarily be discussed under the "Prefer Feature and Attribute Based Access Control over RBAC" section. 
- Currently, I did not have plans to formally and explicitly explore the concept of permissions, but would be glad to do so if wanted.
- Far as IDOR, I think the discussion under the header "Ensure Lookup IDs are Not Accessible Even When Guessed or Cannot Be Tampered With" will perhaps most closely relate to the current IDOR CS, though other sections (such as "Validate the Permissions on Every Request") will probably also be relevant. 

Hope I covered everything. Looking forward to getting more work done on this soon.

jmanico:
I support this work, bigtime! It’s a needed change.

--
Jim Manico
@Manicode
Secure Coding Education
+1 (808) 652-3805

jmanico (approved):
Well done, I am sorry this took me 16 days to approve.

PR #481, opened 2020-09-02
Web Services Security: Add "Resources Limiting" sub-section under "Availability"

PauloASilva:
Limiting resources a web service can use will help keeping the
host system stable and mitigate attacks such as Denial-of-Service.

This is an initial draft, open for discussion, and it includes
recommendations to limit the following resources:

* CPU cycles
* Memory
* Simultaneous open files, network connections and started
  processes

The idea to add such recomendations to this Cheat Sheet was first
discussed while working in the GraphQL CS PR #434.


ThunderSon (approved):
By reviewing the doc, it's apparent that resources are not tackled anywhere else. Nice catch!

mackowski (approved):
lgtm

PR #502, opened 2020-10-16
Fix of the bug that prevents us to update new offline site

mackowski:
Fix of the bug that prevents us to update new offline site: https://github.com/OWASP/CheatSheetSeries/actions/runs/303875353


mackowski:
@jmanico can you please approve this? This is a bug that needs to be fixed to generate website.

jmanico (approved):

PR #506, opened 2020-10-22
Minor fix in crypto storage CS

kingthorin:
- The DB level bullet had mismatched brackets.
image

I don't see anything else that's off. I do note both usage of `i.e,` and `e.g,` however that's not necessarily a bad thing.


mackowski (approved):

mackowski:
@kingthorin thanks for this PR!

PR #514, opened 2020-11-05
Remove dangerous SSRF deny list python code

zsims:
The example python code for checking if an IP is private is dangerous.

It's perfectly valid in many network stacks to write IPs in octal. An example in Python 3:

```
$ nc -lv 3333
hi
```

```
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(("0177.0.0.1", 3333))
s.send('hi'.encode())
```

The example given will return `is_private_ip('0177.0.0.1') == false`

The recommendation here should be an allow list of IPs (e.g. public CIDRs perhaps), but the current deny list example is dangerous.


mackowski:
@zsims I will close this PR for now, if you still see the issue please re-open it. Thank you for raising this!

PR #516, opened 2020-11-08
XSS: remove forward slash

a8nJ:
Encoding forward slash is an unneeded usage of the encoding. The only
characters that should be encoded would be the reserved characters `&`, `<`,
`>` and `"`:

https://developer.mozilla.org/Glossary/Entity

Further these implementations from major programming language vendors dont
encode forward slash:

- https://docs.microsoft.com/dotnet/api/system.web.httputility.htmlencode
- https://golang.org/pkg/html#EscapeString
- https://php.net/function.htmlentities
- https://docs.python.org/library/html.html#html.escape
- https://ruby-doc.org/stdlib/libdoc/cgi/rdoc/CGI/Util.html#method-i-escapeHTML

I found other implementations as well with same behavior. Note that these
implementations are doing encoding, but just not with the forward slash
character. Further, its not recommended by the HTML5 specification:

https://w3.org/TR/html52/syntax.html#serializing-html-fragments

or even the HTML Living Standard:

https://html.spec.whatwg.org/multipage/parsing.html#serialising-html-fragments

### DIFF
--- a/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md
@@ -97,7 +97,7 @@ Rule \#1 is for when you want to put untrusted data directly into the HTML body 
 </div>
 ```
 
-Encode the following characters with HTML entity encoding to prevent switching into any execution context, such as script, style, or event handlers. Using hex entities is recommended in the spec. In addition to the 5 characters significant in XML (`&`, `<`, `>`, `"`, `'`), the forward slash is included as it helps to end an HTML entity.
+Encode the following characters with HTML entity encoding to prevent switching into any execution context, such as script, style, or event handlers. Using hex entities is recommended in the spec. The 5 characters significant in XML (`&`, `<`, `>`, `"`, `'`):
 
 ```text
  & --> &amp;
@@ -105,11 +105,8 @@ Encode the following characters with HTML entity encoding to prevent switching i
  > --> &gt;
  " --> &quot;
  ' --> &#x27;
- / --> &#x2F;
 
  &apos; not recommended because its not in the HTML spec (See: section 24.4.1) &apos; is in the XML and XHTML specs.
-
- Forward slash is included as it helps end an HTML entity
 ```
 
 ### RULE \#2 - Attribute Encode Before Inserting Untrusted Data into HTML Common Attributes


a8nJ:
https://github.com/OWASP/CheatSheetSeries/issues/515

jmanico (approved):
As per the OWASP Java Encoder and extensive research is done there to support encoding only the necessary characters when encoding is done correctly, I support this.

I would also ask that we build a unit test engine so we can test the encoder from any language to see how they satisfy our requirements. Well done, this is a rough one.

And as a side note, Jeff's opinion on maximum encoding is still very insightful and he might be right and this may haunt us someday.

a8nJ:
Thanks. I value dissenting opinions as well, but I think they should be backed by data, examples, links and references, not FUD and boogeymen. If I end up being wrong in this case, so be it. But at least if its discovered that I am wrong, we will know the reason, and we can document and revert this PR.

As it is now, we really dont have good justification as to why this was ever recommended, so it was either never a good idea, or the reason was forgotten.

planetlevel:
This is not about FUD and boogeymen. It's about assurance. There are different policies we could adopt.  I don't think you've justified your choice over any other the others.

1- Minimum -- probably &amp; &lt; and &quot would be sufficient in most situations
2- What everyone else does -- no known attacks that violate these rules
3- OWASP existing policy - based on characters relevant to parsing HTML, no downside to anyone
4- Maximum - encode all special characters? possible processing speed, document size downsides

In the late 1990's and early 2000s, we figured out context-sensitive encoding. I personally created this cheat sheet to capture some rules about what to encode and what not to. Before that, there was no concept of context.  I'm not trying to say we got it perfect.  I'm saying we created conservative rules that would keep developers safe even when parser implementations were changing rapidly, inconsistent, and full of bugs. Encoding all the special chars used in delimiters (3) seems like the simplest, easiest policy to follow in this environment (which we're probably still in). This policy makes closing a context (and therefore injection) impossible, at the tiny cost of encoding characters which are not strictly necessary to prevent injection in some parsers.

Maybe I can help.  Perhaps you're arguing that more developers would benefit by being able to use their platform default encoding instead of building their own, and are therefore more likely to do encoding properly.  If that's what you're thinking  I'm totally on board.

a8nJ:
> I don't think you've justified your choice

**You** are the one that needs to do the justification.

The current OWASP standing **deviates** from both the HTML5 specification, and
the HTML Living Standard, as Ive repeated noted and linked.

So in my opinion, any recommendation that deviates from the specifications, and
from at least 10 major implentation, **that recommendation needs to justify
itself**.

That hasnt happened, and likely cant happen for reasons already discussed. So
unless you have hard evidence of real and likely threats due to the issue in
question, then I dont think evidence wise, that you have a leg to stand on.
Youre welcome to your personal opinions of course, but we arent talking about
that. We are talking about what should be in the OWASP standard. And I think
what goes in the OWASP docs should defer to existing specifications unless good
reason not to.

mackowski:
I am approving it for now. 
Reasoning: There is no proof that escaping forward slash will improve defense against XSS, if all other special characters are escaped properly, but it forces developers to use non-standard implementation of the HTML escaping, what increases the risk of the mistake and makes the implementation harder.

That said I created a new issue to refresh the current XSS Prevention Cheat Sheet: https://github.com/OWASP/CheatSheetSeries/issues/517.

See my comment in the: https://github.com/OWASP/CheatSheetSeries/issues/515 with more details.

mackowski (approved):

PR #524, opened 2021-01-01
Extend Web Storage

esarafianou:
This PR is a result of the discussion in https://github.com/OWASP/ASVS/issues/843.

1. Extends the `HttpOnly` section for cookies to clearly state what an HttpOnly cookie protects and how an XSS attack is combined with a CSRF attack can use the HttpOnly cookie
2. Adds a new section about how Web Workers can be used to store secrets.


mackowski (approved):
LGTM

mackowski:
Thank you @esarafianou!

PR #528, opened 2021-01-23
Content Security Policy Cheat Sheet with more info

shaialon:

### DIFF
--- a/cheatsheets/Content_Security_Policy_Cheat_Sheet.md
+++ b/cheatsheets/Content_Security_Policy_Cheat_Sheet.md
@@ -81,7 +81,7 @@ Document directives instruct the browser about the properties of the document to
 
 Navigation directives instruct the browser about the locations that the document can navigate to.
 
-- `navigate-to` restricts the URLs which a document can navigate to by any mean (not yet supported by modern browsers in Jan 2021).
+- `navigate-to` restricts the URLs which a document can navigate to by any mean ([not yet supported](https://caniuse.com/?search=navigate-to) by modern browsers in Jan 2021).
 - `form-action` restricts the URLs which the forms can submit to.
 - `frame-ancestors` restricts the URLs that can embed the requested resource inside of  `<frame>`, `<iframe>`, `<object>`, `<embed>`, or `<applet>` elements.
     - If this directive is specified in a `<meta>` tag, the directive is ignored.
--- a/cheatsheets/Content_Security_Policy_Cheat_Sheet.md
+++ b/cheatsheets/Content_Security_Policy_Cheat_Sheet.md
@@ -29,8 +29,10 @@ CSP can be delivered to the user agent in different techniques.
 The following are headers for CSP.
 
 - `Content-Security-Policy` : W3C Spec standard header. Supported by Firefox 23+, Chrome 25+ and Opera 19+
-- `Content-Security-Policy-Report-Only` : W3C Spec standard header. Supported by Firefox 23+, Chrome 25+ and Opera 19+, whereby the policy is non-blocking ("fail open") and a report is sent to the URL designated by the `report-uri` directive. This is often used as a precursor to utilizing CSP in blocking mode ("fail closed")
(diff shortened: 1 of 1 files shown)


cheatsheets/Content_Security_Policy_Cheat_Sheet.md
```diff
@@ -79,7 +81,7 @@ Document directives instruct the browser about the properties of the document to
 
 Navigation directives instruct the browser about the locations that the document can navigate to.
 
-- `navigate-to` restricts the URLs which a document can navigate to by any mean.
+- `navigate-to` restricts the URLs which a document can navigate to by any mean (not yet supported by modern browsers in Jan 2021).
```
jmanico, line 84: Can you link this to https://caniuse.com/?search=navigate-to as a reference?

cheatsheets/Content_Security_Policy_Cheat_Sheet.md
```diff
@@ -191,7 +193,7 @@ object-src 'none'; base-uri 'none';
 
 ### Refactoring inline code
 
-By default CSP disables any unsigned JavaScript code placed inline in the HTML source, such as this:
+When `default-src` or `script-src*` directives are active, CSP by default disables any unsigned JavaScript code placed inline in the HTML source, such as this:
```
jmanico, line 196: unsigned? I do not believe digital signatures are a part of CSP

jmanico (changes requested):
Can you just link up caniuse for navigate-to and clarify the use of the term unsigned?

cheatsheets/Content_Security_Policy_Cheat_Sheet.md
jmanico: Suggest changing this to:

When `default-src` or `script-src*` directives are active, CSP by default disables any JavaScript code placed inline in the HTML source, such as this:

And can you also please remove the other use of the term signed in this document?

shaialon: I believe this references two main ways of "signing" code with CSP:
1. `nonce` 
2. SHA hashes (as describes right after.

We could change to 
```suggestion
When `default-src` or `script-src*` directives are active, CSP by default disables any inline JavaScript code, specified as script-elements or script-attributes, such as this:
```

jmanico: signed in the wrong term. Here is the exact text from the CSP 2 standard to consider.

The script-src directive lets developers specify exactly which script elements on a page were intentionally included for execution. Ideally, developers would avoid inline script entirely and whitelist scripts by URL. However, in some cases, removing inline scripts can be difficult or impossible. For those cases, developers can whitelist scripts using a randomly generated nonce.

and

The script-src directive lets developers whitelist a particular inline script by specifying its hash as an allowed source of script.

jmanico: And here is similar text from the CSP 3 standard to consider

The script-src directive governs four things:

Script requests MUST pass through §4.1.3 Should request be blocked by Content Security Policy?.

Script responses MUST pass through §4.1.4 Should response to request be blocked by Content Security Policy?.

Inline script blocks MUST pass through §4.2.3 Should element’s inline type behavior be blocked by Content Security Policy?. Their behavior will be blocked unless every policy allows inline script, either implicitly by not specifying a script-src (or default-src) directive, or explicitly, by whitelisting "unsafe-inline", a nonce-source or a hash-source that matches the inline block.

The following JavaScript execution sinks are gated on the "unsafe-eval" source expression:

eval()

Function()

setTimeout() with an initial argument which is not callable.

setInterval() with an initial argument which is not callable.

jmanico: Sincerely, there are zero mentions of digial signatures or similar in the CSP standard. =)

mackowski (approved):
Looks good! Thank you!

shaialon:
@jmanico , are we good to merge?
(it seems to be stuck pending your approval...)
Thanks.
image

jmanico (commented):
Great work!

Issue #517, opened 2020-11-09
Refresh XSS Prevention Cheat Sheet 


mackowski:
It is clear that this CS needs refreshing after ~10 years. That does NOT mean that this CS is wrong or give bad advice - quite the opposite, this is very conservative and if you implement it you should be free from XSSes. But this CS in the current form does not make it simple because it does not leverage modern frameworks and browser features. Keep the spirit of the CS and I will leave current recommendations (including escaping forward slashes) in the “conservative” section because I know that there are applications that needs to implement it in this fashion.

Main points of the new XSS CS:  
* Use safe contextual-autoescaping templating libraries
   * If above is not possible implement escaping in contact of HTML, HTML Attributes, JavaScript, HTML URL Parameter, CSS … preferably using library designed to to the job (e.g. OWASP Java Encoder)
 * Implement CSP (nonce-based preferably)
 * Implement Trusted Types 
   * If this is not possible other DOM-based prevention rules
In general we have all the rules currently here but I want to enforce using modern, well-designed and tested solutions over custom made implementation. 

During refactoring address 3 other open issues for this CS: 
https://github.com/OWASP/CheatSheetSeries/issues/133
https://github.com/OWASP/CheatSheetSeries/issues/141 
https://github.com/OWASP/CheatSheetSeries/issues/95

See also comments from https://github.com/OWASP/CheatSheetSeries/issues/515 and https://github.com/OWASP/CheatSheetSeries/pull/516

chris-morgan:
This has *reintroduced* the erroneous escaping of slash, which was removed in d7a6823db92af82da3f9538b23b03851d7556d78.

kwwall:
@mackowski - I know this particular issue is closed and I may be biased, but I'd really like to see the XSS Prevention Cheat Sheet include something about the anti-patterns that the ESAPI team (and probably others) have see pop up time and time again. I run into these constantly, because despite AppSec people constantly pointing devs to the XSS Prevention Cheat Sheet, this practice continues.

II have documented this in our local ESAPI GitHub wiki page [XSS Defense: No Silver Bullets](https://github.com/ESAPI/esapi-java-legacy/wiki/XSS-Defense:-No-Silver-Bullets).

[...]

PR #554, opened 2021-03-10
Updates XSS libraries from Microsoft

cosmo0:
The provided XSS library is hosted on CodePlex, which will disappear in july: https://archive.codeplex.com/?p=wpl (plus the library has been sunset for several years now)

Edited the part to add more up-to-date libraries.


cosmo0:
References issue #517

PR #597, opened 2021-03-18
Update advice about hashing long passwords with PBKDF2

NeilMadden:
As request by [Jim on Twitter](https://twitter.com/manicode/status/1372254081876656135) this PR updates the guidance around hashing long passwords:

- Removed the generic warning about long passwords from the bcrypt section. This didn't sit well in that section (bcrypt already enforces a limit), and all of the recommended KDFs already effectively "pre-hash" the input, so there is negligible cost for a longer password.
- Moved the warning to the PBKDF2 section to specifically warn about naive implementations that just iterate a hmac(password, data) function and so perform that pre-hashing step on every iteration.


NeilMadden:
To double-check my theory about the Django issue I tracked down https://github.com/django/django/commit/68540fe4df44492571bc610a0a043d3d02b3d320 which was fixed a week or so after the security disclosure. Prior to that fix they did what I suspected: only pre-hash long passwords inside the `hmac` implementation (which is called thousands of times). The fix does the smart thing and moves that out to the `pbkdf2` implementation so that it is done once. Same result, much faster. So that confirms that the Django issue was a case of sub-optimal implementation ([accidentally quadratic](https://accidentallyquadratic.tumblr.com) - look at me linking to Tumblr in 2021 😆 ).

jmanico (approved):
We are very grateful for your controbution!

PR #614, opened 2021-03-24
Not installing security updates is bad advice

itamarst:
This is obviously bad advice: security updates are important, and should be installed.

For more detailed explanation of why installing security updates in Docker builds is a good idea (I still can't believe I have to write this sentence, or had to write this article...), see https://pythonspeed.com/articles/security-updates-in-docker/


jmanico (approved):

jmanico:
Good call, what a scary mistake on our end, my apologies

itamarst:
Thank you!

chadlwilson:
Thanks for this @itamarst  - couldn't agree more. Technically the line `Ensure the OS packages versions are pinned` is potentially contradictory too; depending on how you construct your `Dockerfile` and order of install vs upgrade.

A number of (all?) distros only allow pinning to a specific build rather than allowing `~ major.minor` expressions; which would have the same effect of preventing patches for security to come through, to my knowledge?

PR #676, opened 2021-06-19
Document build process / add requirements / better bash

Naramsim:
- This PR adds a `Makefile` so as to have an entry point to understand how to locally generate the website. I had a hard time understanding how to generate it and finally ended up looking in the Github Actions files. A Makefile will make it easier for new users to start contributing.
- Add a `requirement.txt` file. So we know which libs we need to install.
- Use some best practices for your Bash script that generates the docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+.PHONY: help
+.SILENT: 
+
+help:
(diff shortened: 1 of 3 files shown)


mackowski (approved):
Nice!

PR #702, opened 2021-07-03
Fix Github Action

Naramsim:
https://github.com/OWASP/CheatSheetSeries/actions/workflows/publishing-check.yml right now is broken.

This PR should fix it, change its name, run it on every branch and every PR.

---

Please squash


.github/workflows/publishing-check.yml
```diff
@@ -1,34 +1,24 @@
...
+on: [push, pull_request]
 
 jobs:
   build:
-    name: Publishing Check
+    name: Build website
     runs-on: ubuntu-latest
     env:
       CI: true
+      WORKFLOW_GOOGLE_ANALYTICS_KEY: 'none'
```
Naramsim, line 11: The issue was the lack of the env var.

This solution isn't elegant at all, though. The best would be to set a default in the `mkdocs.yml`

https://github.com/OWASP/CheatSheetSeries/blob/8d4a5517cccf7d5798d50d6bf1c0cab0b27ea98c/mkdocs.yml#L16

But I don't know how to do it with that strange syntax.

Naramsim: It was easy

mackowski:
Awesome!

PR #690, opened 2021-06-25
Various Authorization and Authentication corrections

chrisdlangton:
addresses #686

PR #712, opened 2021-07-21
REST_Security_Cheat_Sheet typos 

chrisdlangton:
from PR #690 

CS is en_US from what I can tell, i'm personally en_AU so these are correct to me natively but I am fixing them to be consistent with the rest (pun?) of CS


mackowski:
I feel that changes in this PR are North America English vs. British/Non-North America English. 
I am not native speaker, but if I am right, I would leave it as it (without changes).
@jmanico do you have an opinion?

mackowski (approved):

jmanico:
Good merge :)

PR #714, opened 2021-07-21
adding some clarity for authn in Session Management CS

chrisdlangton:
from PR #690


cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -246,7 +246,7 @@ Session IDs must be considered untrusted, as any other user input processed by t
...
+The session ID must be renewed or regenerated by the web application after any privilege level change within the associated user session. The most common scenario where the session ID regeneration is mandatory is during the authentication process, as the privilege level of the user changes from the unauthenticated (or anonymous) state to the unauthorized state. Other common scenarios must also be considered, such as password changes, permission changes or switching from a regular user role to an administrator role within the web application. For all these web application critical pages, previous session IDs have to be ignored, a new session ID must be assigned to every new request received for the critical resource, and the old or previous session ID must be destroyed.
```
mackowski, line 249: I do not like this change; it is harder to understand and it was not wrong, the state changed to authenticated.

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -98,7 +98,7 @@ A web application should make use of cookies for session ID exchange management.
 
 ### Transport Layer Security
 
-In order to protect the session ID exchange from active eavesdropping and passive disclosure in the network traffic, it is essential to use an encrypted HTTPS (TLS) connection for the entire web session, not only for the authentication process where the user credentials are exchanged.
+In order to protect the session ID exchange from active eavesdropping and passive disclosure in the network traffic, it is essential to use an encrypted HTTPS (TLS) connection for the entire web session, not only for the authentication process where the user credentials are exchanged. This can be assured with HTTP Strict Transport Security (HSTS).
```
mackowski, line 101: This is not entirely true because HSTS is a browser header. If you are using client that is not supporting HSTS adding this harder will not add any value. Moreover, the word 'this' can be misinterpreted. 

I suggest to either drop this change or make clear here what HSTS header is doing etc.

mackowski: Ok actually I think that we should just link to the 
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Strict_Transport_Security_Cheat_Sheet.html

cheatsheets/Session_Management_Cheat_Sheet.md
```diff
@@ -334,7 +334,7 @@ Web applications can use JavaScript code to capture all the web browser tab or w
 
 ### Disable Web Browser Cross-Tab Sessions
 
-Web applications can use JavaScript code once the user has logged in and a session has been established to force the user to re-authenticate if a new web browser tab or window is opened against the same web application. The web application does not want to allow multiple web browser tabs or windows to share the same session. Therefore, the application tries to force the web browser to not share the same session ID simultaneously between them.
+Web applications can use JavaScript code once the user has logged in and a session has been established to force the user to authenticate again intentionally if a new web browser tab or window is opened against the same web application. The web application does not want to allow multiple web browser tabs or windows to share the same session. Therefore, the application tries to force the web browser to not share the same session ID simultaneously between them.
```
mackowski, line 337: what is wrong with re-authenticate?

mackowski (changes requested):

cheatsheets/Session_Management_Cheat_Sheet.md
chrisdlangton: Agree on both points

cheatsheets/Session_Management_Cheat_Sheet.md
chrisdlangton: @mackowski 
Fair. however there is an argument that in context there is a distinction but generally speaking you are correct, they may seem the same.

That context is the heading `Renew the Session ID After Any Privilege Level Change` which essentially speaks to an Authorization mechanism. Therefore it is distinct from merely being authentic, to gain elevated access one must be authorised to that access level.

cheatsheets/Session_Management_Cheat_Sheet.md
chrisdlangton: Well what exactly is 're-authenticate'? 

Are when you 're-authenticate' do you mean you are just 'authenticating' again now because the previous _perceived active_ authenticated 'session' has somehow expired or otherwise entered an untrusted state?

The paragraph talks to the point that, quote:

> The web application does not want to allow multiple web browser tabs or windows to share the same session.

Meaning the new tab was never authenticated to be 're' authenticated. 

In any context, re-authentication is either just authenticate or it can be confused with a 2nd factor authentication, or more problematic it an be used in place of authorization and sound _correct_, but we know better. Let's avoid a semantic reference to re-authentication and just use the proper terms 'authenticate', 'authorize', and 2fa where one might be using re-authentication.

cheatsheets/Session_Management_Cheat_Sheet.md
mackowski: Apologies that I did not respond for over a week on that...
You are right but it is still harder to understand by a regular IT person that does not work in security.
What about: state changed from the unauthenticated (or anonymous) to the authenticated, unauthorized state.

chrisdlangton: No need to apologise, collaboration is like eventual consistency :+1: 
Very true, and good compromise. I feel silly for not realising it myself :facepalm:

mackowski (approved):
Looks good, thanks!

mackowski:
@chrisdlangton awesome contribution!

PR #715, opened 2021-07-21
Transport Layer Protection CS en_US

chrisdlangton:
from PR #690 

CS is en_US from what I can tell, i'm personally en_AU so these are correct to me natively but I am fixing them to be consistent with the rest of CS


mackowski:
Good reasoning what to change to en_US in this CS

mackowski (approved):

PR #716, opened 2021-07-21
Web Service Security CS auth updates

chrisdlangton:
from PR #690


mackowski (approved):
Awesome update :) Thank you!

PR #755, opened 2021-10-10
XSS Prevention 2021

colecornford:
Hello Friends, Cole from Slack.

I have rewritten the XSS Prevention CheatSheet. I think there is significant value in the old XSS Prevention CS, but I think the content is not digestible for most developers, dated, and very long.

Today I wrote this from scratch to address those points. I believe that adopting framework security mechanisms, getting your content-security policy up to date, and then looking at some other mitigations like SRI, HTTPOnly, and choosing the right technology options are modern approaches.

Things that need fixing:

- I didn't want to plagiarise the content from the Trusted-Types or CSP RFC's, so provided a link and moved on.
- I want to incorporate the OE advice, but rewrite it for 2021 and cut 80% of the content so that it can be more consumable.
- I didn't include links or references.

Thank you, hope this addresses #517 and next weekend I will work on whatever the next CS on the agenda is.

Cole Cornford

This PR covers issue #517

### DIFF
--- a/cheatsheets_draft/Cross_Site_Scripting_Prevention_Cheat_Sheet_2021.md
+++ b/cheatsheets_draft/Cross_Site_Scripting_Prevention_Cheat_Sheet_2021.md
@@ -80,7 +80,7 @@ Refactor inline code. Moving inline code into hosted files will prevent allow yo
 - Nonces
 - Trusted Types
 
-A fully mature CSP uses a combination of nonces, strict-dynamic, and trusted types over domain and path allowlisting of content. The (w3 explanation)[https://www.w3.org/TR/CSP3/#strict-dynamic-usage] of Strict-Dynamic and nonce based validation is clear about the benefits and implementation.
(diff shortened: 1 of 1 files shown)


jmanico:
Even with React, you need to escape CSS contexts, validate URL’s that land in a for tags (JS urls) and there is dangerouslySetInnerHTML. Angular is better, Vue is not.

The point I’m making is, the auto-escaping in modern frameworks is not (nearly) enough to avoid XSS so I think some of your earlier prose is a little bit off.

What’s changed is, contextual output encoding is not needed manually in new frameworks, but now developers need to understand obscure ways to use and handle data in frameworks properly, you dont get XSS protection for free. You have to understand React, Vue, and Angular’s quirks and very few devs have this knowledge.

Just something to think about as we mature this CS. 

--
Jim Manico
@Manicode

jmanico:
You can use any content from standards so long as you quote and cite them properly.

jmanico:
Before I merge this and overwrite the previous work I just want to say that although I have small criticism I like the re-organization. We are about to over-write the most popular page in all of OWASP's history, so I want to do this right and scrutinize every sentence before we say go.

Would you like to move this to google docs and/or do a live session to go over this and debating a few points together?

Issue #746, opened 2021-09-23
New CS proposal: HTTP Headers Cheat Sheet


TheSmartScanner:
## Configuring Web application/servers by implementing proper HTTP headers
This cheat sheet will include all security related HTTP headers and their proper values which can have benefit for the security of websites.
This cheat sheet will include samples and guides for configuring HTTP headers in different technologies like PHP, Apache, etc.

## What security issues are commonly encountered related to this area?
- Cross-Site Scripting
- Click-jacking
- Packet sniffing
- Spectre
- information disclosure

## What is the objective of the Cheat Sheet?
A single source and guide for configuring and testing all security HTTP headers.

## What other resources exist in this area?
[OWASP Secure Headers Project](https://owasp.org/www-project-secure-headers/)
[Mozilla HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)
[OWASP Content Security Policy Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Content_Security_Policy_Cheat_Sheet.md)

jmanico:
I absolutely love this idea. Let's do it and I will actively help.

TheSmartScanner:
So, is it approved?

jmanico:
![download](https://user-images.githubusercontent.com/531465/136714892-0e07fdb9-7aba-4c07-800b-afe55b79c9b7.jpeg)

TheSmartScanner:
Thanks for the changes. It looks good to me.👍👌

PR #760, opened 2021-10-18
Add HTTP Headers Cheat Sheet

TheSmartScanner:
This PR covers issue #746


jmanico:
X-XSS-Protection in block mode causes harm if should be set to 0 and be fully disabled. Browsers are removing this header anyways.

PR #798, opened 2021-12-16
Update README slack invite link

kingthorin:
Use the foundation invite link as heroku sometimes experiences resource/usage limit issues late in the month.


mackowski:
Thanks @kingthorin !

PR #833, opened 2022-01-31
Detection and Response sections via @thatsjet

thatsjet:
Submitting sections previously numbered 9 for IR and &11 detections now in reverse order (detect first, then respond).


thatsjet:
I'm closing and submitting a new one based on #832

PR #852, opened 2022-02-08
[Certificate Pinning Cheat Sheet] General Improvements

cpholguera:
- enhance outline.
- highlight Benefits and Downsides in Certificate/Public Key/Hash sections.

I was about to add this information to the MSTG but thought that it might be better here and we'd simply reference back here since it applies to _any application_ (not only mobile). In the MSTG we'll then keep only the Android/iOS specifics extending what you have here.

- add steps for "What Should Be Pinned" including first step about deciding for root CA, intermediate CA or leaf certificate.


mackowski (approved):
Awesome, thanks @cpholguera

cpholguera:
Thanks @mackowski, I've just marked it "ready for review" so that it can be also merged. I have some more suggestions & improvements but it will be easier to do on a separate PR.

PR #903, opened 2022-05-03
Remove 'shame' to 'developers' language

LewisArdern:
The way this part is written its shaming developers, the language does not help at all. It's the fact libraries still allow for exposing raw SQL queries we shouldn't be shaming developers for that.


jmanico (approved):

jodavis1:
Merge it

jmanico:
I am so very glad you pointed this out - well done

PR #959, opened 2022-08-05
enabling Edit button

noraj:
It seems that specifying explicitly `edit_uri` in the configuration with a blank value `edit_uri: ""`, overrides the dynamic value computed from `repo_url` and so disables the _Edit button_.

cf. mkdocs et material theme doc

- https://www.mkdocs.org/user-guide/configuration/#edit_uri
- https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/?h=edit#edit-button

this would be nice to have the _Edit button_ on top of every page instead of having to open the github repository and look for the right page based on the URL of the page.

image

PS: unfortunately the github action CI https://github.com/OWASP/CheatSheetSeries/actions/runs/2805308503 just build the website base on the PR but doesn't deploy a staging preview

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -5,7 +5,6 @@ site_description: Website with the collection of all the cheat sheets of the pro
 # Repository
 repo_name: OWASP/CheatSheetSeries
 repo_url: https://github.com/OWASP/CheatSheetSeries
-edit_uri: ""
 
 # Copyright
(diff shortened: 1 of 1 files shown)


jmanico (approved):

noraj:
The edit button is not live despite this PR being merge and deployed.

Material doc says the edit button is automatically set and that `edit_uri: ""` must be used to hide it (which was just removed).

image

So theoretically the edit button should be displayed now. [requirements.txt](https://github.com/OWASP/CheatSheetSeries/blob/master/requirements.txt) doesn't show which version of `mkdocs-material` is used but since the feature was added in the first version 0.1.0 it should be present anyway.

When I build locally it appears, but the deployed version at https://cheatsheetseries.owasp.org/ doesn't

image

PR #1004, opened 2022-11-07
Link to current PostgreSQL Setup and Operation docs

jeremycook:
PostgreSQL provides evergreen documentation that follows the current stable version under https://www.postgresql.org/docs/current/


mackowski (approved):

mackowski:
Thanks @jeremycook!

Issue #901, opened 2022-04-29
Update: Cryptographic Storage Cheat Sheet


vcsjones:
## What is missing or needs to be updated?

The "Cryptographic Storage Cheat Sheet" has a couple of things I would like to help improve.

1. The "Cryptographically Secure Functions" for random number generation lists `RNGCryptoServiceProvider` for C# / .NET. This API has been [obsolete](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rngcryptoserviceprovider?view=net-6.0) in newer versions of .NET. The replacement is just [`RandomNumberGenerator`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator?view=net-6.0).

2. The mention of using AES XTS mode is somewhat out of place. XTS is _extremely_ specialized and those that do need it likely already know they are in such a specialized scenario. I would perhaps recommend removing the mention of XTS entirely as not to cause people to believe it is applicable to them. I don't feel super strongly about this - however if I were looking at a system which uses AES-XTS I would be very skeptical going in.

3. I think it would be worthwhile to discuss nonce generation and exhaustion with various modes. While AES-GCM is a preferable cipher, reusing a nonce (IV) is catastrophic. For this reason, NIST specifically recommends rotating the key after it has been used 2<sup>32</sup> times. (See NIST SP800-38D section 8.3.). CTR mode also fails catastrophically with IV reuse.

## How should this be resolved?

Using the suggestions above.

1. Mention "both", I suppose, so that those on older versions of .NET still know to use `RNGCryptoServiceProvider`.
2. Consider dropping the paragraph mentioning XTS mode.3. 
3. Create a section discussing each modes expectations around nonces, and the consequence of duplicate nonces.

mackowski:
Awesome! Thank you @vcsjones! Do you want to create PR with the updates?

vcsjones:
@mackowski Yep. Will probably be in the next few days.

szh:
@vcsjones are you still planning to work on this? Otherwise I may be able to take it.

mackowski:
@szh I think that we can assume that you can work on this :)

vcsjones:
@szh yep, please do! Sorry I didn't get to this soon :-(

PR #1012, opened 2022-11-23
Update .NET Random function

szh:
This PR covers two thirds of issue #901

- Replace `RNGCryptoServiceProvider` with `RandomNumberGenerator` in the .NET section, because `RNGCryptoServiceProvider` is obsolete. The original issue (#901) suggested keeping both but since this is a cheat sheet I think for brevity we should jettison the old one. I'm open to input on this.

- Remove XTS section because it's too specialized to be useful in a cheat sheet.

I did not address the third point - discussing nonce generation and exhaustion, because I don't have expertise there. It may be a good idea to open a new issue to make tracking that easier.


mackowski (approved):

mackowski:
Thank you @szh for your next contribution!

PR #1020, opened 2022-12-17
Update: Microservices Security Cheat Sheet

### LINKED ISSUE #981: Update: Microservices Security Cheat Sheet

This [line](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Microservices_security.md?plain=1#L75) states

> One of the simplest way to propagate external entity identity is to re-use the access token received by the edge and pass it to internal microservices. It should be mentioned that approach is highly insecure due to possible external access token leakage and may decrease an attack surface because the communication relies on proprietary token-based system implementation and internal microservices have to understand external access token.

I think, this issue results in an **increase** in attack surface rather than a decrease.

Also, from reading this paragraph, I'm not quite sure how this approach makes the Software more / less secure. One scenario that I thought of is: If somehow an internal microservice is unintentionally exposed to the outside world, then that microservice can be called directly using the access token. 

Whereas if the internal microservice accepts a token format known only to internal microservices, then the above scenario can't occur.

Sorry if my question above is naive, thank you! :)

Zhiyuan-Amos:
Fixes #981.


cheatsheets/Microservices_security.md
```diff
@@ -72,7 +72,7 @@ Netfix presented ([link](https://www.youtube.com/watch?v=R6tUNpRpdnY), [link](ht
...
...vice level microservice has to understand caller context (e.g. user ID, user roles/groups). In order to allow internal service layer to enforce authorization edge layer has to propagate authenticated external entity identity (e.g., end user context) along with a request to downstream microservices. One of the simplest way to propagate external entity identity is to re-use the access token received by the edge and pass it to internal microservices. It should be mentioned that approach is highly insecure due to possible external access token leakage and may increase an attack surface because the communication relies on proprietary token-based system implementation: If an internal service is unintentionally exposed to the external network, then it can be directly accessed using the leaked access token. This attack is not possible if the internal service only accepts a token format known only to internal services. This pattern also is not external access token agnostic, i.e. internal services have to understand external access token and support a wide range of authentication techniques to extract identity from different types of external tokens (e.g. JWT, cookie, OpenID Connect token).
```
szh, line 75:
```suggestion
To make fine-granted authorization decision at the microservice level, microservice has to understand caller context (e.g. user ID, user roles/groups). In order to allow internal service layer to enforce authorization, edge layer has to propagate authenticated external entity identity (e.g., end user context) along with a request to downstream microservices. One of the simplest way to propagate external entity identity is to re-use the access token received by the edge and pass it to internal microservices. It should be mentioned that approach is highly insecure due to possible external access token leakage and may increase an attack surface because the communication relies on proprietary token-based system implementation: If an internal service is unintentionally exposed to the external network, then it can be directly accessed using the leaked access token. This attack is not possible if the internal service only accepts a token format known only to internal services. This pattern also is not external access token agnostic, i.e. internal services have to understand external access token and support a wide range of authentication techniques to extract identity from different types of external tokens (e.g. JWT, cookie, OpenID Connect token).
```

szh (changes requested):
💯 
I just suggested a couple commas to make this easier to read.

cheatsheets/Microservices_security.md
Zhiyuan-Amos: @szh thank you for the suggested changes! 😄

mackowski:
Thanks @szh amd @Zhiyuan-Amos

mackowski (approved):

PR #1026, opened 2022-12-20
Update GH Actions versions; Pin Ubuntu version

szh:
This is an attempt to fix the Python setup errors that were occurring in the GitHub actions due to a Python version not being available in a new version of Ubuntu. This PR fixes that and additionally updates the versions of all the GitHub actions and pins the latest available Ubuntu version (22.04).


mackowski (approved):
looks reasonable, let see if this works

Issue #1043, opened 2023-01-02
Update: Password Storage Cheatsheet


oddcb:
## What is missing or needs to be updated?
PBKDF2 iterations.

Looking into the past commits to the cheat sheet it seems the current ones were sourced from: 

https://twitter.com/sc00bzt/status/1322164349738262528

These recommendations were updated again this December: 

> Parallel PBKDF2
> PPBKDF2-SHA512: cost 2 (Based on RTX 4090 as "1.5 GPUs")
> PPBKDF2-SHA256: cost 5 (Based on RTX 4090 as "1.5 GPUs")
> PPBKDF2-SHA1: cost 10 (Based on RTX 4090 as "1.5 GPUs")
> 
> PBKDF2 (worst)
> PBKDF2-HMAC-SHA512: 210,000 iterations (Based on RTX 4090 as "1.5 GPUs")
> PBKDF2-HMAC-SHA256: 600,000 iterations (Based on RTX 4090 as "1.5 GPUs")
> PBKDF2-HMAC-SHA1: 1,300,000 iterations (Based on RTX 4090 as "1.5 GPUs")
> 

https://infosec.exchange/@sc00bz/109599362314030488

See also: https://tobtu.com/minimum-password-settings/

## How should this be resolved?

Update cheat sheet with new values. I'd also add a reference to source for anyone wondering where these numbers come from.

jmanico:
This is super useful. Would you care to PR? If not I can.

Moksh45:
Could you please provide me with additional information regarding this issue?

kwwall:
@Moksh45 Specifically what additional information met regarding this issue are you seeking? If you are referring to specific comments made on these 2 PRs, then please reference a link to the specific comment and when asking about it, please also @-mention the individual who made the comment. Otherwise, we have to look for it and try to guess what you are specifically asking about. Thanks!

Moksh45:
@kwwall The issue I had regarding these two pull requests has been resolved. 

Thank You.

mackowski:
I see that this is already addressed. I am closing the issue

PR #1055, opened 2023-01-11
Update PBKDF2 work factors according to RTX4000 #1043

oddcb:
Thank you for submitting a Pull Request (PR) to the Cheat Sheet Series. 

> :triangular_flag_on_post: If your PR is related to grammar/typo mistakes, please double-check the file for other mistakes in order to fix all the issues in the current cheat sheet.

Please make sure that for your contribution:

- [x] All the markdown files do not raise any validation policy violation, see the [policy](https://github.com/OWASP/CheatSheetSeries/actions?query=workflow%3A%22Markdown+Link+Check%22).
- [x] All the markdown files follow these [format rules](https://github.com/OWASP/CheatSheetSeries/blob/master/CONTRIBUTING.md#markdown).
- [x] The CI build of your PR pass, see the build status [here](https://github.com/OWASP/CheatSheetSeries/actions).

If your PR is related to an issue, please finish your PR text with the following line:

This PR covers issue #1043 

Thank you again for your contribution :smiley:

PR #1057, opened 2023-01-13
Updated the Password Storage Cheatsheet

suyash5053:
This changes resloves #1043 by updating the Password Storage Cheatsheet.
Signed-off-by: Suyash Srivastava <suyashsrivastava5053@gmail.com>


kwwall (changes requested):
+1 for leaving a reference, but this doesn't explain all of my concerns, albeit for a different PR. Rather than repeating the here, please see the comments I left at:
https://github.com/OWASP/CheatSheetSeries/pull/1055/files/79b0198fefd61097b86c6475190d3d75d022b861#r1067644067

That was for PR #1055 but the same changes were made there.

Given that fact that NIST specifies at different figures for privileged and non-privileged accounts, I think that should be noted here. Which is this for? Also, since these iteration counts are different from the various NIST documents, I think we need the significance of that explained. For example, some industry sectors in the USA (e.g., typically financial institutions and companies that do contract work for the US government) generally need to adhere to FIPS-140 compliance. See the relevant comments at https://github.com/OWASP/CheatSheetSeries/pull/1055/files#r1067641302 for that context.

jmanico (approved):

jmanico:
Can you check this against the lint errors and the previous update to the CS?

Issue #962, opened 2022-08-13
Update: OS Command Injection Defense


h4r1337:
## What is missing or needs to be updated?
In this cheat sheet under the [PHP](https://cheatsheetseries.owasp.org/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.html#php) section, it recommends to use `escapeshellarg()` or `escapeshellcmd()` to escape malicious user inputs. The PHP documentation of these functions says that: [`escapeshellarg()`](https://www.php.net/manual/en/function.escapeshellarg.php) adds single quotes around the given string and [`escapeshellcmd()`](https://www.php.net/manual/en/function.escapeshellcmd.php) escapes any characters in a string that might be used to trick a shell command into executing arbitrary commands using a `\`. 
Even though these functions help to avoid executing system level commands, it is still vulnerable to argument injection. 
This is a great research on PHP supply chain attack affecting composer back-end, which exploits argument injection: https://blog.sonarsource.com/php-supply-chain-attack-on-composer/   

## How should this be resolved?

I think developers should know about these kind of attacks. And it would be much better if we could add any notes on this attack vector and how to fix it.
According to this [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) this can be fixed by just adding `--` to the command which will then prevent the evaluation of additional arguments to the command:
```php
system("curl -- ".escapeshellarg($arg);
```
Do we also need to specify how these 2 functions work?
I think a small note will be helpful.

Spikebady:
This video was taken on a mobile phone

mackowski:
Uhh apologies @h4r1337 that I am responding after so long.... I do not know why I missed this issue. You are right! Please create PR with that!

h4r1337:
Hey @mackowski, no problem... I will create a PR by tomorrow.

mackowski:
hey @h4r1337 any updates on this?

PR #1084, opened 2023-02-03
Add note on Argument Injection

h4r1337:
- Added a new section about argument injection under `Introduction`
- Removed `escapeshellcmd()` from `Defense option 2` because it does not prevent argument injection and explained the working of `escapeshellarg()` and how it responds to an argument.
- Added proper defense mechanism for Argument Injection under `Layer 2` of `Defense option 3`
- Added `-` character inside `Note A` to prevent an argument

This PR covers issue #962

### DIFF
--- a/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
+++ b/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
@@ -64,7 +64,7 @@ If there are available libraries or APIs for the language you use, this is the p
 
 For examples, see [escapeshellarg()](https://www.php.net/manual/en/function.escapeshellarg.php) in PHP.
 
-The `escapeshellarg()` sorrounds the user input in single quotes, so if the malformed user input is something like `& echo "hello"`, the final output will be like `calc '& echo "hello"'` which will be parsed as a single argument to the command `calc`.
+The `escapeshellarg()` surrounds the user input in single quotes, so if the malformed user input is something like `& echo "hello"`, the final output will be like `calc '& echo "hello"'` which will be parsed as a single argument to the command `calc`.
 
 Even though `escapeshellarg()` prevents OS Command Injection, an attacker can still pass a single argument to the command.
 
--- a/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
+++ b/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
@@ -89,7 +89,7 @@ If calling a system command that incorporates user-supplied cannot be avoided, t
 **Note A:**
 
 ```text
-& |  ; $ > < ` \ ! -
+& |  ; $ > < ` \ ! ' " ( )
 ```
 
 ## Additional Defenses
--- a/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
+++ b/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
@@ -84,7 +84,7 @@ If calling a system command that incorporates user-supplied cannot be avoided, t
 - In regards to the **arguments** used for these commands, they should be validated using the following options:
     - **Positive or allow list input validation**: Where are the arguments allowed explicitly defined.
     - **Allow list Regular Expression**: Where a list of good, allowed characters and the maximum length of the string are defined. Ensure that metacharacters like ones specified in `Note A` and white-spaces are not part of the Regular Expression. For example, the following regular expression only allows lowercase letters and numbers and does not contain metacharacters. The length is also being limited to 3-10 characters: `^[a-z0-9]{3,10}$`
-- According to **Guidline 10** of this [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html), *The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.* For example, `curl -- $url` will prevent an argument injection even if the `$url` is malformed and contains an additional argument.
+- According to **Guideline 10** of this [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html), *The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.* For example, `curl -- $url` will prevent an argument injection even if the `$url` is malformed and contains an additional argument.
 
 **Note A:**
 
--- a/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
+++ b/cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
@@ -26,6 +26,28 @@ Now, both the *Calculator* application and the value *test* are displayed:
 
 The problem is exacerbated if the compromised process does not follow the principle of least privileges and attacker-controlled commands end up running with special system privileges that increase the amount of damage.
 
+### Argument Injection
+
+Every OS Command Injection is also an Argument Injection. In this type of attacks, user input can be passed as arguments while executing a specific command.
+
+For example, if the user input is passed through an escape function to escape certain characters like `&`, `|`, `;`, etc.
+
+```php
+system("curl " . escape($url));
+```
+
+which will prevent an attacker to run other commands.
+
+However, if the attacker controlled string contains an additional argument of the `curl` command:
+
+```
+system("curl " . escape("--help"))
+```
+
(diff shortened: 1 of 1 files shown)


cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
```diff
@@ -40,7 +62,11 @@ If there are available libraries or APIs for the language you use, this is the p
 
 **TODO: To enhance.**
 
-For examples, see [escapeshellarg()](https://www.php.net/manual/en/function.escapeshellarg.php) or [escapeshellcmd()](https://www.php.net/manual/en/function.escapeshellcmd.php) in PHP.
+For examples, see [escapeshellarg()](https://www.php.net/manual/en/function.escapeshellarg.php) in PHP.
+
+The `escapeshellarg()` sorrounds the user input in single quotes, so if the malformed user input is something like `& echo "hello"`, the final output will be like `calc '& echo "hello"'` which will be parsed as a single argument to the command `calc`.
```
kwwall, line 67: Fix typo: surrounds

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
```diff
@@ -40,7 +62,11 @@ If there are available libraries or APIs for the language you use, this is the p
 
 **TODO: To enhance.**
```
kwwall, line 63: Can / should we remote this "TODO" now that more content is present? I'm not sure how much more was expected here.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
```diff
@@ -58,11 +84,12 @@ If calling a system command that incorporates user-supplied cannot be avoided, t
 - In regards to the **arguments** used for these commands, they should be validated using the following options:
     - **Positive or allow list input validation**: Where are the arguments allowed explicitly defined.
     - **Allow list Regular Expression**: Where a list of good, allowed characters and the maximum length of the string are defined. Ensure that metacharacters like ones specified in `Note A` and white-spaces are not part of the Regular Expression. For example, the following regular expression only allows lowercase letters and numbers and does not contain metacharacters. The length is also being limited to 3-10 characters: `^[a-z0-9]{3,10}$`
+- According to **Guidline 10** of this [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html), *The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.* For example, `curl -- $url` will prevent an argument injection even if the `$url` is malformed and contains an additional argument.
```
kwwall, line 87: Do keep in mind that it says _should_, **not** _must_. Also, even when intended, it often isn't implemented correctly.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
```diff
@@ -58,11 +84,12 @@ If calling a system command that incorporates user-supplied cannot be avoided, t
...
 
 **Note A:**
 
 ```text
-& |  ; $ > < ` \ !
+& |  ; $ > < ` \ ! -
```
kwwall, line 92: Somewhere, we should note that the set of shell meta characters will depend on the specific shell that is being used. It should be obvious, but there are many who've only ever used a single command line shell.

kwwall (changes requested):
The typo, at least, ought to be fixed. YMMV on the others, so I'd wait for other reviewer's comments.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
h4r1337: Yeah. like for example in `fish` there's no need for a `$(command)` just `(command)` work as command substitution. And `(` `)` these characters are not added in the list.

jmanico (dismissed):

szh (commented):
There's one remaining lint error

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
mackowski: @h4r1337 can you look into that?

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
mackowski: Agree on that. What do you think @h4r1337 should we change _should_ to _must_ here?

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
mackowski: +1

kwwall: @h4r1337 -  A few things things... 
1. If we wish to ensure that  the characters are truly excluded from any regex allow list as this earlier statement implies:
> Ensure that metacharacters like ones specified in `Note A` and white-spaces are not part of the Regular Expression.

Then  we should be tell them that they need to be careful with ranges in character classes that might accidentally include ranges. since the grave accent appears between uppercase and lowercase alphabetic characters in ASCII so some regular expression libraries could interpret something like '[A-z0-9]' as including the grave accent.

2. Why are we including '-' here at all? I'm not getting that one. I can introduce command flags, but that doesn't seem to be sufficient reason to me so I must be missing something. In ESAPI's `Encoder.encodeForOS()`, we consider '-' "immune from encoding" just as we do for alphanumeric characters.
3. I do think both `(` and `)` should be added to the list, not just because of fish, but for other shells such as bash, unmatched unquoted parens can be used as to cause syntax errors which could have unintended consequences or even in some cases be used as a DoS-like attack.
4. Whatever character is used to quote shell-metacharacters (typically '\') should be considered here as well.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
h4r1337: @mackowski Actually that exact sentence was copied from the Guidline 10 of the POSIX page. And as @kwwall said it often isn't implemented correctly. And if it gives more clarity then I think it is better to change that to `must` instead of `should`.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
h4r1337: @kwwall I think adding `-` was a mistake from my side. Since we are adding `--` there's no need to include `-` in the list and this decision to add `-` in the list may not be appropriate in all contexts. So I will be removing it from the list.

Also adding single and double quotes and parantheses to the list as well.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
h4r1337: Well, I'm not so sure how much more was expected there. I removed `escapeshellcmd()` becasue generally `escapeshellarg()` is considered more secure than `escapeshellcmd()` because it is designed to escape individual command line arguments rather than entire command lines.
I also described how the function behaves.
if the idea that we can utilise language-specific functions to escape values supplied to OS commands is what this section is supposed to convey, then it is done there with a PHP function as an example.

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
kwwall: Should that be 'Guideline 10' instead of 'Guidline 10' or does POSIX just not know how to spell.

h4r1337: No its my bad :sweat_smile: Fixing it...

cheatsheets/OS_Command_Injection_Defense_Cheat_Sheet.md
jmanico: This looks solid to me, solid enough. I'm pushing it live, please stop me or ask me to reverse it @kwwall and others if I am in error.

jmanico (approved):

kwwall:
@jmanico - Good enough for me. It's at least more thorough than it was. From a shell perspective, depending on how the command and the arguments get passed to the OS, even `escapeshellarg` can be bypassed, especially in the case where it the command ends up getting passed as something like
```bash
/bin/sh -c "commandName arg1 arg2 ..."
```
which is often the case. But let's not let the perfect become the enemy of the good here. If we find something wrong, we can always fix it with a subsequent PR.

PR #1096, opened 2023-02-22
Fix cs filename

szh:
This fixes the file name of the Microservices cheat sheet to conform to the same format as the other cheat sheets. This will fix it incorrectly displaying in the table of contents:

image


kwwall (approved):
LGTM.

mackowski (approved):

Issue #1094, opened 2023-02-22
Update: [XML_External_Entity_Prevention_Cheat_Sheet.md]


n-of-one:
## What is missing or needs to be updated?
The example setup for preventing XXE for jaxb is not protecting against entity expansion attacks (i.e. Billion Laughs). 

``` java
//Disable XXE
SAXParserFactory spf = SAXParserFactory.newInstance();
spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
spf.setXIncludeAware(false);

//Do unmarshall operation
Source xmlSource = new SAXSource(spf.newSAXParser().getXMLReader(),
                                new InputSource(new StringReader(xml)));
JAXBContext jc = JAXBContext.newInstance(Object.class);
Unmarshaller um = jc.createUnmarshaller();
um.unmarshal(xmlSource);
```

With this setup, an XML file like this:

```xml
<!DOCTYPE lolz [
<!ENTITY lol "lol">
<!ENTITY lol1 "&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;">]>
<person><name>&lol1;</name></person>
```

will be unmarshalled into:

```xml
<person><name>lollollollollollollollollollol</name></person>
```

In addition to this: the example setup also is not strictly better than not having any settup.

Using org.glassfish.jaxb:jaxb-runtime 4.0.2 for XML files with a DTD containing a general entity like "file: " will result in exception when trying to unmarshal them.

Using the OWASP recommendations will result in an object, but with the entity fields replaced with an empty string.

See: https://github.com/n-of-one/jaxb_xxe for a demo.

## How should this be resolved?
Using the `disallow-doctype-decl` completely protects against XXE. 

Recommended setup:
```java
//Disable XXE
SAXParserFactory spf = SAXParserFactory.newInstance();
spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);

//Do unmarshall operation
Source xmlSource = new SAXSource(spf.newSAXParser().getXMLReader(),
                                new InputSource(new StringReader(xml)));
JAXBContext jc = JAXBContext.newInstance(Object.class);
Unmarshaller um = jc.createUnmarshaller();
[...]

PR #1097, opened 2023-02-22
Improve JAXB example/recommendation with 'disallow-doctype-decl' feat…

n-of-one:
Improvement suggestion for protecting against XXE when using JAXB.

This PR covers issue #1094.

### DIFF
--- a/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
@@ -333,11 +333,7 @@ Since a `javax.xml.bind.Unmarshaller` parses XML and does not support any flags 
 ``` java
 //Disable XXE
 SAXParserFactory spf = SAXParserFactory.newInstance();
-spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
-spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
-spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
-spf.setXIncludeAware(false);
+spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
 
 //Do unmarshall operation
 Source xmlSource = new SAXSource(spf.newSAXParser().getXMLReader(),


kwwall (commented):
My concern is is, does this depend on the version that JAXB a client application is using?
If so, maybe is should say something along the lines of "If you are using a version of JAXB later than x.y, use this approach, otherwise use the second approach" and show them both.

I mean, there should be a fallback if someone is unable to use a version of JAXB that doesn't support `disallow-doctype-decl`.

n-of-one:
Yes, I guess that's why the recommended setup was created in the first place :) 

We stumbled upon this when testing with RedHat openJDK 8, which has jaxb bundled. When I created the demo project I started out with Oracle jdk 8 (if I remember correctly) and then switched to Java 17 and 19 to be more up to date. So far all those versions exhibit the same behavior. However, for bundled jaxb (java 8-?)  I'm not sure there is a well defined behavior.

I could do some more testing with various JDKs if that would be useful. The recommendation could also be a bit less concise and explain the situation a bit. Or it could even just add the `disallow-doctype-decl` to the current recommendation with a comment that the others are fallback in case the jaxb implementation does not support disallow-doctype-decl.

kwwall:
Hmm, okay. Since it could be hard to tell (at least w/out a lot of research, which versions of bundled or unbundled JAXB support `disallow-doctype-decl`, perhaps the best thing would be to include them all. Or maybe, write it something like this:
```java
// Disable XXE
SAXParserFactory spf = SAXParserFactory.newInstance();
try {
    // First try the newer approach ...
    spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
} catch( ParserConfigurationException | SAXNotRecognizedException | SAXNotSupportedException e) {
    // Try the older alternatives ...
    spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
    spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
    spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
    spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
    spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
    spf.setXIncludeAware(false);
}
```
Something like that should work and wouldn't require nearly as much research.

n-of-one:
You have a point, I'm not sure which JVMs I would need to check to have a confident answer. I had checked several openJDK 8 implementations already, but XML and the JVM is a bit of a mess, so it would never be an airtight recommendation.

On the other hand I dislike the idea of having a JAXB configuration where you don't really know what it's protecting you against. 

My proposal would be to change it into something like this:

```java
SAXParserFactory spf = SAXParserFactory.newInstance();

// This is the PRIMARY defense. If DTDs (doctypes) are disallowed, almost all
// XML entity attacks are prevented.
// If this works, the other setFeature lines are not needed.
spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);

// In case your SAXParserFactory throws an exception when setting the `disallow-doctype-decl` feature,
// remove that line and use the alternatives below. 
// Note that these don't necessarily protect against expansion attacks such as the Billion Laughs
spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
spf.setXIncludeAware(false);
```

This way, if the JVM+xml libraries  don't support `disallow-doctype-decl` at least you know that you are less protected.

jmanico (approved):
Are you sure removing all of this is safe?

PR #1102, opened 2023-03-14
Add pinning cautions

heymarcel:
There are situations where pinning can lead to an outage. This edit adds some cautionary points to help implementers identify those situations ahead of time.


jmanico (approved):

jmanico:
Well done, thank you!

PR #1117, opened 2023-04-13
Improved JAXB Example

fw10:
Hi there, 

I read through the last PR that changed the JAXB code snippet (https://github.com/OWASP/CheatSheetSeries/pull/1097) which undid some of my previous changes and don't agree with it 100%. I feel like this new PR here would improve the JAXB example, because:

1. Even if doctypes are completely disabled, XXE may be possible via `XIncludes`
2. I'm not sure if `FEATURE_SECURE_PROCESSING` would add a security benefit if `disallow-doctype-decl` is set to `true`. However, I don't see any harm. I went down the rabbit hole of `FEATURE_SECURE_PROCESSING` before, and I think I recall it does different things depending on the Java version and the parser it's used on.
3. I think it makes sense to suggest the alternative option when disabling doctypes is not possible

Thanks,
Florian

### DIFF
--- a/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md
@@ -388,9 +388,19 @@ documentBuilder.setEntityResolver(noop);
 Since a `javax.xml.bind.Unmarshaller` parses XML and does not support any flags for disabling XXE, it's imperative to parse the untrusted XML through a configurable secure parser first, generate a source object as a result, and pass the source object to the Unmarshaller. For example:
 
 ``` java
-//Disable XXE
 SAXParserFactory spf = SAXParserFactory.newInstance();
+
+//Option 1: This is the PRIMARY defense against XXE
 spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+spf.setXIncludeAware(false);
+
+//Option 2: If disabling doctypes is not possible
+spf.setFeature("http://xml.org/sax/features/external-general-entities", false);
+spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+spf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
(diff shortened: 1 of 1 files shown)


mackowski:
@jmanico you approved that PR, what do you think.
Also @kwwall you were involved in https://github.com/OWASP/CheatSheetSeries/pull/1097 - what do you think?

mackowski:
@jmanico and @kwwall this is related to https://github.com/OWASP/CheatSheetSeries/pull/1097

mackowski:
@n-of-one can you please take a look? I agree with @fw10 but I want someone else to review and discuss before merging.

n-of-one:
I tested this and it looks like xInclude is turned off by default, both on Oracle Java 8 and java org.glassfish.jaxb.jaxb-runtime implementations. I had set both setXIncludeAware and setNamespaceAware to true to get xInclude working.  So adding this to the recommendation would just make sure that users would not think it is safe to use this feature in practice. Not sure that is worth it tbh, but it does no harm.

As far as the FEATURE_SECURE_PROCESSING  rabbit hole is concerned: the only downside I see is that the page get's more cluttered, but I don't object to the fallback option.

fw10:
If someone would actually use `XIncludes` in their app, and copy in the recommendations from the Cheat Sheet, they may very well be vulnerable to XXE while thinking that they are implementing best practices. 

I reference the Cheat Sheet all the time to provide security guidance to software developers, and I think the snippets should be as "self-contained" as possible meaning if people just paste in the recommendations, the chances of not making any security mistakes should be as high as possible.

I think that this may be the reason why `XIncludes` is also explicitly disabled at different parts in the Cheat Sheet (e.g. https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxp-documentbuilderfactory-saxparserfactory-and-dom4j). So adding it in would be consistent.

jmanico:
Hey @n-of-one and others. I'm keeping an eye on this. Is there any more work you would like to do, considering the feedback above?

n-of-one:
My personal style is more minimalistic, but I am fine with having a slightly more heavy handed approach with less chance of someone ending up with a vulnerable configuration. 

So I'm fine with the changes proposed by @fw10 .

jmanico:
Would someone care to apply the advice of @n-of-one to this PR?

n-of-one:
Sorry, @jmanico I don't fully understand your question. My advice is just to approve this PR and merge it :)

PR #1122, opened 2023-04-21
document custom request headers as a standalone CSRF protection

gwensmu:
This change updates the CSRF Prevention Cheat Sheet to document custom request headers as a standalone CSRF protection for APIs, rather than as a "defense in depth" protection. The goal is to clarify that in some cases, custom request headers are on their own a sufficient defense against CSRF, without requiring a token. It also goes into more detail about what CORS configuration is appropriate for this defense.

This PR covers issue #1010


mackowski (dismissed):

mackowski:
hey @gwensmu there is one linter error, please fix it and we will be able to merge this change!

mackowski (approved):

mackowski:
Thanks @gwensmu!

PR #1131, opened 2023-05-09
HTTP_Headers_Cheat_Sheet: Add missing comma in Permissions-Policy

IwfY:
Missing comma in given Permissions-Policy leads to "Parse of permissions policy failed because of errors reported by structured header parser."

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy


szh (approved):
LGTM, thanks for catching!

PR #1146, opened 2023-06-09
Creating dedicated Java Security Cheat Sheet

tghosth:
Started preparing this, I will ping people when it is ready for review.

Key changes:
- Moved contents of Injection Prevention Cheat Sheet in Java to here.
- Made some very minor grammatical/context changes to the copied in content from the "Injection Prevention Cheat Sheet in Java"


jmanico (approved):
These changes look ok so far

tghosth:
Thanks @jmanico

I have now added some large crypto examples. Basically everything starting from "## Cryptography" is new and needs reviewing. 

Please let me know what you think :)

Thanks

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+// No special action needed because security actions are
+// performed at the logging policy level
+Logger logger = LogManager.getLogger(MyClass.class);
+logger.info(logMessage);
+...
+```
+
+#### Example using Logback with the OWASP Security Logging library
+
+Configuration of a logging policy to roll on 10 files of 5MB each, and encode/limit the log message using the [CRLFConverter](https://github.com/javabeanz/owasp-security-logging/wiki/Log-Forging), provided by the [OWASP Security Logging Project](https://owasp.org/www-project-security-logging/), and the *-500msg* message size limit:
```
kwwall, line 431: The link https://owasp.org/www-project-security-logging/ returns a 404 and needs fixed. (Looks like it is also wrong in the [README.md](https://github.com/augustd/owasp-security-logging/blob/master/README.md) for that project, where the readme references https://www.owasp.org/index.php/OWASP_Security_Logging_Project which redirects to the one I first mentioned, which is not found.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+- [LOG4J Configuration](https://logging.apache.org/log4j/2.x/manual/configuration.html)
+- [LOG4J Appender](https://logging.apache.org/log4j/2.x/manual/appenders.html)
+- [Log Forging](https://github.com/javabeanz/owasp-security-logging/wiki/Log-Forging) - See the Logback section about the `CRLFConverter` this library provides.
+- [Usage of OWASP Security Logging with Logback](https://github.com/javabeanz/owasp-security-logging/wiki/Usage-with-Logback)
+
+## Cryptography
+
+### General cryptography guidance
+
+- **Never, ever write your own cryptographic functions.**
```
kwwall, line 498: Agree. I also would agree that even writing our own implementations of cryptography that uses JCA/JCE is wrong-headed. Personally, I suggest replacing this entire chuck of code to instead tell Java devs to use Google Tink. Period. (By default, it uses GCM cipher mode under the hood.)

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+
+    public static byte[] Encrypt(String plaintext, byte[] nonce, SecretKey secretKey) throws Exception {
+        return CryptoOperation(plaintext.getBytes(StandardCharsets.UTF_8), nonce, secretKey, Cipher.ENCRYPT_MODE);
+    }
+
+    public static String Decrypt(byte[] ciphertext, byte[] nonce, SecretKey secretKey) throws Exception {
+        return new String(CryptoOperation(ciphertext, nonce, secretKey, Cipher.DECRYPT_MODE), StandardCharsets.UTF_8);
+    }
+
+    private static byte[] CryptoOperation(byte[] text, byte[] nonce, SecretKey secretKey, int mode) throws Exception {
```
kwwall, line 559: And where is `CryptoOperation` defined?

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+    }
+
+}
+```
+
+### Encryption for transmission
+
+Again, follow the algorithm guidance in the [OWASP Cryptographic Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#algorithms).
+
+The following code snippet shows an example of using Eliptic Curve/Diffie Helman (ECDH) together with AES-GCM to perform encryption/decryption of data between two different sides without the need the transfer the symmetric key between the two sides. Instead, the sides exchange public keys and can then use ECDH to generate a shared secret which can be used for the symmetric encryption.
```
kwwall, line 573: No, no, no. A thousand times NO. There is much more to it than this and you are essentially ignoring all the lessons learned in TLS v1.3.  If you want to illustrate this, don't show it from first principles, but instead show them how to configure TLS v1.3 and use that in Java. All the cipher suites in TLS v1.3 are thought to be safe. This, OTOH, is not, as it does not ensure that you ensure the authenticity of the intended end-point.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+
+class AesGcmSimple {
+
+    public static final String ALGORITHM = "AES";
+    public static final String CIPHER_ALGORITHM = "AES/GCM/NoPadding";
+    public static final int KEY_SIZE = 256;
+    public static final int TAG_LENGTH = 128;
+    public static final int IV_LENGTH = 12;
+
+    public static byte[] Encrypt(String plaintext, byte[] nonce, SecretKey secretKey) throws Exception {
```
kwwall, line 551: Was this copied from some .NET C# code? By convention, Java methods start with lower case.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+
+        var message = "This message to be encrypted";
+        System.out.println(message);
+
+        // Encrypt the message
+        byte[] cipherText = AesGcmSimple.Encrypt(message, nonce, secretKey);
+        System.out.println(Base64.getEncoder().encodeToString(cipherText));
+
+        // Decrypt the message
+        var message2 = AesGcmSimple.Decrypt(cipherText, nonce, secretKey);
```
kwwall, line 538: In general, this is mostly useless. The if you are encrypting this, generally you are going to persist it or transfer the encrypted blob to another party who shares the same secret key. But you are not preserving the whole IV, but only the nonce. Google Tink will work out-of-the-box, already has simple examples, and does way more than just encryption, but supports key management and key rotation. Plus it is supported by some well-respected cryptographers. When I redo ESAPI for 3.x, the symmetric encryption there is just going to be a thin wrapper around Google Tink.

kwwall (changes requested):
Some of this is just my opinion, such as disliking lengthy inline code snippets in cheat sheets. But I definitely don't approve the actual cryptography examples. Then need to be addressed IMO.

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Not sure I understand your question. I just used this method as otherwise the code from lines 560-563 gets duplicate for encrypt and decrypt.

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: I hear that. The concept was to try and provide a solution for a scenario where we want to encrypt something in addition to TLS (e.g. a particularly sensitive data item). I have seen this done before using AES to encrypt the data, then encryption the AES DEK with an RSA KEK and sending the encrypted DEK and the encrypted data to the other side. I was hoping to try this using eliptic curve and I ended up with this mechanism because ECIES is not supported out of the box in Java.

What solution would you suggest for this use case?

tghosth (commented):
Hi @kwwall,

Thanks so much for your detailed review.

I have responded to your comments and I would love your feedback to figure out how we can make this work.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: I'm more concerned because it looks like the wiki page is hosed, so it's not really a specifically a CS problem. I looked for an alternative page but didn't find one. The best way to address it is probably to leave a GitHub issue on the https://github.com/javabeanz/owasp-security-logging repo. I've been swamped with other stuff lately, but if I get some spare cycles and remember to do so (which is the harder part), I will create one.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: Do I think that? Absolutely. I observed it firsthand during the 8 years of performing numerous secure code reviews at a large bank. JCE is just too hard to for the average developer with no training in applied cryptography, or at least AppSec, to use correctly. There are lots of common errors made with it that I’ve discussed I the past. (For example, see my “[Common Developer Crypto Mistakes with illustrations in Java” slide deck](https://github.com/kwwall/presentations/blob/master/bsidescmh-2019-kwwall-prezo-common-dev-crypto-mistakes-WithSpeakersNotes.pdf) that I gave at the 2019 BSIDES Columbus conference or if you have really a lot of time to waste, a recording of the actual [2019 BSIDES prezo](https://www.youtube.com/watch?v=UvU-5PLxaO4),)

Your average Java developers think that they'll just Google “how to do encryption in Java” and most likely ends up on an ancient post to Stack Overflow where someone told them how to do encryption in AES. Sure, that example unfortunately uses ECB cipher mode, but to the Java developer, it looks easy/peazy, so they copy/paste the “solution” and think they have a secure solution. In reality, that’s one reason why we still have a lot of bad crypto. The same with a lot of instructional books. They start with ECB mode, since that’s the simplest, and a dev stops reading as soon as they get to the first workable example. Why is it like this? Because JCE was designed as primitives so that applied cryptographers could create any higher-level abstraction out of it. It wasn’t designed for those new to cryptography. You can that because JCA/JCE doesn’t make it difficult to do the wrong thing. Other the other hand, Google Tink (and to a much lesser degree, ESAPI) makes it harder to mess up.  Tink also addresses key management issues, which is probably the single largest cryptography related screw up, so that’s big.

You can find existing Java examples for Google Tink created by the Tink developers here: https://github.com/tink-crypto/tink-java/tree/main/examples

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+
+#### Symptom
+
+Injection of this type occur when the application uses untrusted user input to build an SQL query using a String and execute it.
+
+#### How to prevent
+
+Use *Query Parameterization* in order to prevent injection.
+
+#### Example
```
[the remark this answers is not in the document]
kwwall: Why not just create something like https://github.com/OWASP/CheatSheetSeries/code-snippets and drop your examples for all the cheat sheets there and then just link to them? That way it keeps all the cheat sheet material together under a single repo. That's how I was thinking anyhow. The only way that I think keeping them inline makes sense is if we can put them into collapsible sections, but I'm not sure that's possible using MarkDown.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,685 @@
...
+    List<String> colors = new ArrayList<>();
+    try (PreparedStatement pStatement = con.prepareStatement(query)) {
+        pStatement.setString(1, "yellow");
+        try (ResultSet rSet = pStatement.executeQuery()) {
+            while (rSet.next()) {
+                colors.add(rSet.getString(1));
+            }
+        }
+    }
+    Assert.assertEquals(1, colors.size());
```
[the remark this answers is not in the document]
kwwall: If you want, you could change it to Java assertions I guess, but no one is going to want to have production code referencing JUnit.

However, in reality, I think they should be changed to throw a RuntimeException and an appropriate error message.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: The question was I didn't see a definition for CryptoOperation, so it seems to be missing. It's possible that I missed it though.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: So, do you mean a hybrid solution? Generally you only do that when you have many parties who want to encrypt data and share the encrypted value with a common recipient, but also keep it from each other. In such cases, using a shared key won't work because other parties that intercept it and share that secret key could decrypt it. To the sender creates a temporary  symmetric ephemeral key (what you are referring to as your 'AES DEK'), and they encrypt their sensitive data using symmetric encryption with that. They then encrypt that symmetric ephemeral key with the public asymmetric key of the recipient and typically prepend that to the beginning of the symmetric ciphertext.

Yes, that's a legitimate use case that's useful because it greatly simplifies the key distribution problem. But that didn't look like what you were doing there. (Or maybe it was, but I was just getting tired at that point.)

Regardless, there's a hybrid example for Google Tink here: https://github.com/tink-crypto/tink-java/tree/main/examples/hybrid

jmanico:
Very astute comment, Kevin.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: The nonce is _part_ of a traditional IV. The nonce is used to ensure that a given key / IV never repeat, which is a essential to ensure since GCM is a streaming mode. See https://crypto.stackexchange.com/questions/26790/how-bad-it-is-using-the-same-iv-twice-with-aes-gcm (not the regular Stack Overflow; these are answered by those clueful in crytopgrahy) for an explanation.

Also, this may help: https://crypto.stackexchange.com/questions/5807/aes-gcm-and-its-iv-nonce-value. If not, then have a look at section 3 of RFC 5288. It can all get rather complicated. Tink solves all that and a lot more.

jmanico:
May I suggest GCM-SIV vs GCM? (Synthetic IV mode). Makes IV management and IV collisions a LOT less likely.

kwwall:
@jmanico - True enough, but I don't think it GCM-SIV is supported out-of-the-box by SunJCE (although, it may be in versions after Java 11, which we're still stuck on). It may be in Bouncy Castle though.  The other side of the equation is if GCM-SIV is not officially recommended by NIST, it won't fly in a lot of FinTech companies.

jmanico:
Bouncy, a norm in Java, has supported GVM-SIV since August 2020 (v1.65).

kwwall:
I've run across a few orgs that refused to use BC though, so there's that. And the NIST recommendation is still a showstopper if they don't recommend. Definitely wouldn't to mention it, but OTOH, I'd just prefer to see us recommend Tink for Java. It gets things like key management right and that's a big win IMO.

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Holding off on this until other discussions are completed

tghosth:
@kwwall I added more comments.

@kwwall @jmanico the original goal was to avoid using 3rd party libraries. I still think that is achievable for symmetric but for asymmetric I am not sure right now.

tghosth:
> I think the *only* way to do good crypto is with third party libraries, tink and libsodium.

Just becomes a multi-platform and/or support nightmare.

Like, which of these libsodium implementations can I rely upon?????
https://libsodium.gitbook.io/doc/bindings_for_other_languages

kwwall:
> Just becomes a multi-platform and/or support nightmare.

In what way? libsodium (and DJB's, et al's original NaCl) have ported to almost every platform and programming language as your URL shows. If you are asking "which of those implementations can / should be trusted?" that's a very different question IMO and one that's not necessarily in scope here because most major F500 companies all have their own way whether a particular library is trustworthy enough for your company to use. Each company can have vastly different criteria and different degrees of formal / informal approval. So, I don't think that's our job. IMO, it's good enough to say use Tink or NaCl or libsodium and leave it at that.

Also, dealing with dependency hell is far better than trying to deal with crypto vulnerabilities like key management issues or side-channel attacks that regular developers just are not going to understand. Devs are already used to dealing with dependency hell and a new 3rd party crypto library is barely going to add to that headache at all given that it's typical to see modern applications consist of 75 or 80% of their total code size to external libraries. (And with Node and NPM, well, don't _even_ get me started there.)  We also have SCA scanners and tools like Dependabot, Renovate, Snyk, etc. that will create PRs for you to keep all that updated. That's not a concern at all to me. But doing it correctly without subtle vulnerabilities is because on someone with suitable expertise in cryptography would see those things and those folks are rather scarce, even among the AppSec community.

tghosth:
> If you are asking "which of those implementations can / should be trusted?" that's a very different question IMO and one that's not necessarily in scope here because most major F500 companies all have their own way whether a particular library is trustworthy enough for your company to use. Each company can have vastly different criteria and different degrees of formal / informal approval. So, I don't think that's our job. IMO, it's good enough to say use Tink or NaCl or libsodium and leave it at that.

So for something like Tink which seems to be well supported by Google then I think that is ok. 

However, just to say to that the developer has to figure out which libsodium or NaCl version to use doesn't feel actionable or useful. It bothers me that we are expecting every company in the world to be able to critically appraise a cryptography library to decide which one to use, especially when that is not a static problem. That is why I wanted to avoid 3rd party libraries in the first place.


> In what way? libsodium (and DJB's, et al's original NaCl) have ported to almost every platform and programming language as your URL shows. 
> ...
> But doing it correctly without subtle vulnerabilities is because on someone with suitable expertise in cryptography would see those things and those folks are rather scarce, even among the AppSec community.

Again, Tink seems the exception because it is very usable and straightforward. LibSodium and Bounce Castle, not so much. As such, it seems as likely that people who don't fall into the "suitable expertise in cryptography" group will make mistakes with these 3rd party libraries as they will with the built-in functionality.

So whilst I accept that Tink is a good solution as long as you are only interoperating with other components which can use Tink, it doesn't solve the wider problem.

Any chance you could look at my responses to your comments @kwwall?

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: @kwwall can you confirm what the issue here as I have indeed defined the function...

tghosth (commented):
@kwwall would be great to get your feedback on the updated content and my responses to your comments :)

jmanico (dismissed):

jmanico:
I see some lint errors, can we fix those?

tghosth:
> I see some lint errors, can we fix those?

Hi @jmanico, I have fixed some linting errors and the remaining ones are caused by the HTML used for the collapsible sections I have used. I think we should allow these to be used as they increase the readability of the checklist whilst also allowing for larger code blocks.

Do you think we can update the linting configuration to allow them?

kwwall:
@tghosth - I may have some of this wrong as this is just off the top of my head, we really should ask someone like @noloader (Jeffrey Walton) who supports Crypto++ and would know this.

Regarding the IV vs nonce question, these terms are loosely used interchangeably for GCM, but technically speaking an IV that is used is always the same length as the # of bits in the cipher block, which for AES (but not Rijndael, upon which it is based and I think has a variable block side) is 128-bits. So if you pick a 96-bit nonce, the leading 32-bits are all 0 bits. So, whether this works depends on how your nonce is constructed and how many encryption operations can happen in a given millisecond or nanosecond. To make things simple, the nonce is generally monotonically increasing rather than random so that you don't need to have code that remembers which nonces have been already used. So typically that might be the current time in milliseconds or nanoseconds. That by itself usually works fine for short-lived keys as the likelihood that two encryption operations in multiple threads using the same random key happens in the same millisecond is pretty small. And if you are using time to the nanosecond level, it's probably slim to none. Where it might fail though is if you let the application store the key. If they store some AES key in (say) HashiCorp Vault or a KMS, and several applications use it, then it will increase the odds of those 96 bits being the same. It's still slim, but if they are all using current ToD and all your server times are sync'd up using NTP, then the chances increase a lot.

Hence, the cryptographers that I have conversed with on the Metzdown Cryptography mailing list had suggested (for a 128-bit cipher block size) using a random 32-bit value and combining that with a monotonically increasing 96-bit nonce value, which they recommended basing on the current time of date expressed in  milliseconds in UTC. The idea being if you had 2 encryption operations occur at the exact same millisecond that 32-bit random value would save your bacon. I think there is an RFC that specifies how t do all this for TLS using AES/GCM, but I'm not sure if it does that or not and if it does, does that apply to potentially much longer lived AES keys that might be used to encrypt . decrypt data-at-rest using AES/GCM.

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: This looks pretty good, but I still have some suggestions. However, it is getting too messy to follow so I am going to mark this conversation as 'resolved' so I can start a new thread that I hope is easier to follow.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,971 @@
...
+## Cryptography
+
+### General cryptography guidance
+
+- **Never, ever write your own cryptographic functions.**
+- Wherever possible, try and avoid writing any cryptographic code at all. Instead try and either use pre-existing secret management solutions or the secret management solution provided by your cloud provider. For more information, see the [OWASP Secrets Management Cheat Sheet](/Secrets_Management_Cheat_Sheet).
+- Make sure your application or protocol can easily support a future change of cryptographic algorithms.
+- Use your package manager wherever possible to keep all of your packages up to date. Watch the updates on your development setup, and plan updates to your applications accordingly.
+- Wherever possible, use a trusted and well known implementation library rather than using the libraries built into JCA/JCE as it is far too easy to make cryptographic errors with them.
+- We will show examples below based on Google Tink which is an opinionated library for performing encryption operations.
```
kwwall, line 513: I'm not sure what you mean by "opinionated library"? A inanimate object like a library, can't really hold an opinion, and library developers almost always do. (I personally have more opinions than I can remember!)

But my guess is that you are trying to say something else here. Give us a hint or use a thesaurus.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,971 @@
...
+      
+        // Read the keyset into a KeysetHandle.
+        KeysetHandle handle =
+        TinkJsonProtoKeysetFormat.parseKeyset(
+            new String(Files.readAllBytes( Paths.get("/home/josh/aead_test_keyset.json")), UTF_8), InsecureSecretKeyAccess.get());
+
+        String message = "This message to be encrypted";
+        System.out.println(message);
+
+        String metadata = "Info about message";
```
kwwall, line 569:
```java
        String metadata = "Info about message";
```

While this works on a very abstract level, I'm not sure developers will find this particularly useful. Even if this is straight from a Tink example, I don't think devs will find it particularly helpfil. Most often the AEAD metadata will be something that will be important to the recipient. E.g., 

```java
    String metadata = "Sender: tghosth";
```

or maybe
```java
    String metadata = "KeyID: 0xA4D4BB0785";  // indicate which key to use.
```

or maybe even something important about the message contents itself, such as:

```java
    String metadata = "Message expiration: 2023-06-21T13:34:56-05:00";
```

etc. It's something that might be useful to the recipient for processing, but it's confidentiality is not important. But ideally, try to give them ideas of what the AEAD metadata is typically used for instead of writing a general "Info about message".

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,971 @@
...
+        );
+
+        KeysetHandle bobPublicKey = bob.getPublicKey();
+
+        // This keypair generation shoud be reperformed every so often in order to
+        // obtain a new shared secret to avoid a long lived shared secret.
+
+        // Alice encrypts a message to send to Bob
+        String plaintext = "Hello, Bob!";
+        String metadata = "Info about the message";
```
kwwall, line 759: Ditto. See previous comments on this. Maybe use something different for this example.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,971 @@
...
+
+
+}
+```
+
+</details>
+
+#### Asymmetric example using built in JCA/JCE classes
+
+If you absolutely cannot use a separate library, it is still possible to use the built JCA/JCE classes but it is strongly recommended to have a cryptography expert review the full design and code.
```
kwwall, line 826: Suggest ending this sentence as:
> ... have a cryptography expert review the full design and code, as even the most trivial error can severely weaken your encryption,

cheatsheets/Java_Security_Cheat_Sheet.md
kwwall: I haven't checked, but Tink likely supports [Cryptographic Message Syntax (CMS)](https://en.wikipedia.org/wiki/Cryptographic_Message_Syntax). If .NET also support that (which wouldn't surprise me, that's what I would suggest as that's commonly used in a lot of standards. If that is not available, export it all as JSON and have .NET parse it. That's how I would address the interoperability issue. Way better than writing something that's likely to be error prone and that I don't presently have time for a line-by-line detailed code review.

kwwall (dismissed):
This looks way better. I think the collapsible sections are a huge win for readability and I like using Tink. It is supported in just about everything but C# and I suggested a path forward for interoperability with that.

We may need further changes later, but we can create specific issues for that. The CS should be considered a work-in-progress anyhow since Best Practices is not a static target. For that reason and for not allowing the perfect to become the enemy of the good, I am going to approve these changes, even though I hope you will consider addressing the new comments that I raised.

tghosth:
> @tghosth - I may have some of this wrong as this is just off the top of my head, we really should ask someone like @noloader (Jeffrey Walton) who supports Crypto++ and would know this.
> 
> Regarding the IV vs nonce question, these terms are loosely used interchangeably for GCM, but technically speaking an IV that is used is always the same length as the # of bits in the cipher block, which for AES (but not Rijndael, upon which it is based and I think has a variable block side) is 128-bits. So if you pick a 96-bit nonce, the leading 32-bits are all 0 bits. So, whether this works depends on how your nonce is constructed and how many encryption operations can happen in a given millisecond or nanosecond. To make things simple, the nonce is generally monotonically increasing rather than random so that you don't need to have code that remembers which nonces have been already used. So typically that might be the current time in milliseconds or nanoseconds. That by itself usually works fine for short-lived keys as the likelihood that two encryption operations in multiple threads using the same random key happens in the same millisecond is pretty small. And if you are using time to the nanosecond level, it's probably slim to none. Where it might fail though is if you let the application store the key. If they store some AES key in (say) HashiCorp Vault or a KMS, and several applications use it, then it will increase the odds of those 96 bits being the same. It's still slim, but if they are all using current ToD and all your server times are sync'd up using NTP, then the chances increase a lot.
> 
> Hence, the cryptographers that I have conversed with on the Metzdown Cryptography mailing list had suggested (for a 128-bit cipher block size) using a random 32-bit value and combining that with a monotonically increasing 96-bit nonce value, which they recommended basing on the current time of date expressed in milliseconds in UTC. The idea being if you had 2 encryption operations occur at the exact same millisecond that 32-bit random value would save your bacon. I think there is an RFC that specifies how t do all this for TLS using AES/GCM, but I'm not sure if it does that or not and if it does, does that apply to potentially much longer lived AES keys that might be used to encrypt . decrypt data-at-rest using AES/GCM.

@kwwall at this point all my code uses a cryptographically random 12 byte IV so this seems sufficient. (In the "hybrid" example I was actually deriving the IV from the shared secret but I have now changed that.)

[This answer](https://crypto.stackexchange.com/a/41610) seems to suggest that a cryptographically random 12 byte IV is an acceptable choice.

As such, I can't see a specific further action at this time.

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Added a link for those who are unclear

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Great point, I have updated this to be more specific

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Updated accordingly

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: See [above](https://github.com/OWASP/CheatSheetSeries/pull/1146#discussion_r1241143817).

tghosth:
I have submitted #1156 to fix the markdown linting config to prevent the error being shown

tghosth:
#1156 and #1158 got merged and linting is now all green so this can now be merged @szh @jmanico

tghosth:
Any chance this can now be merged @szh @jmanico ?

szh:
I'm not so well-versed in Java so I'll wait for another maintainer to review this PR. @kwwall

kwwall:
@szh - Will take a look.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -507,10 +507,10 @@ other log viewing/analysis tools that don't expect the log data to be pre-HTML e
...
+- If you cannot use a pre-existing secret management solution, try and use a trusted and well known implementation library rather than using the libraries built into JCA/JCE as it is far too easy to make cryptographic errors with them.
 - Make sure your application or protocol can easily support a future change of cryptographic algorithms.
 - Use your package manager wherever possible to keep all of your packages up to date. Watch the updates on your development setup, and plan updates to your applications accordingly.
-- Wherever possible, use a trusted and well known implementation library rather than using the libraries built into JCA/JCE as it is far too easy to make cryptographic errors with them.
-- We will show examples below based on Google Tink which is an opinionated library for performing encryption operations.
+- We will show examples below based on Google Tink which is an [opinionated](https://www.baeldung.com/cs/opinionated-software-design#opinionated-design) library for performing encryption operations.
```
kwwall, line 513: I still _really_ dislike the word "opinionated" here, regardless of the link defining it. In part, that's because of the normal use of the word, and in part, that link starts off with "Generally, we can see the opinionated design as an **_oracle_** that indicates the right way to accomplish tasks" (emphasis mine). The word "oracle" has generally negative connotations in the cryptography world (e.g., as in "padding oracle attack").

**Suggestion**: Change the sentence
> We will show examples below based on Google Tink which is an opinionated library for performing encryption operations.

to something like

> We will show examples below based on Google Tink, which is a  library created by cryptography experts for using cryptography safely (in the sense of minimizing foolish coding mistakes for regular developers).

Yes, it's more words and perhaps not as pithy, but I think it better captures the design intent of its creators. The whole point is while this is "opinionated" in the sense of the Baeldung reference you provided, to me it better explains why a developer should be more apt to trust it than say JCA/JCE which doesn't restrict implementation choices but is very much an underlying cause of the mess we are in with cryptography.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -561,12 +563,14 @@ public class App {
...
-            new String(Files.readAllBytes( Paths.get("/home/josh/aead_test_keyset.json")), UTF_8), InsecureSecretKeyAccess.get());
+            new String(Files.readAllBytes( Paths.get("/home/fred/aead_test_keyset.json")), UTF_8), InsecureSecretKeyAccess.get());
 
         String message = "This message to be encrypted";
         System.out.println(message);
 
-        String metadata = "Info about message";
+        // Add some relevant context about the encrypted data that should be verified
+        // on decryption
+        String metadata = "Sender: fredbloggs";
```
kwwall, line 573: I personally would have went with something like "/home/pete/aead_test_keyset.json" above and
```java
       String metada = "Sender: Pete Moss";
```
because it's hard to hard for me to resist inserting puns, but that's just me. But this is better than the original. More typically, the 'Sender' metadata would be either DN of a user from (say) AD or an email address, but this works.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -598,20 +602,24 @@ class AesGcmSimple {
...
-If you absolutely cannot use a separate library, it is still possible to use the built JCA/JCE classes but it is strongly recommended to have a cryptography expert review the full design and code.
+If you absolutely cannot use a separate library, it is still possible to use the built JCA/JCE classes but it is strongly recommended to have a cryptography expert review the full design and code, as even the most trivial error can severely weaken your encryption.
 
 The following code snippet shows an example of using AES-GCM to perform encryption/decryption of data.
 
-A few contraints/pitfalls with this code:
+A few constraints/pitfalls with this code:
 
 - It does not take into account key rotation or management which is a whole topic in itself.
 - It is important to use a different nonce for every encryption operation, even if the same key is used.
```
kwwall, line 616: (We could emphasis the fact that 'nonce' is short for 'number used once'.) But my critique of this bullet is not that but the wording "_even_ if the same key is used". It is _**critical**_ that is has to be different when a key is reused. The whole point is that the same key / IV pair (where here, nonce is your IV) must always be unique for all time (or as long as any adversary might be collecting your ciphertexts). Otherwise it can let to decrypting some of those collected ciphertexts later.

**Strong suggestion**: Change 'even' to 'especially'.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -598,20 +602,24 @@ class AesGcmSimple {
...
 
-If you absolutely cannot use a separate library, it is still possible to use the built JCA/JCE classes but it is strongly recommended to have a cryptography expert review the full design and code.
+If you absolutely cannot use a separate library, it is still possible to use the built JCA/JCE classes but it is strongly recommended to have a cryptography expert review the full design and code, as even the most trivial error can severely weaken your encryption.
 
 The following code snippet shows an example of using AES-GCM to perform encryption/decryption of data.
 
-A few contraints/pitfalls with this code:
+A few constraints/pitfalls with this code:
 
 - It does not take into account key rotation or management which is a whole topic in itself.
```
kwwall, line 615: I think this would be a good place for a reference to that link you had referenced earlier in your response to one of my comments: 
https://crypto.stackexchange.com/questions/41601/aes-gcm-recommended-iv-size-why-12-bytes/41610#41610

In particular, the 2nd answer is very relevant here and I think that fits well with this bullet. It's way less obtuse than most of the relevant NIST special publications.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -598,20 +602,24 @@ class AesGcmSimple {
...
 - It is important to use a different nonce for every encryption operation, even if the same key is used.
 - The key will need to be stored securely.
```
kwwall, line 617: Perhaps a link to the [Cryptographic Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html) is in order here???

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -677,6 +685,8 @@ class AesGcmSimple {
 
 </details>
 
+_**End of code snippet.**_
```
kwwall, line 688: Agree with the previous reviewer, I think we can ditch this now that these are in collapsible sections.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -756,7 +768,10 @@ class App {
 
         // Alice encrypts a message to send to Bob
         String plaintext = "Hello, Bob!";
-        String metadata = "Info about the message";
+        
+        // Add some relevant context about the encrypted data that should be verified
+        // on decryption
+        String metadata = "Sender: alicesmith";
```
kwwall, line 774: As mentioned earlier, this often is either a DN or an email. To illustrate the point, using something like 'alice@example.com' rather than 'alicesmith' would be preferred here. Email addresses are unique; combination of a givenname / surname, not so much.

cheatsheets/Java_Security_Cheat_Sheet.md
```diff
@@ -770,12 +785,17 @@ class App {
 
         // Bob encrypts a message to send to Alice
         String plaintext2 = "Hello, Alice!";
+
+        // Add some relevant context about the encrypted data that should be verified
+        // on decryption
+        String metadata2 = "Sender: bobjones";
```
kwwall, line 791: Ditto. Use an email here. E.g., 'bob@example.com'.

kwwall (approved):
The comments are _only_ suggestions. For some, I feel more strongly than others, but I wouldn't recommend rejecting the PR should you decide to reject any of them. (Although, I may follow them up by a PR of my own. ;-)

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Addressed in  #1160

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Not going with a pun but I made it an email address

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Addressed in #1160

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Addressed in #1160

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: That is linked at the very beginning

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Made into an email

cheatsheets/Java_Security_Cheat_Sheet.md
tghosth: Made into an email

PR #1157, opened 2023-06-25
Initial start for prototype pollution prevention cheatsheet

tghosth:


jmanico (dismissed):
Looking good so far, you can also start in the staging directory if you want more rapid commits https://github.com/OWASP/CheatSheetSeries/tree/master/cheatsheets_draft

tghosth:
Thanks @jmanico, I think this will do for now and then people add to it as they see fit. The important thing is that the information was captured.

tghosth:
@jmanico can you now merge it?

tghosth:
Any chance this can now be merged @szh @jmanico ?

cheatsheets/Prototype_Pollution_Prevention_Cheat_Sheet.md
```diff
@@ -0,0 +1,56 @@
...
+let obj = {__proto__:null};
+```
+
+### Use object "freeze" and "seal" mechanisms
+
+You can also use the `Object.freeze()` and `Object.seal()` APIs to prevent built-in prototypes from being modified however this can break the application if the libraries they use modify the built-in prototypes.
+
+### Node.js configuration flag
+
+Node.js also offers the ability to remove the `__proto__` property completely using the `--disable-proto=delete` flag. Note this is a defence in depth measure.
```
szh, line 45: Let's keep spellings consistent with US English
```suggestion
Node.js also offers the ability to remove the `__proto__` property completely using the `--disable-proto=delete` flag. Note this is a defense in depth measure.
```

cheatsheets/Prototype_Pollution_Prevention_Cheat_Sheet.md
```diff
@@ -0,0 +1,56 @@
...
+
+Prototype pollution is still possible using constructor.prototype properties but removing `__proto__` helps reduce attack surface and prevent certain attacks.
```
szh, line 47:
```suggestion
Prototype pollution is still possible using `constructor.prototype` properties but removing `__proto__` helps reduce attack surface and prevent certain attacks.
```

szh (changes requested):
Two nits. Also should we include examples of what not to do, and show example exploits? Even just one simple example would be helpful for understanding the danger.

cheatsheets/Prototype_Pollution_Prevention_Cheat_Sheet.md
tghosth: Fixed in #1161

PR #1158, opened 2023-06-26
Fix linting config

tghosth:
Fix a minor bug in the syntax of the lint configuration which made it not work 🤦‍♂️

Sorry :(


szh (approved):
LGTM

PR #1166, opened 2023-07-10
Bug - Incorrect markdown link in Cloud Arch Sheet 

Josh-Beck:
# Bug Fix

Used the wrong brackets type for the `iam-access` link in the most recent PR. Quick fix here.


szh (approved):
LGTM

PR #1170, opened 2023-07-12
Add reference to Secrets Management CS in Key Storage CS

szh:
This PR adds a reference to the secrets management cheat sheet in the storage section of the key management cheat sheet. I was looking through the sheet and was surprised that there was no link provided so I added one. While I was at it I also fixed some typos.

- [x] Any references to websites have been formatted as [TEXT](URL)

### DIFF
--- a/cheatsheets/Key_Management_Cheat_Sheet.md
+++ b/cheatsheets/Key_Management_Cheat_Sheet.md
@@ -4,10 +4,10 @@
 
 This Key Management Cheat Sheet provides developers with guidance for implementation of cryptographic key management within an application in a secure manner. It is important to document and harmonize rules and practices for:
 
-1. key life cycle management (generation, distribution, destruction)
-2. key compromise, recovery and zeroization
-3. key storage
-4. key agreement
+1. Key life cycle management (generation, distribution, destruction)
+2. Key compromise, recovery and zeroization
+3. Key storage
+4. Key agreement
 
 ## General Guidelines and Considerations
 
@@ -108,11 +108,11 @@ Review `NIST SP 800-57` (Recommendation for Key Management) for recommended guid
 
 ### Memory Management Considerations
 
-Keys stored in memory for a long time can become "burned in". This can be mitigated by splitting the key into components that are frequently updated. `NIST SP 800.57`).
+Keys stored in memory for a long time can become "burned in". This can be mitigated by splitting the key into components that are frequently updated. `NIST SP 800-57`).
 
-Loss or corruption of the memory media on which keys and/or certificates are stored, and recovery planning, according to `NIST SP 800.57`.
+Loss or corruption of the memory media on which keys and/or certificates are stored, and recovery planning, according to `NIST SP 800-57`.
 
-Plan for the recovery from possible corruption of the memory media necessary for key or certificate generation, registration, and/or distribution systems, subsystems, or components as recommended in `NIST SP 800.57`.
+Plan for the recovery from possible corruption of the memory media necessary for key or certificate generation, registration, and/or distribution systems, subsystems, or components as recommended in `NIST SP 800-57`.
 
 ### Perfect Forward Secrecy
 
@@ -130,7 +130,7 @@ There are several reasons for this:
 
 ### Cryptographic Module Topics
 
-According to `NIST SP800-133`, cryptographic modules are the set of hardware, software, and/or firmware that implements security functions (including cryptographic algorithms and key generation) and is contained within a cryptographic module boundary to provide protection of the keys.
+According to `NIST SP 800-133`, cryptographic modules are the set of hardware, software, and/or firmware that implements security functions (including cryptographic algorithms and key generation) and is contained within a cryptographic module boundary to provide protection of the keys.
 
 ## Key Management Lifecycle Best Practices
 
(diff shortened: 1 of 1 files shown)


jmanico (approved):

cheatsheets/Key_Management_Cheat_Sheet.md
```diff
@@ -151,13 +151,15 @@ The generated keys shall be transported (when necessary) using secure channels a
 1. Developers must understand where cryptographic keys are stored within the application. Understand what memory devices the keys are stored on.
 2. Keys must be protected on both volatile and persistent memory, ideally processed within secure cryptographic modules.
 3. Keys should never be stored in plaintext format.
-4. Ensure all keys are stored in cryptographic vault, such as a [hardware security module](https://en.wikipedia.org/wiki/Hardware_security_module) (HSM) or isolated cryptographic service.
+4. Ensure all keys are stored in a cryptographic vault, such as a [hardware security module](https://en.wikipedia.org/wiki/Hardware_security_module) (HSM) or isolated cryptographic service.
 5. If you are planning on storing keys in offline devices/databases, then encrypt the keys using Key Encryption Keys (KEKs) prior to the export of the key material. KEK length (and algorithm) should be equivalent to or greater in strength than the keys being protected.
 6. Ensure that keys have integrity protections applied while in storage (consider dual purpose algorithms that support encryption and Message Code Authentication (MAC)).
```
kwwall, line 156: For for FIPS compliance AES and Triple DES, I would suggest recommending "key wrapping" for this purpose, as it was designed for this for this purpose. There generally is "key wrapping" and "key wrapping with padding", the later when you are wrapping keys that is longer than the cipher's block size. (E.g., AES has a cipher block size of 128-bits, so you would generally use "key wrapping with padding" for wrapping AES key sizes of 192 or 256 bits.)

See [NIST SP 800-38F](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-38F.pdf) and https://en.wikipedia.org/wiki/Key_Wrap for details. As the Wikipedia article points out SIV (see RFC 5297) is probably a better replacement, but I doubt SIV is FIPS compliant. Key wrapping also is not relevant for algorithms other than AES or TDEA.

Java's javax.crypto.Cipher class supports this via the 'opmode' of` Cipher.WRAP_MODE` and `Cipher.UNWRAP_MODE`. Some JCE providers (e.g., BouncyCastle) I believe also implemented GCM-SIV (RFC 8452).  Beyond Java, I'm not sure how well either of these are is supported. There are probably some C and C++ implementations, but I'm not sure about things like Python, Perl, .NET C#, etc. YMMV.

cheatsheets/Key_Management_Cheat_Sheet.md
```diff
@@ -151,13 +151,15 @@ The generated keys shall be transported (when necessary) using secure channels a
...
 7. Ensure that standard application level code never reads or uses cryptographic keys in any way and use key management libraries.
 8. Ensure that keys and cryptographic operation is done inside the sealed vault.
 9. All work should be done in the vault (such as key access, encryption, decryption, signing, etc).
```
kwwall, line 159: This may not always be practical (i.e., scalable) especially with HSM's because of costs. Cloud-based KMS's are probably a good alternative for those storing their encrypted data in the cloud.

cheatsheets/Key_Management_Cheat_Sheet.md
```diff
@@ -108,11 +108,11 @@ Review `NIST SP 800-57` (Recommendation for Key Management) for recommended guid
 
 ### Memory Management Considerations
 
-Keys stored in memory for a long time can become "burned in". This can be mitigated by splitting the key into components that are frequently updated. `NIST SP 800.57`).
+Keys stored in memory for a long time can become "burned in". This can be mitigated by splitting the key into components that are frequently updated. `NIST SP 800-57`).
```
kwwall, line 111: If would be useful to include the relevant SP 800-57 section numbers as well. It might even have the side effect of having some people read the special publication, or at least the relevant parts of it.  Ditto with the other references to this as well.

kwwall (commented):
Key management is probably the area that is the most difficult problem in cryptography, so you certainly are braver than me for even attempting this. I've pretty much avoided it because I've found trying to answer these questions generally results in a lot more questions, many of which the correct answer starts out with either "It depends..." or "What is your threat model?". And I've never had the resources and often knowledge to answer. However, my first suggestion for this CS would be to admit all this in an introductory paragraph up front. When it comes to practicality though, you will find that most of time the "good enough" approach here is sufficient because rarely is cryptography the weakest link. (In fact, in my entire 25 years that I've been doing AppSec and applied cryptography, the only time I recall cryptography being the weakest line was in the original Wi-Fi specification of Wired Equivalent Privacy (WEP), which apparent included no cryptographers in its specs.)

Lastly, these comments are suggestions. There is nothing that I see here that I wouldn't approve. I was just trying to fill in the gaps a bit.  [I would have let you know this by selecting 'Suggestion' or whatever, but when I got to the end and it said 'Finish code review', at that put it didn't give me any of the usual options so I must have done something out of order. But hence this end comment.]

szh:
@kwwall Thank you so much for the suggestions. I'm going to leave this in my GH inbox and hopefully get back to it after I finish work on #1171 and some other projects. If anyone else wants to take a crack at it first that's also great.

Issue #1187, opened 2023-08-29
Update: DotNet Security Cheat Sheet


DanCarterGMSL:
## What is missing or needs to be updated?

In the "Hashing" section, it suggests that [Microsoft.AspNetCore.Cryptography.KeyDerivation.Pbkdf2](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/consumer-apis/password-hashing) is not available on .NET Framework, only on .NET Core. However, it's actually available on .NET Framework >= 4.6.1.

## How should this be resolved?

The advice should be changed to recommend using `KeyDerivation.Pbkdf2` on .NET Framework >= 4.6.1.

szh:
Hey @DanCarterGMSL, Can you please post a link to where this is class is available in .NET Framework 4.6.1?

DanCarterGMSL:
@szh The NuGet package targets .NET Standard 2.0 (see https://www.nuget.org/packages/Microsoft.AspNetCore.Cryptography.KeyDerivation/2.1.0#dependencies-body-tab), which is supported by .NET Framework 4.6.1 (but not 4.6). Hence this implementation is available on .NET Framework 4.6.1 and later.

I've just confirmed this by creating a .NET Framework 4.6 app and I couldn't install the NuGet package; whereas when I changed it to v4.6.1 I could install the NuGet package and access the `Pbkdf2()` method.

szh:
Awesome, thank you for the details.

PR #1188, opened 2023-08-29
Fix recommended password hashing implementations in .NET

DanCarterGMSL:
This PR covers issue #1187.


szh (changes requested):
I've left a comment on issue #1187 asking for clarification

jmanico:
nice work!

Issue #1217, opened 2023-10-16
Update: Deserialization_Cheat_Sheet.html#java


yakovsh:
## What is missing or needs to be updated?
Information about deserialization libraries in Java that use other formats besides Java's native serialization format

## How should this be resolved?
Add information about these libraries and how they can be used safely.

szh:
Yup, fixed in #1218

PR #1218, opened 2023-10-16
Deserialization_Cheat_Sheet.html#java - Other formats/libraries

yakovsh:
Proposed PR to add information about other Java deserialization libraries and how they can be used safely

This PR covers issue #1217.


szh (changes requested):
Thank you for all that detailed info! This is great, just needs some stylistic fixes - see the lint results.

szh (approved):
LGTM! Thank you so much!

PR #1226, opened 2023-11-16
Update Password_Storage_Cheat_Sheet.md

jmanico:
Big update from Robert Thorton

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -79,7 +79,7 @@ There is no golden rule for the ideal work factor - it will depend on the perfor
 
 One key advantage of having a work factor is that it can be increased over time as hardware becomes more powerful and cheaper.
 
-The most common approach to upgrading the work factor is to wait until the user next authenticates, then re-hash their password with the new work factor. The different hashes will have different work factors and hashes may never being upgraded if the user doesn't log back into the application. Depending on the application, it may be appropriate to remove the older password hashes and require users to reset their passwords next time they need to login in order to avoid storing older and less secure hashes.
+The most common approach to upgrading the work factor is to wait until the user next authenticates, then re-hash their password with the new work factor. The different hashes will have different work factors and hashes may never be upgraded if the user doesn't log back into the application. Depending on the application, it may be appropriate to remove the older password hashes and require users to reset their passwords next time they need to login in order to avoid storing older and less secure hashes.
 
 ## Password Hashing Algorithms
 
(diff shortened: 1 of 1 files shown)


cheatsheets/Password_Storage_Cheat_Sheet.md
```diff
@@ -153,24 +151,24 @@ The work factor for PBKDF2 is implemented through an iteration count, which shou
...
 
 ### International Characters
 
-Ensure your hashing library is able to accept a wide range of characters and is compatible with all Unicode codepoints. Users should be able to use the full range of characters available on modern devices, in particular mobile keyboards. They should be able to select passwords from various languages and include pictograms. Prior to hashing the entropy of the user's entry should not be reduced. Password hashing libraries need to be able to use input that may contain a NULL byte.
+Your hashing library must be able to accept a wide range of characters and should be compatible with all Unicode codepoints, so users can use the full range of characters available on modern devices--especially mobile keyboards. They should be able to select passwords from various languages and include pictograms. Prior to hashing the entropy of the user's entry should not be reduced, and password hashing libraries need to be able to use input that may contain a NULL byte.
```
szh, line 174:
```suggestion
Your hashing library must be able to accept a wide range of characters and should be compatible with all Unicode codepoints, so users can use the full range of characters available on modern devices - especially mobile keyboards. They should be able to select passwords from various languages and include pictograms. Prior to hashing the entropy of the user's entry should not be reduced, and password hashing libraries need to be able to use input that may contain a NULL byte.
```

szh (changes requested):
I left some proofreading comments, but otherwise looks great! Thank you!

kwwall:
@jmanico - Just an FYI, I am starting the requested code review today. Had been working on a PR for a new ESAPI release.

kwwall:
@jmanico - Sorry I was late to this requested review. This is great work. The only minor comment that I would have is the paragraph under the "Peppering" section that states:
```text
For example, one peppering strategy is hashing the passwords as usual (using a password hashing algorithm) and then using HMAC or encrypting the hashes with a symmetrical encryption key before storing the password hash in the database, with the key acting as the pepper.
```
I personally would recommend changing the part "then using HMAC or encrypting the hashes with a symmetrical encryption key" (should be "symmetric encryption key", not "symmetrical" BTW), to instead ONLY refer to using HMACs. I'd recommend revising it as something like this:
```text
For example, one peppering strategy is hashing the passwords as usual (using a password hashing algorithm) and then using an  HMAC (e.g., HMAC-SHA256, HMAC-SHA512, depending on the desired output length) on the original password hash before storing the password hash in the database, with the pepper acting as the HMAC  key.
```
I recommend using an HMAC over symmetric encryption because there's just less to get wrong using HMACs than there is correctly implementing the encryption / decryption using a symmetric cipher.

Regardless of whether that change is made or not though, the last part of that paragraph, where it states:
```text
... with the key acting as the pepper.
```
is incorrect. It should say
```text
... with the pepper acting as the key.
```
Otherwise, looks good.

Issue #1254, opened 2023-12-07
Update: [XSS Filter Evasion Cheat Sheet]


daivdwe:
## What is missing or needs to be updated?
XSS Locator (Polygot) is missing some (important) break-out-chars.
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.md#xss-locator-polygot

- JavaScript Multiline String declarator: \`
- Newline (breaks out of "normal" JS Strings delimited by " or '): `
` (URL encoded CRLF newline: %0D%0A)

## How should this be resolved?
Please add both to XSS Locators "polygot test XSS payload.".

szh:
Great idea. Do you want to submit a PR for this?

daivdwe:
Thank you, yes.

PR #1255, opened 2023-12-07
Update: [XSS Filter Evasion Cheat Sheet] #1254

daivdwe:
XSS Locator (Polygot):
- added newline
- added `

This PR covers issue #1254.

### DIFF
--- a/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.md
+++ b/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.md
@@ -18,7 +18,10 @@ This is a normal XSS JavaScript injection, and most likely to get caught but I s
 
 The following is a "polygot test XSS payload." This test will execute in multiple contexts including html, script string, js and URL. Thank you to [Gareth Heyes](https://twitter.com/garethheyes) for this [contribution](https://twitter.com/garethheyes/status/997466212190781445).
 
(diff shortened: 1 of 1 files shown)


cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.md
```diff
@@ -18,7 +18,10 @@ This is a normal XSS JavaScript injection, and most likely to get caught but I s
 
 The following is a "polygot test XSS payload." This test will execute in multiple contexts including html, script string, js and URL. Thank you to [Gareth Heyes](https://twitter.com/garethheyes) for this [contribution](https://twitter.com/garethheyes/status/997466212190781445).
```
szh, line 19:
```suggestion
The following is a "polyglot test XSS payload." This test will execute in multiple contexts including html, script string, js and URL. (Based on this [tweet](https://twitter.com/garethheyes/status/997466212190781445) by [Gareth Heyes](https://twitter.com/garethheyes)).
```

szh (changes requested):
Looks good, we just need to update the attribution since it's no longer an exact copy. I also think polygot is a typo and it should say polyglot (also in the header)

daivdwe:
> Looks good, we just need to update the attribution since it's no longer an exact copy. I also think polygot is a typo and it should say polyglot (also in the header)

It seems like your suggested change did not change the header.
I can change it tomorrow or accept another change request today. If it is ok for you? 

Thanks

Issue #1252, opened 2023-12-06
Update: [Django Security Cheat Sheet] - default url to admin panel


Cmaiek:
## What is missing or needs to be updated?

It is advisable to modify the default url leading to the admin panel (example.com/admin/), in order to make automated attacks this bit more difficult.

## How should this be resolved?

In the default app folder within your project, locate the urls.py file managing the top-level urls. Within the file, modify the ''urlpatterns'' variable (list), so that the url leading to admin.site.urls is different than "admin/".

I.e. change the variable from:
```python
urlpatterns = [
    path("admin/", admin.site.urls),
    ...
    ]
```

to:
```python
urlpatterns = [
    path("something-other-than-admin/", admin.site.urls),
    ...
    ]
```

jmanico:
Great! Looking forward to a PR :)

Cmaiek:
> Great! Looking forward to a PR :)

Can I assume this means "the issue has been discussed and approved"? ;)

jmanico:
Looking forward to a PR

PR #1261, opened 2023-12-14
DjangoCheatSheet - modification of admin panel url

Cmaiek:
As discussed in the issue, I have added a brief section on how and why to change a default url leading to the admin panel.

Also, I added ```django-axes``` as an alternative package for handling brute-force attacks.

This PR covers issue #1252

### DIFF
--- a/cheatsheets/Django_Security_Cheat_Sheet.md
+++ b/cheatsheets/Django_Security_Cheat_Sheet.md
@@ -10,7 +10,7 @@ The Django framework provides some built-in security features that aim to be sec
 
 - Always keep Django and your application's dependencies up-to-date to keep up with security vulnerabilities.
 - Ensure that the application is never in `DEBUG` mode in a production environment. Never run `DEBUG = True` in production.
-- Use packages like [`django_ratelimit`](https://django-ratelimit.readthedocs.io/en/stable/) to prevent brute-force attacks.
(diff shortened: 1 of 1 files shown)


jmanico (changes requested):
The text you provided is mostly accurate, but it contains a few minor issues that could be improved for clarity and correctness:

Phrase "to make automated attacks this bit more difficult": This phrase is a bit informal and could be made clearer. A more formal and clearer way to express this might be "to slightly increase the difficulty for automated attacks."

Use of "url": The term "URL" is an acronym for "Uniform Resource Locator" and is typically written in uppercase.

Code Formatting: In Python code, it's important to maintain consistent formatting for readability. The phrase "urlpatterns variable (list)" could be more clearly written with backticks to denote code elements.

Clarity in Instructions: While the instructions are clear, adding a bit more detail about why changing the admin URL is beneficial could enhance the explanation.

Here's a revised version of your text with these corrections:

"It is advisable to modify the default URL leading to the admin panel (example.com/admin/), in order to slightly increase the difficulty for automated attacks. Here’s how to do it:

In the default app folder within your project, locate the urls.py file managing the top-level URLs. Within the file, modify the urlpatterns variable, a list, so that the URL leading to admin.site.urls is different from 'admin/'. This approach adds an extra layer of security by obscuring the common endpoint used for administrative access."
This revised version improves clarity, formality, and technical accuracy.

Cmaiek:
Thanks for your feedback, @jmanico! Should I keep the code examples, or drop those as well?

jmanico (approved):

PR #1344, opened 2024-02-26
Copy editing on SQL Injection Prevention cheat sheet

neurosie:
Initially I was just going to fix a typo that I noticed but I went and edited the whole page.

Notably I removed this whole section:

> #### Language-Specific Recommendations for Prepared Statements:
> 
> --Java EE: `PreparedStatement()` with bind variables
> --.NET: Parameterized queries like `SqlCommand()` or `OleDbCommand()` with bind variables
> --PHP: PDO with strongly typed parameterized queries (using bindParam())
> --Hibernate: `createQuery()` with bind variables (called named parameters in Hibernate)
> --SQLite: `sqlite3_prepare()` to create a [statement object](http://www.sqlite.org/c3ref/stmt.html)
> 
> Occasionally, prepared statements can harm performance. If this occurs, you should a) strongly validate all data or b) instead of using a prepared statement, escape all user supplied input using an escaping routine specific to your database vendor as described below.

3 out of 5 of the bullet points were redundant to other blocks in that section. There's already a link out to a longer sheet with recommendations for way more languages at the end of the prepared statement section. 

As for the last sentence, I did a little research on the relationship between prepared statements and performance and it's complex, sometimes helping and sometimes hurting, with mitigations being very situation dependent. Seems bad to recommend the less secure method of input escaping when that's heavily recommended against elsewhere in this same sheet.

### DIFF
--- a/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md
@@ -122,7 +122,7 @@ However, stored procedures require execute rights, a role that is not available 
 
 **Safe Java Stored Procedure Example**:
 
-The following code example uses Java's implementation of the stored procedure interface (`CallableStatement`) to execute the same database query. The `sp_getAccountBalance` stored procedure has top be predefined in the database and use the same functionality as the query below.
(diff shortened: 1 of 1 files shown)


mackowski (approved):
Loog good to me. Thanks @neurosie

PR #1381, opened 2024-04-15
Fix openssl dhparam example command

sehraramiz:
current openssl dhparam example command throws the following error:
```
dhparam: Extra option: "2048"
```


szh (approved):
Thank you for finding this. I've confirmed it on my machine. Change LGTM

jmanico (approved):

PR #1422, opened 2024-06-04
Fix typo in OAuth2_Cheat_Sheet.md

jsphpl:
Checked for occurences of the same typo in the file & repo


szh (approved):
Good catch, thanks for fixing!

PR #1454, opened 2024-07-25
Remove duplicated example of query string parsing

reznakt:
Removes a duplicate of the example at line 218.


szh (approved):
LGTM, good catch.

jmanico (approved):

PR #1460, opened 2024-08-02
Addition of a Checklist - Create The Essential Security Checklist for… #984

SNCherry:


jmanico:
I am ok with this general checklist. But can we add do this is a way that references all of the other cheatsheets? The main list of cheatsheets is here. https://cheatsheetseries.owasp.org/  I would prefer that each sections links to the relevant cheatsheet or cheatsheets...

kwwall:
I was just about to say, this looks like an almost verbatim rehash of what's mentioned on the referenced link, https://www.yslingshot.com/security-checklist/. As such, this looks too much like trying to take advantage of OWASP Cheat Sheats as your personal ad agency. I'm not sure the OWASP board would approve, especially since Slingshot is a commercial endeavor selling services (see https://www.yslingshot.com/services/) and not a non-profit. So, as this is presented, I think it needs extensive rewrite. (I also am generally not in favor of checklists. We have plenty of those already. If one is going to take a checklist approach, then the OWASP ASVS would be a better reference for this, very vague checklist.)

Just my $.02. But at a minimum, I think this needs to do what @jmanico is suggesting and reference the other Cheat Sheets in the relevant sections rather than pointing to https://www.yslingshot.com/security-checklist/.

jmanico:
Thank you Kevin. Unless this is refactored in a very significant way, I am going with Kevin's perspective on this.

szh:
Given the above comments, I'll close this for now. Feel free to reopen if major changes are made that would change the decision.

Issue #1508, opened 2024-10-01
Update: REST Security Cheat Sheet


JoostK:
## What is missing or needs to be updated?

The following section 

https://github.com/OWASP/CheatSheetSeries/blame/9c67e3ac45e2bfacd96e5dc324757cfb80e9ec1d/cheatsheets/REST_Security_Cheat_Sheet.md#L101

mentions:

> Reject requests containing unexpected or missing content type headers with HTTP response status `406 Unacceptable` or `415 Unsupported Media Type`.

forgoing requests with `Content-Length: 0` where no content is present. When there is no content, there isn't a suitable `Content-Type` header that applies, but this section doesn't specify the exact requirements here.

The [coreruleset/coreruleset](https://github.com/coreruleset/coreruleset/blob/a6036b984405798f5accec6a1fc5caaa90de46ac/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L709-L727) repository implements a detection rule that does account for `Content-Length: 0` and exempts the request from requiring a `Content-Type` header, as would be expected.

## How should this be resolved?
I'd propose to at least clarify the requirements here, going as far as suggesting that `Content-Length: 0` should not require a `Content-Type` header at all.

JoostK:
Thanks for the quick reply! Sure thing, I've opened #1510 with a proposal. Happy to adjust as requested.

randomstuff:
> forgoing requests with Content-Length: 0 where no content is present. When there is no content, there isn't a suitable Content-Type header that applies, but this section doesn't specify the exact requirements here.

Or it it? Let's say, you can PUT a text/plain or URI list to your API in order to override some text or some list of URIs. Your API would expect the proper content-type and reject the request if the proper content type is not user **but** you might still want to use it with an empty file in some cases.

In other words, an empty `text/plain` is still a `text/plain` and an empty `text/uri-list` is still a `text/uri-list`.

jmanico:
I get the feeling you are both right.

The approach taken by the OWASP Core Rule Set (CRS), which exempts requests with Content-Length: 0 from requiring a Content-Type header, is practical. It aligns with the principle that when no content exists, a Content-Type header is irrelevant. This is especially useful in cases where an empty body is legitimately used to indicate certain states in an API call, such as "no change" or a placeholder.

[...]

PR #1510, opened 2024-10-02
Clarify that requests with Content-Length: 0 do not require a Content…

JoostK:
…-Type header

This PR fixes issue #1508.

### DIFF
--- a/cheatsheets/REST_Security_Cheat_Sheet.md
+++ b/cheatsheets/REST_Security_Cheat_Sheet.md
@@ -98,7 +98,7 @@ A REST request or response body should match the intended content type in the he
 
 ### Validate request content types
 
-- Reject requests containing unexpected or missing content type headers with HTTP response status `406 Unacceptable` or `415 Unsupported Media Type`.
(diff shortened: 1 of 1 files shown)


randomstuff:
>  For requests with `Content-Length: 0` no `Content-type` header is expected.

It might be ambiguous clear whether this is 1) "we might not have a `Content-type` in some cases" or 2) "you must not use a `Content-Type`". I would claim that this is (1).

jmanico:
I think "we might not have a Content-type in some cases" is the better direction here. There are some API's that will require it even with empty content.

JoostK:
> > For requests with `Content-Length: 0` no `Content-type` header is expected.
> 
> Is there an ambiguity here whether this is 1) "we might not have a `Content-type` in some cases" or 2) "you must not use a `Content-Type`"? I would claim we want (1).

I was trying to avoid this but the current wording does indeed leave this ambiguity. I mentioned some potential alternatives in https://github.com/OWASP/CheatSheetSeries/issues/1508#issuecomment-2387892001.

jmanico (approved):

mackowski (approved):

Issue #1532, opened 2024-11-06
Potentially conflicting information in Password Storage: Pre-Hashing Passwords with bcrypt


regunakyle:
So the recent Okta username-longer-than-52 security advisory got me interested in best practice for handling longer-than-72 password length with bcrypt. 

Someone in X/Twitter suggests we can SHA256 hash the password before applying bcrypt. The [OWASP cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pre-hashing-passwords-with-bcrypt) explicitly discourage this:

> An alternative approach is to pre-hash the user-supplied password with a fast algorithm such as SHA-256, and then to hash the resulting hash with bcrypt (i.e., `bcrypt(base64(hmac-sha256(data:$password, key:$pepper)), $salt, $cost)`). This is a dangerous (but common) practice that should be avoided due to [password shucking](https://www.youtube.com/watch?v=OQD3qDYMyYQ) and other issues when [combining bcrypt with other hash functions](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html).

This [blog post cited in the above excerpt](https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.htm) explained that null bytes might show up after hashing if we don't encode the raw output with base64 or hex.
Quoting the blog post:
> You are 100% safe if you do one of the following:
> - Use straight bcrypt (don’t pre-hash)
> - Use hex output from the pre-hash
> - **Base64 encode the raw output of a pre-hash**

However, the OWASP excerpt above uses `bcrypt(base64(hmac-sha256(data:$password, key:$pepper)), $salt, $cost)` as an example of unsafe prehashing practice (this is how I interpret it anyway). To me, they feel like conflicting information.

Could someone clarify this? 

Note: I created a relevant question on the [security stack exchange](https://security.stackexchange.com/questions/279407/revisiting-pre-hash-password-before-applying-bcrypt-to-avoid-restricting-passwo).

jmanico:
I really think "just dont use bcrypt" since pre-hashing can cause problems and NOT pre-hashing can cause problems. We are at an era where Argon2id is the right call and bcrypt is not.

regunakyle:
[...]

PR #1551, opened 2024-11-28
Updated the bcrypt description as discussed in #1532

raphaelahrens:
In #1532 there was a side discussion by @ja2048  and @jmanico on why the CS mentioned that HMAC would be vulnerable to password shucking, but the linked talk mentioned that even a hash with a pepper would prevent it. So I correcteed the guidance.

I further removed the mentioning of PBKDF2, since this suggests that bcrypt + PBKDF2 would give FIPS-140 compliance.

I also added that bcrypt should only be used for legacy system which don't have agron2 or scrypt, as discussed in #1532

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -119,7 +119,7 @@ These configuration settings provide an equal level of defense. The only differe
 
 ### bcrypt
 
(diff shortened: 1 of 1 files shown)


jmanico (approved):

jmanico:
This is excellent work. Thank you.

mackowski (approved):

Issue #1538, opened 2024-11-09
Update: Forgot Password


gl4nce:
## What is missing or needs to be updated?

The section [Offline Methods](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#offline-methods) contains wrong information. The real-world examples are all leading to backup codes in connection with MFA, which is out of scope of this CS.

## How should this be resolved?

The Section should be removed. AFAIK, there is no secure offline method for account recovery. Instead of removing it, this could be clearly stated there.

mackowski:
Backup Codes are listed and described as example of Offline Methods https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#backup-codes.
I think that we should not remove it but improve. @jmanico what do you think?

jmanico:
That works for me @mackowski - backup-codes and MFA are a legitimate part of forgot passwords flows and I want to get this right :)

gl4nce:
In this CS, MFA is mixed up with backup codes and all examples refer to MFA. If you want to talk about MFA recovering, that should be stated clear and not mixed up with best practice forgotten password methods without MFA.

Do you have any real world example for an application with offline backup codes without MFA?

jmanico:
Yea, I have used backup codes to reset MFA when I lost the MFA device or similar.

jmanico:
I get your point. But since backup codes could be used for MFA recovery, they seem just as strong as MFA itself, if not stronger... For example you can use backup codes to sign in to your Google account if you've forgotten your password:

gl4nce:
100% agree.

My intention for this issue is to make it clear that this section is related to MFA and has nothing to do with default password recovery methods without of MFA. Alternatively, the section can be removed completely because MFA is out of scope of that CS.

That's all.

mackowski:
@gl4nce can you make a PR for this change. I think that we should clarify that. We should remove that section and replace it with link to the Multifactor Authentication Cheat Sheet to the 'Resetting MFA' section.

gl4nce:
@mackowski PR for removing the *Backup Codes* subsection? If this is accepted, I can do that, of course.

mackowski:
[...]

PR #1553, opened 2024-12-03
Update Forgot_Password_Cheat_Sheet.md

gl4nce:
- removed section about backup codes from CS because it's out of scope
- added hint for MFA recovery methods with link to corresponding CS

This PR fixes issue #1538.


jmanico (commented):
Should we also mention that these codes should be random?

kwwall (changes requested):
I'm fine with the general direction you are taking this, but can we please link to the specific section on "[MFA Recovery](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Multifactor_Authentication_Cheat_Sheet.md#resetting-mfa)" instead of just the general MFA CS?

kwwall:
> Should we also mention that these codes should be random?

Yes, but in the MFA Reset section rather than here.

mackowski (approved):

kwwall (approved):
LGTM

Issue #1520, opened 2024-10-21
Update: Authentication_Cheat_Sheet.md


NicolaiSoeborg:
## What is missing or needs to be updated?

The current Authentication page is not aligned with:
 * best practice for mandatory password rotation (which is to avoid mandatory periodically rotations, see a comparison here: https://søb.org/password/ )
 * OWASP's own "A07:2021 – Identification and Authentication Failures" recommendations (which tells to follow NIST SP 800-63b)

Some good quotes:

> NIST: _Verifiers SHOULD NOT require memorized secrets to be changed arbitrarily (e.g., periodically)_

> Microsoft: _Password expiration requirements do more harm than good_

## How should this be resolved?

This line does not make sense:
> Ensure credential rotation when a password leak occurs, or at the time of compromise identification.

Password leak occurs all the time, and we do not want to force people to change their (potentially) good password for a new (potentially) bad password. Instead we should "annoy" the user with 2FA/MFA, tell them to use a password manager, etc

Suggestion for a new phrasing:

> Avoid requiring periodic password changes; instead, encourage users to pick a strong passwords and enable MFA.

jmanico:
NIST suggests that the only time to rotate passwords is in the case of password data being compromised. So I suggest:

Avoid requiring periodic password changes; instead, encourage users to pick a strong passwords and enable MFA. Consider password rotation only in case of compromise or when authenticator technology is changed.

eth-developer03:
Hi @jmanico, could you please assign this issue to me? I'd like to take ownership and resolve it

eth-developer03:
@jmanico ?

jmanico:
All set, thank you @szh

PR #1570, opened 2024-12-29
issue #1520 Updated authentication.md 

eth-developer03:
# You're A Rockstar


Dockerfile
```diff
@@ -1,9 +1,33 @@
...
 
 EXPOSE 8000
 
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    python3-pip \
+    dos2unix \
+    && pip3 install mkdocs
+
+RUN dos2unix scripts/Generate_Site_mkDocs.sh
```
kwwall, line 30: Okay. I do not like this. If someone using Windows previously edited a bash script so it ended up with \r\n as EOL terminators rather than just \n and broke the script, the better way to deal with this is to create an issue for the particular script and just fix the script. This is a kludge, and if we start using this, we may eventually find ourselves having to do this for all the bash scripts under 'scripts'.

Furthermore, if this is becoming a problem, a better way to address it is to have contributors set the 'core.input' attribute in their .gitconfig file. In ESAPI, we mention this:
> Finally, we recommend setting the git property 'core.autocrlf' to 'input' in your $HOME/.gitconfig file; e.g., that file should contain something like this:
> 
> [core]
> &nbsp;&nbsp;&nbsp;&nbsp;autocrlf = input

and since people have been using that, we haven't had that problem. (And, if they ignore that and persist on messing up bash scripts by using \r\n for EOL terminators, we force them to program for 3 months while wearing mittens.)

So, of course, if we don't execute dos2unix, we don't need to install it either line 27 so it can be removed from there as well.

cheatsheets/Authentication_Cheat_Sheet.md
kwwall, line 1: I''m fine with these changes. I don't know if we actually want a blank line to start this .md file, but if it's consistent with the other CS pages, it's fine. (I'm just too lazy to check right now.)  As for the content changes, it LGTM.

kwwall (changes requested):
IMO,the Docker change (if we want it; I think it's kludge), should be addressed in a separate GitHub issue and PR. Please pull it out of this PR.

PR #1601, opened 2025-01-30
This PR fixes issue #1300

### LINKED ISSUE #1300: Shortlinks for cheatsheets
Does the cheatsheet site have a mechanism for shorter links, e.g.

https://cheatsheetseries.owasp.org/SSRF

would redirect to:

https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html

Farreeda:
# Add URL Shortening for Cheat Sheets

## Description
This PR adds automatic URL shortening functionality to make cheat sheets more accessible through memorable shortcuts. Instead of typing long URLs, users can now access cheat sheets using abbreviated paths.

For example:
- `/XSS.html` → `cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html`
- `/SSRF.html` → `cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html`
- `/SQLi.html` → `cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html`

## Changes
- Added URL shortening logic to `Generate_Site_mkDocs.sh`
- Automatically generates shortcuts based on first letters of each word in filenames
- Creates clean URL .html version (`/XSS.html`)
- Adds HTML redirects with proper meta refresh tags
- Includes verification steps to ensure shortcuts are created correctly

## Testing
After running
Test shortcuts:
   ```bash
   curl -L http://localhost:8000//CCS.html
   curl -L http://localhost:8000/CCSP.html
   curl -L http://localhost:8000/DOS.html
   ```

## Benefits
- Easier to remember and share URLs
- Faster access to frequently used cheat sheets
- Maintains compatibility with existing full URLs
- Works with both clean URLs and .html extensions

## Related Issues
Closes #1300


jmanico (dismissed):
I am approving just to move this along but can other folks take a closer look at this before we merge?

szh:
This is fantastic. I'm going to do a full review on it hopefully in the next day or two but I'm really excited about this. Thank you!

jmanico:
This is fantastic - thank you very very much!

mackowski:
Build is failing here @Farreeda

scripts/Generate_Site_mkDocs.sh
```diff
@@ -95,53 +95,91 @@ if ! python -m mkdocs build; then
...
+    cp "$redirect_file" "${redirect_file}.html"
+    
+    # Verify creation
+    if [ -f "$redirect_file" ] && [ -f "${redirect_file}.html" ]; then
+       # echo "✅ Created shortcuts:"
+        echo "  - /${shortcut}"
+        echo "  - /${shortcut}.html"
+    else
+        #echo "❌ Failed to create shortcuts for ${shortcut}"
+    fi
```
szh, line 150: The script fails here, I assume because there's nothing in the else clause except for the comment. Uncommenting the line fixes this.
*However*, when I run this locally and run `make serve` the shortcuts don't seem to work. I also don't see where the script reads redirects.yml - it looks like the script is automatically generating the shortcuts which makes me wonder how it will handle possible duplicates. Can you please clarify the process?

Farreeda: Thank you for your feedback, @szh! I appreciate your insights and apologize for not testing every case @mackowski. The redirects.yml file now includes abbreviations for each path, and it can be easily modified and updated as needed. The `XSS` and `auth` issues should be resolved now, but please note that abbreviations should be followed by `.html`
Thank you, @jmanico, for bringing attention to my PR!

mackowski (dismissed):

mackowski:
@szh can you review it again?

szh:
Sorry for the delay, I've been busy. I'll try to review it this week.

Farreeda:
The issue happens because the shortcut files (without .html) are being treated as raw files instead of actual HTML pages when accessed in a browser. The browser doesn't know how to interpret them, so it downloads them instead of rendering.

> I'm ready to approve. Just one question - I see it generates both the clean urls like `/xss` and with the html suffix, such as `/xss.html` but the clean URL version doesn't actually get served. @Farreeda do you know if there's a way to get those to work? If not, we should probably not bother creating them.

The issue happens because the shortcut files (without .html) are being treated as raw files instead of actual HTML pages when accessed in a browser. The browser doesn't know how to interpret them, so it downloads them instead of rendering.

szh:
OK, so maybe we should just not create them? Or maybe there's some kind of config file (like a .htaccess file) that can instruct the server to serve them as html?

mackowski:
@Farreeda and @szh what do we need to do to release it? :)

Farreeda:
Thank you for re-checking my PR @mackowski.
The feature works with the /{shortcut}.html
/{shortcut} alone is treated as a raw file and gets downloaded instead :/
We can merge it and open a new issue for this.

jmanico:
ping

Farreeda:
Hello. After a year.
Skills upgraded. Works now.

jmanico:
Website build failed. not sure what to approve here.

szh:
```
Generate_Site_mkDocs.sh: line 21: syntax error near unexpected token `done'
```

mackowski:
I am closing this because this in inactive PR

PR #1634, opened 2025-03-21
Add HMAC validation pseudocode.

chr15m:
I thought the "Pseudo-Code For Implementing HMAC CSRF Tokens" section could use an example of how to validate the CSRF token once returned by the client to the server. This PR adds the pseudocode for that.

Checklist:

### DIFF
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -107,6 +107,36 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
 response.setCookie("csrf_token=" + csrfToken + "; Secure") // Set Cookie without HttpOnly flag
 ```
 
+And the following code demonstrates validation of the CSRF token that is sent back from the client:
+
+```code
+// Get the CSRF token from the request
+csrfToken = request.getParameter("csrf_token") // From form field or header
+
+// Split the token to get the randomValue
+const [hmacFromRequest, randomValue] = csrfToken.split(".");
+
+// Recreate the HMAC with the current session and the randomValue from the request
+secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key
+sessionID = session.sessionID // Current authenticated user session
+message = sessionID.length + "!" + sessionID + "!" + randomValue.length + "!" + randomValue
+
+// Generate the expected HMAC
+expectedHmac = hmac("SHA256", secret, message)
+
+// Compare the HMAC from the request with the expected HMAC
+if (!constantTimeEquals(hmacFromRequest, expectedHmac)) {
+    // HMAC validation failed, reject the request
+    response.sendError(403, "Invalid CSRF token")
+    return
+}
+
+// CSRF validation passed, continue processing the request
+// ...
+```
+
+Note: The `constantTimeEquals` function should be used to compare the HMACs to prevent timing attacks. This function compares two strings in constant time, regardless of how many characters match.
+
 ### Naive Double-Submit Cookie Pattern (DISCOURAGED)
 
 The _Naive Double-Submit Cookie_ method is a scalable and easy-to-implement technique which uses a cryptographically strong random value as a cookie and as a request parameter (even before user authentication). Then the server verifies if the cookie value and request value match. The site must require that every transaction request from the user includes this random value as a hidden form value or inside the request header. If the value matches at server side, the server accepts it as a legitimate request and if they don't, it rejects the request.
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -94,20 +94,23 @@ Below is an example in pseudo-code that demonstrates the implementation steps de
 
 ```code
 // Gather the values
-secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key
+secret = getSecretSecurely("CSRF_SECRET") // HMAC secret key
 sessionID = session.sessionID // Current authenticated user session
-randomValue = cryptographic.randomValue() // Cryptographic random value
+randomValue = cryptographic.randomValue(64) // Cryptographic random value
 
 // Create the CSRF Token
-message = sessionID.length + "!" + sessionID + "!" + randomValue.length + "!" + randomValue // HMAC message payload
+message = sessionID.length + "!" + sessionID + "!" + randomValue.length + "!" + randomValue.toHex() // HMAC message payload
 hmac = hmac("SHA256", secret, message) // Generate the HMAC hash
-csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash to create the final CSRF token. Avoid using the `message` because it contains the sessionID in plain text, which the server already stores separately.
+// Add the `randomValue` to the HMAC hash to create the final CSRF token.
+// Avoid using the `message` because it contains the sessionID in plain text,
+// which the server already stores separately.
+csrfToken = hmac.toHex() + "." + randomValue.toHex()
 
 // Store the CSRF Token in a cookie
 response.setCookie("csrf_token=" + csrfToken + "; Secure") // Set Cookie without HttpOnly flag
 ```
 
-Below is an example in psuedo-code that demonstrates validation of the CSRF token once it is sent back from the client:
+Below is an example in pseudo-code that demonstrates validation of the CSRF token once it is sent back from the client:
 
 ```code
 // Get the CSRF token from the request
@@ -119,7 +122,7 @@ const hmacFromRequest = tokenParts[0];
 const randomValue = tokenParts[1];
 
 // Recreate the HMAC with the current session and the randomValue from the request
-secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key
+secret = getSecretSecurely("CSRF_SECRET") // HMAC secret key
 sessionID = session.sessionID // Current authenticated user session
 message = sessionID.length + "!" + sessionID + "!" + randomValue.length + "!" + randomValue
 
@@ -130,6 +133,7 @@ expectedHmac = hmac("SHA256", secret, message)
 if (!constantTimeEquals(hmacFromRequest, expectedHmac)) {
     // HMAC validation failed, reject the request
     response.sendError(403, "Invalid CSRF token")
+    logError("Invalid CSRF token", hmacFromRequest, expectedHmac)
     return
 }
 
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -107,11 +107,11 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
 response.setCookie("csrf_token=" + csrfToken + "; Secure") // Set Cookie without HttpOnly flag
 ```
 
-And the following code demonstrates validation of the CSRF token that is sent back from the client:
+Below is an example in psuedo-code that demonstrates validation of the CSRF token once it is sent back from the client:
 
 ```code
 // Get the CSRF token from the request
-csrfToken = request.getParameter("csrf_token") // From form field or header
+csrfToken = request.getParameter("csrf_token") // From form field, cookie, or header
 
 // Split the token to get the randomValue
 const tokenParts = csrfToken.split(".");
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -114,7 +114,9 @@ And the following code demonstrates validation of the CSRF token that is sent ba
 csrfToken = request.getParameter("csrf_token") // From form field or header
 
 // Split the token to get the randomValue
-const [hmacFromRequest, randomValue] = csrfToken.split(".");
+const tokenParts = csrfToken.split(".");
+const hmacFromRequest = tokenParts[0];
+const randomValue = tokenParts[1];
 
 // Recreate the HMAC with the current session and the randomValue from the request
 secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key


jmanico (commented):
Splitting the token throws me off, can I get other folks to review this code?

chr15m:
If it helps I can remove the destructuring to make it that step a bit more straightforward.

jmanico:
Anything you can do to simplify this so it's more generic would help me approve this now. :)

chr15m:
I pushed that change.

jmanico (approved):

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
 response.setCookie("csrf_token=" + csrfToken + "; Secure") // Set Cookie without HttpOnly flag
 ```
 
+And the following code demonstrates validation of the CSRF token that is sent back from the client:
```
kwwall, line 110: This should probably also explicitly refer to "pseudo-code" same as the previous code section.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
...
+
+```code
+// Get the CSRF token from the request
+csrfToken = request.getParameter("csrf_token") // From form field or header
```
kwwall, line 114: If you're getting this from a cookie, as implied on line 107, then this would probably be 'request.getHeader("Cookie")' and then you'd have to parse out the "csrf_token" cookie from any other cookies returned.

Also, related, despite the comment on line 107, I really don't see why you would want to create the cookie without the HttpOnly flag since I can never see a legitimate need to every have client-side JavaScript access it.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
...
+
+// Split the token to get the randomValue
+const tokenParts = csrfToken.split(".");
+const hmacFromRequest = tokenParts[0];
+const randomValue = tokenParts[1];
+
+// Recreate the HMAC with the current session and the randomValue from the request
+secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key
```
kwwall, line 122: While passing secrets as environment variables is less bad than certain other scenarios, it is still not truly secure. E.g., see:

- https://www.trendmicro.com/en_us/research/22/h/analyzing-hidden-danger-of-environment-variables-for-keeping-secrets.html
- https://blog.diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/

We generally to not allow this except as a last resort. Rather, retrieve the secret CSRF_SECRET in something like AWS KMS or HashiCorp Vault, etc. (If we change this here, a corresponding change is needed on line 97 as well.)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
...
+sessionID = session.sessionID // Current authenticated user session
+message = sessionID.length + "!" + sessionID + "!" + randomValue.length + "!" + randomValue
+
+// Generate the expected HMAC
+expectedHmac = hmac("SHA256", secret, message)
```
kwwall, line 127: HMac-SHA256 is going to be expected to return a byte array, not a string, so for here and on line 103, I would recommend changing the call from "hmac" to something like (say) "hmacAsHexString". If you don't do that, then someone trying to implement the pseudo-code on line 104, where it has:

```
   csrfToken = hmac + "." + randomValue;
```

is likely to come out with a mangled string.  Alternate;y, you could do something like:

```
   csrfToken = hmac.asHex() + "." + randomValue.asHex();
```

here. That would save from having to base64-encoding the cookie value.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
...
+
+// Compare the HMAC from the request with the expected HMAC
+if (!constantTimeEquals(hmacFromRequest, expectedHmac)) {
+    // HMAC validation failed, reject the request
+    response.sendError(403, "Invalid CSRF token")
```
kwwall, line 132: I would also recommend logging this and bogus csrfToken that is being passed in here as well.

kwwall (changes requested):
Generally, this looks pretty solid, but I think there's a bit of clean-up needed because every hmac function that I've ever used has returned a byte array, not a String. And ditto for any call to a CSRNG returning a random value.

Lastly, rather than call a call to `cryptograpic.randomValue()` on line 99, I think you want to use something like:
```
    cryptograpic.randomValue(64)    // return 64-byte random value
```
Otherwise, people will be asking how long of a random value do we need here.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: From what I understand the major web frameworks (Django, Next.js etc.) recommend that client side JavaScript takes the cookie from the httpOnly=false cookie and send it back as a header during a fetch POST. I think this is what @psibean discovered as well.

https://docs.djangoproject.com/en/5.1/howto/csrf/#using-csrf-protection-with-ajax

psibean: My understanding is that getting the token from the cookie is also Angulars default interceptor behaviour for CSRF protection:

> When performing HTTP requests, an interceptor reads a token from a cookie, by default XSRF-TOKEN, and sets it as an HTTP header, X-XSRF-TOKEN.

From: https://v17.angular.io/guide/http-security-xsrf-protection

@kwwall for a long time I have completely agreed with you. The cookie should be `httpOnly`, then the token should be provided to the frontend via a response payload (HTML, JSON), to then be included in a custom header.

However not only do the existing OWASP docs suggest otherwise (that line 107), but many existing implementations already work this way out of the box.

I get the impression the general consensus is, if something can access your cookie, that's a different vulnerability and it isn't the responsibility of CARF protection.

It's be great if we could get the cheat sheet cleaned up and filled with accurate information, because now I'm just not sure any more.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: Will update.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: This is what the pseudocode in the existing section says. I kept it consistent.

Complicating pseudocode with details of proprietary services like AWS & HashiCorp doesn't seem like it will age well.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: I think those types of implementation details don't belong in pseudocode. The existing pseudocode gives return values the same treatment.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: Sure. That seems like something implementors should know to do.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: I've updated the comment to add "cookie":
// From form field, cookie, or header
I think `request.getParameter("csrf_token")` is abstract enough to convey the idea in pseudo-code.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
psibean: Maybe just `getSecret("CSRF_SECRET")` ?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
kwwall: I missed the part that we were using cookies for compatibility with some JS frameworks and Django. Of course, I could just as easily argue that none of those frameworks should be relying on a Set-Cookie header to set a Cookie for a CSRF token. Seems to me it would have been better for them to settle on some standard header name such as "X-CSRF" or whatever. At least then if a framework botches the implementation, you don't need to worry about the browser automatically returning that cookie. AJAX works, except when it doesn't. But that's water under the bridge I guess.

As to why I was nothing things like `request.getHeader("csrf_token")'` being better pseudo-code, it because in the not-too-distant future, shortly after we publish it, someone somewhere is going to point their favorite generative AI LLM at this pseudo code and ask it to generate code in their favorite programming language. If the humans would actually scrutinize and thoroughly test the generated code, the problem of trying to read this a a query parameter on a URL would be noticed, but someone will inevitably miss it. Many of them overly trust AI-generated code and are largely ignorant of AppSec. So I just figured to try to make the inevitable a bit less likely. YMMV.

kwwall: @psibean wrote:
> @kwwall for a long time I have completely agreed with you. 

I hope not (on the "completely" part). If that's the case, one of us is redundant. :smile:

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
kwwall: > Maybe just `getSecret("CSRF_SECRET")` ? Environment variable, secret, or otherwise (future proof).

Make it `getSecretSecurely` or similar, and I could stand behind that.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
kwwall: Depends. May be true if you do think that no one will feed the pseudo-code to a generative AI. But I don't believe that for a minute.

jmanico:
Are we ok to merge this yet?

kwwall:
I'm not going to _stop_ anyone from merging it, but I think we could do better.

jmanico:
Do you have time to fix what is wrong @kwwall so we can go live?

chr15m:
If you give me a couple of days I can tweak the PR with some of @kwwall's suggestions.

kwwall:
@jmanico - Let's give @chr15m  a chance first, okay?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
 response.setCookie("csrf_token=" + csrfToken + "; Secure") // Set Cookie without HttpOnly flag
 ```
 
+Below is an example in psuedo-code that demonstrates validation of the CSRF token once it is sent back from the client:
```
randomstuff, line 110: "pseudo-code"

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -107,6 +107,38 @@ csrfToken = hmac + "." + randomValue // Add the `randomValue` to the HMAC hash t
...
+// Get the CSRF token from the request
+csrfToken = request.getParameter("csrf_token") // From form field, cookie, or header
+
+// Split the token to get the randomValue
+const tokenParts = csrfToken.split(".");
+const hmacFromRequest = tokenParts[0];
+const randomValue = tokenParts[1];
+
+// Recreate the HMAC with the current session and the randomValue from the request
+secret = readEnvironmentVariable("CSRF_SECRET") // HMAC secret key
```
randomstuff, line 122: Any need to link to some recommendations about the secret?

* randomly generated, with proper entropy
* not reused between environments
* not reused for another usage
* not reused for authenticating the data (unless some "usage" field is included which is not done currently)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: What do you mean by putting scare quotes around the word "pseudo-code"? Do you mean that you do not think the code presented is pseudo-code? Or is it the opposite and you are highlighting that the code is pseudo-code and not real code?

randomstuff: Sorry I meant there was a typographic error: "psuedo-code" → "pseudo-code"

chr15m: Ah! Thanks. 😊

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mackowski: I think we have somwhere in cheatsheet about secrets generation so ideally we just link to that. We are avoiding duplicating content because than we will have instructions how to generate secrets in 3-4 places.

chr15m: I note in the explanation immediately above it has this, with a link to the key storage bit:

> Ideally, store this key as discussed in the [Cryptographic Storage page](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#key-storage).

chr15m:
Alright, I've pushed some changes addressing the concerns raised. I also went ahead and modified the existing psuedo-code that I didn't write, as @kwwall commented on that as well and that makes it consistent with my changes.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
chr15m: Have added psuedo-code logging below.

kwwall (approved):
LGTM. Sorry this took so long for me to get back to.

jmanico (approved):

PR #1652, opened 2025-04-24
XSS_Filter_Evasion_Cheat_Sheet: Fix Markdown link syntax

noelleleigh:
The link under the "Malformed A Tags" heading was wrapped in backticks, preventing it from being rendered into an `<a>` element.


szh (approved):
LGTM, thank you for the fix!

PR #1653, opened 2025-04-26
Update links to Django docs

mjnaderi:
Updated links to Django documentation:

- Django 3.2 is not supported and is unsafe, so I updated to Django 5.2 docs.
- The link to django security middleware was broken. Fixed the link.

The cheatsheet suggested setting `SECURE_PROXY_SSL_HEADER` to `True` (if behind a proxy). But according to the [Django docs](https://docs.djangoproject.com/en/5.2/ref/settings/#secure-proxy-ssl-header), this settings should be set to a tuple with two elements, not a boolean value. So I removed the `to TRUE` phrase.


jmanico (approved):

andrzejsydor (approved):

szh (approved):
LGTM

PR #1693, opened 2025-06-12
Secret cheat sheet: added segmentation scheme

sergiomarotco:
# Adding segmentation scheme for Secret Management System.
Secrets_Management_Cheat_Sheet_Vault_scheme drawio

### DIFF
--- a/cheatsheets/Secrets_Management_Cheat_Sheet.md
+++ b/cheatsheets/Secrets_Management_Cheat_Sheet.md
@@ -100,7 +100,7 @@ generally is much more difficult.
     use Java's
     [GuardedString](https://docs.oracle.com/html/E28160_01/org/identityconnectors/common/security/GuardedString.html)
     or .NET's
-    [SecureString](https://learn.microsoft.com/en-us/dotnet/api/system.security.securestring#string-versus-securestring)
+    [SecureString](https://learn.microsoft.com/dotnet/api/system.security.securestring#string-versus-securestring)
     which are designed to solve precisely this problem.
 
 - **Zeroing Memory:** After a secret has been used, the memory it occupied
@@ -217,7 +217,7 @@ Placing the Secret Management System in your own infrastructure means that all o
 
 Access to secrets must be carried out exclusively through the System's front-end server and only via a secure protocol.
 
-! [Scheme](https://raw.githubusercontent.com/OWASP/CheatSheetSeries/master/assets/Secrets_Management_Cheat_Sheet_Vault_scheme.drawio.png)
+![Scheme](https://raw.githubusercontent.com/OWASP/CheatSheetSeries/master/assets/Secrets_Management_Cheat_Sheet_Vault_scheme.drawio.png)
 
 ## 3 Continuous Integration (CI) and Continuous Deployment (CD)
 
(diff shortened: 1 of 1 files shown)


sergiomarotco:
@jmanico hi! 😅

cheatsheets/Secrets_Management_Cheat_Sheet.md
sergiomarotco, line 1: Will be work after merging

sergiomarotco:
@kwwall need approval

andrzejsydor (approved):

kwwall (approved):
My only comment is I think that the Middleware component is optional. There are lots of examples of where some front-end application server is using a back end key service  like HashiCorp Vault or AWS KMS to retrieve something like a DB password and there is no middleware service layer at all. I would advise as either adding some explanation around that or minimally denote it as an optional layer. Otherwise, LGTM.

sergiomarotco:
> My only comment is I think that the Middleware component is optional. There are lots of examples of where some front-end application server is using a back end key service like HashiCorp Vault or AWS KMS to retrieve something like a DB password and there is no middleware service layer at all. I would advise as either adding some explanation around that or minimally denote it as an optional layer. Otherwise, LGTM.

We should publish best practices, right?
Using HashiCorp Vault as an example:
- FRONTWARE - NginX/Apache or API Management/Security software for balancing for first 
- MIDDLEWARE - Vault server;
- BACKEND - Consul Server.

sergiomarotco:
@szh

cheatsheets/Secrets_Management_Cheat_Sheet.md
```diff
@@ -211,6 +211,14 @@ A secret management solution should provide the capability to store at least the
 
 Note: if you don't store metadata about the secret nor prepare to move, you will increase the probability of vendor lock-in.
 
+### 2.12 Network segmentation of Secrets Management System
+
+Placing the Secret Management System in your own infrastructure means that all of its elements must be located in isolated networks (VLAN). Other network devices must not be located in networks where the Secret Management System servers are located. Each element (application) must be located in a separate suitable network according to the [Network Segmentation Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Network_Segmentation_Cheat_Sheet.md).
```
szh: Incorrect linking
```suggestion
Placing the Secret Management System in your own infrastructure means that all of its elements must be located in isolated networks (VLAN). Other network devices must not be located in networks where the Secret Management System servers are located. Each element (application) must be located in a separate suitable network according to the [Network Segmentation Cheat Sheet](Network_Segmentation_Cheat_Sheet.md).
```

szh (changes requested):
I've held off on reviewing this because my opinion is that it needs more explanation. We should assume that a reader of the secrets management cheat sheet is an app developer who has limited understanding of network segmentation. I feel that this text is a bit too advanced for such a reader, and the diagram needs more explanation.

cheatsheets/Secrets_Management_Cheat_Sheet.md
sergiomarotco: @szh validation build "Markdown Link Checker" not find that mistake. Is it possible to write rule to find same in next time?

szh: We should. It's a valid link, just not the desired location. It will always link to Github, even on the published site. Using a relative URL allows it to work both on Github and on the published site.

szh: It's a super common mistake though and we really should automate it

kwwall:
> We should publish best practices, right? Using HashiCorp Vault as an example:
> 
> * FRONTWARE - NginX/Apache or API Management/Security software for balancing for first
> * MIDDLEWARE - Vault server;
> * BACKEND - Consul Server.

Well, except in that case, HashiCorp Consul is the component that's _optional_. That's especially true in lower risk and/or smaller operations. I've seen plenty of places where Vault is used alone, without Consul. Also, "best practices" generally implies more than simply choosing the most secure solution. It may not even be that the "most secure" solution is a "best practice" if a basic working assumption is "assume breach" as adding Consul expands the potential attack service and likely increases the operational security costs substantially.  Cost-benefit exercises is one side of the  risk analysis that you have to do to truly know if what you are intending to deploy is the _best_ solution. (If you can't tell, I'm one of those annoying people who don't believe that "best practices" is a one-size-fits-all solution.)

mackowski:
@sergiomarotco what is your take on @kwwall comment?

sergiomarotco:
Hello! 
I think, u all can add some fix into my code because u have rights. Or close as not interested.

Everyone I know agrees that hosting a web server (FRONTWARE) on a separate server is the best option for security reasone, not for economy reason. I think cheatsheets about security, and readers must have their own critical thinking.

My opinion may differ from the generally accepted, best or cheapest option.

Issue #1683, opened 2025-05-31
Update: Password_Storage_Cheat_Sheet.md


CyberskiZ:
## What is missing or needs to be updated?

It states that Password Peppers should be rotated: "Like any other cryptographic key, a pepper rotation strategy should be considered." This is not possible - the password hashes are all created using the original pepper, and the plaintext password isn't stored on the server and so the hash cannot be regenerated with a rotated pepper. 

## How should this be resolved?

Either that line should be removed, or further explanation should be added how this can be done.

jmanico:
Since a pepper is basically a symmetric key that encrypts the password hash, rotating this key is very straight forward. Why do you suggest this is not possible?

CyberskiZ:
Yes, it's very straightforward to rotate. The problem is that the new key can't replace the old one. All the password hashes were generated using the original pepper. The new pepper can be used for creating hashes for new passwords going forwards, but the original pepper will still be needed for all the password hashes that have been created and stored prior to the key rotation. (And of course we can't generate new hashes for the old passwords with the new pepper, as we don't store the plaintext password.)

CyberskiZ:
I reread that paragraph in the Cheat Sheet, and you are right - the cheat sheet defines a pepper as a key used to encrypt the hash, in which case of course it's simple to rotate the key.

However, this isn't the standard definition of a password pepper. Encrypting the password hashes would usually be referred to simply as encrypting the password database. Peppering is used to describe using a secret key as additional input to the hashing algorithm. The password+salt+pepper are put through the hashing algorithm to produce the hash that is stored. I believe this is the standard definition of a Password Pepper https://en.wikipedia.org/wiki/Pepper_(cryptography). For more sources see [here](https://www.baeldung.com/cs/password-salt-pepper), [here](https://nordpass.com/blog/pepper-password/) and [here](https://docs.moodle.org/500/en/Password_salting). 

As such, I would adjust my update request to change the way Pepper is defined in the Cheat Sheet.

jmanico:
Ah, those references are good. I always thought the pepper was separate from the hash. I see it's not. I am happy to get a PR from you any way you see it. I'm with you now.

mackowski:
I agree @CyberskiZ! Do you want to make a PR?

kwwall:
@jmanico wrote a while back:
> Since a pepper is basically a symmetric key that encrypts the password hash, rotating this key is very straight forward. Why do you suggest this is not possible?

[...]

PR #1730, opened 2025-07-10
Add peppering strategies to Password Storage Cheat Sheet (#1683)

arunjohnkuruvilla:
This PR fixes issue #1683.


cheatsheets/Password_Storage_Cheat_Sheet.md
```diff
@@ -58,14 +58,22 @@ Salting also protects against an attacker's pre-computing hashes using rainbow t
...
 
-For example, one peppering strategy is hashing the passwords as usual (using a password hashing algorithm) and then using an HMAC (e.g., HMAC-SHA256, HMAC-SHA512, depending on the desired output length) on the original password hash before storing the password hash in the database, with the pepper acting as the HMAC  key.
+#### Pre-hashing peppers
+
+In this strategy, a pepper is added to a password before being hashed by a password hashing algorithm. The computed hash is then stored in the database. In this case the pepper should be a random value generated securely. See the [Cryptographic_Storage_Cheat_Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#secure-random-number-generation) for more information on securely generating random values.
```
kwwall: I think I would change:
>  In this case the pepper should be a random value generated securely.

to

>  In this case the pepper should be a random value generated securely, but stored outside the database of password hashes.

cheatsheets/Password_Storage_Cheat_Sheet.md
```diff
@@ -58,14 +58,22 @@ Salting also protects against an attacker's pre-computing hashes using rainbow t
...
+
+#### Post-hashing peppers
+
+In this strategy the passwords hashed as usual using a password hashing algorithm. The resulting password hash is then hashed again using an HMAC (e.g., HMAC-SHA256, HMAC-SHA512, depending on the desired output length) before storing the resulting hash in the database. In this case the pepper is acting as the HMAC key and should be generated as per requirements of the HMAC algorithm.
```
kwwall: Again, maybe a mention not to store the `pepper` in the same DB. Alternately, you could note that earlier so you don't have to keep repeating it.

cheatsheets/Password_Storage_Cheat_Sheet.md
```diff
@@ -58,14 +58,22 @@ Salting also protects against an attacker's pre-computing hashes using rainbow t
...
+
+#### Common requirements for peppering strategies
```
kwwall: Hmm. Okay. So you have it here.  What do you think of moving this section closer to the top?

cheatsheets/Password_Storage_Cheat_Sheet.md
```diff
@@ -58,14 +58,22 @@ Salting also protects against an attacker's pre-computing hashes using rainbow t
...
 
 - The pepper is **shared between stored passwords**, rather than being *unique* like a salt.
-- Unlike a password salt, the pepper **should not be stored in the database**.
+- Unlike a password salt, the pepper should not be public and **should not be stored along with the generated hash**. The pepper should be stored separately from the password database.
 - Peppers are secrets and should be stored in "secrets vaults" or HSMs (Hardware Security Modules). See the [Secrets Management Cheat Sheet](Secrets_Management_Cheat_Sheet.md) for more information on securely storing secrets.
-- Like any other cryptographic key, a pepper rotation strategy should be considered.
+- In the event of a pepper value's compromise, the pepper value will have to be rotated. Peppers cannot be rotated without knowledge of a user's password. Therefore rotation of a pepper value will require forcing all users whose passwords were protected by the rotated-out pepper to reset their passwords.
```
kwwall: I prefer the term 'changed' over rotated. First, it is less technical and secondly, 'rotation' usually implies not only an action that you take upon learning that a secret has been compromised, but an operation that is done as a regularly scheduled event. And I'm pretty sure you don't want change the pepper yearly because of the impact it has to your user base.

kwwall (dismissed):
I like the changes in the PR. Take my comments as optional suggestions and not changes that I am demanding.

cheatsheets/Password_Storage_Cheat_Sheet.md
arunjohnkuruvilla: I have moved the **Common requirements for peppering strategies** prior to mentioning the two strategies.

cheatsheets/Password_Storage_Cheat_Sheet.md
arunjohnkuruvilla: I have moved the **Common requirements for peppering strategies** prior to mentioning the two strategies.

cheatsheets/Password_Storage_Cheat_Sheet.md
arunjohnkuruvilla: Great suggestion. I have moved this section to the top prior to mentioning the two pepper strategies.

cheatsheets/Password_Storage_Cheat_Sheet.md
arunjohnkuruvilla: Good point. I do not want to imply that changing the pepper should be a recurring event, and change of a pepper should only be done in the event of a compromise of the pepper.

PR #1731, opened 2025-07-17
References to SecureString and GuardedString removed from Secret Mgmt CS

hmmueller:
This PR fixes issue #`1704`.

### DIFF
--- a/cheatsheets/Secrets_Management_Cheat_Sheet.md
+++ b/cheatsheets/Secrets_Management_Cheat_Sheet.md
@@ -96,24 +96,15 @@ generally is much more difficult.
 - **Structures and Classes:** In .NET and Java, do not use immutable structures
     such as Strings to store secrets, since it is impossible to force them to
     be garbage collected. Instead use primitive types such as byte arrays or
-    char arrays, where the memory can be directly overwritten. You can also
-    use Java's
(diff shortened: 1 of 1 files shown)


cheatsheets/Secrets_Management_Cheat_Sheet.md
```diff
@@ -96,24 +96,15 @@ generally is much more difficult.
 - **Structures and Classes:** In .NET and Java, do not use immutable structures
```
kwwall, line 96: Not in this PR, since it had a specific purpose, but I wonder if we eventually we should remind developers that also avoid treating secretes with immutable types such as strings is not even possible without rewriting significant portions of the standard SDKs. (E.g.,the Jakarta Servlet API is a good example. All the parameter values are returned as Strings.) [Note: I'm not volunteering to do that PR, just asking thoughts if whether we should consider it.] Sometimes it keeping secrets confidential is the primary goal, maybe one should favor C/C++ over languages that are garbage collected. Sometimes you have to choose your poison.

kwwall (approved):
As long as we don't feel that we are going to be fielding questions from Cheat Sheet community users as to why we've removed references to SecureString and GuardedString, this is fine. (And, personally, that's what I'd recommend initially at least.) But we need to be prepared to maybe add a reason or two for this recommendation should we start to get questions.

szh (approved):
LGTM!

PR #1733, opened 2025-07-21
X-Frame-Options + browsers

drwetter:
* Remove deprecated X-Frame-Options
* Clarify use of headers

# You're A Rockstar

### DIFF
--- a/cheatsheets/REST_Security_Cheat_Sheet.md
+++ b/cheatsheets/REST_Security_Cheat_Sheet.md
@@ -133,9 +133,9 @@ Services including script code (e.g. JavaScript) in their responses must be espe
 
 ## Security Headers
 
-There are a number of [security related headers](https://owasp.org/www-project-secure-headers/) that can be returned in the HTTP responses to instruct browsers to act in specific ways. However, some of these headers are intended to be used with HTML responses, and as such may provide little or no security benefits on an API that does not return HTML.
+There are a number of [security related headers](https://owasp.org/www-project-secure-headers/) that can be returned in the HTTP responses to instruct browsers to act in specific ways. However, some of these headers are intended to be used with HTML responses, and as such may provide little or no security benefits on an API that does not return HTML and the client might not be a browser.
 
-The following headers should be included in all API responses:
+Given those constraints, the following headers can be included in all API responses:
 
 | Header | Rationale |
 |--------|-----------|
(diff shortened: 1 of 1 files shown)


szh:
Discussing in Slack. Main question is what the source is that the x-frame-options header is deprecated.

jmanico:
According to CanIUse X-Frame-Options support has been consistent (not complete, but consistent) for years.

https://caniuse.com/?search=x-frame-options

psiinon:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options gives no indications that its depreciated.
The `ALLOW-FROM origin` is obsolete, but not the whole header.

drwetter:
First of all, this is a REST API where the client is likely not a browser. I'd suggest to phrase that better in the cheat sheet. curl / java / php client etc do not care about those headers -- IMHO.


@jmanico : Caniuse says _The X-Frame-Options header has been obsoleted by [the frame-ancestors directive](https://www.w3.org/TR/CSP2/#directive-frame-ancestors) from Content Security Policy Level 2._ It's way more flexible, maybe deprecated was the wrong term. Given that and the fact we're talking about a REST API I believe it doesn't hurt to remove that line here.

jmanico:
> First of all, this is a REST API where the client is likely not a browser. I'd suggest to phrase that better in the cheat sheet. curl / java / php client etc do not care about those headers -- IMHO.

Good suggestion, clarifying that this is for web browsers is a good idea.

> @jmanico : Caniuse says _The X-Frame-Options header has been obsoleted by [the frame-ancestors directive](https://www.w3.org/TR/CSP2/#directive-frame-ancestors) from Content Security Policy Level 2._ It's way more flexible, maybe deprecated was the wrong term. Given that and the fact we're talking about a REST API I believe it doesn't hurt to remove that line here.

REST API's are frequently used for web clients. But this header is very legacy and we need to clarify the details as the OP suggests.

I do not want to delete this but I do suggest we change with these two considerations:

1. Change X-Frame-Options to X-Frame-Options AND/OR CSP frame-ancestors
2. Explain this is only needed if this API serves web clients and that is not necessary if it does not

Would these suggestions help satisfy your concerns, @drwetter ?

mackowski:
@drwetter what do you think about @jmanico suggestion?

mackowski:
@drwetter ping ;-)

drwetter:
Sorry for letting this hang. And thanks for your persistence!

My head is occupied with too much stuff. Like to get back latest on this in December during the first week.

andrzejsydor (approved):

mackowski:
@drwetter if you have time please review @jmanico comment

jmanico:
Closing this. I have addressed the concerns directly in b2052b84.

Changes made:
- Clarified that security headers are browser directives and may not apply to non-browser API clients (server-to-server, mobile apps, CLI tools)
- Noted X-Frame-Options as a legacy header superseded by CSP frame-ancestors, but kept for older browser compatibility
- Changed table intro to say headers should be included in responses consumed by browser clients

Thanks for raising this @drwetter. The points about REST APIs not always serving browsers and frame-ancestors superseding X-Frame-Options were valid.

PR #1752, opened 2025-08-13
added x-robots-tag header

r3dg33k:
# You're A Rockstar

### DIFF
--- a/cheatsheets/HTTP_Headers_Cheat_Sheet.md
+++ b/cheatsheets/HTTP_Headers_Cheat_Sheet.md
@@ -239,6 +239,20 @@ MvcHandler.DisableMvcResponseHeader = true;
 
 - *NOTE*: Remember that attackers have other means of fingerprinting your tech stack.
 
+### X-Robots-Tag
+
+The HTTP `X-Robots-Tag` response header helps control how search engines and AI bots index and show files like PDFs, images, and other non-webpage content. It works like the robots meta tag but is set in the server response, giving more control and flexibility.
+
+```lang-none
+X-Robots-Tag: noindex, nofollow
+```
+
+- *NOTE*: Only well-behaved crawlers follow these rules, and they must first access the resource to read its headers and meta tags.
+  
+#### Recommendation
+
+Disable sending this header. To remove the `X-AspNetMvc-Version` header, add the below line in `Global.asax` file.
+
 ### X-DNS-Prefetch-Control
 
 The `X-DNS-Prefetch-Control` HTTP response header controls DNS prefetching, a feature by which browsers proactively perform domain name resolution on both links that the user may choose to follow as well as URLs for items referenced by the document, including images, CSS, JavaScript, and so forth.


jmanico (dismissed):

cheatsheets/HTTP_Headers_Cheat_Sheet.md
```diff
@@ -239,6 +239,20 @@ MvcHandler.DisableMvcResponseHeader = true;
...
+
+```lang-none
+X-Robots-Tag: noindex, nofollow
+```
+
+- *NOTE*: Only well-behaved crawlers follow these rules, and they must first access the resource to read its headers and meta tags.
+  
+#### Recommendation
+
+Disable sending this header. To remove the `X-AspNetMvc-Version` header, add the below line in `Global.asax` file.
```
szh: Why?

r3dg33k: 1. The X-Robots-Tag header tells crawlers how to index a URL.
2. awareness

jmanico:
I'm not 100% sure on this one, @mackowski ?

cheatsheets/HTTP_Headers_Cheat_Sheet.md
szh: Can you provide a reference / source for this recommendation?

r3dg33k: https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag

mackowski: This recommendation is copied from X-AspNetMvc-Version header is not not valid for X-Robots-Tag right?

mackowski:
hey @righettod can you help with this PR? :)

cheatsheets/HTTP_Headers_Cheat_Sheet.md
jmanico: Here is a better reference.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Robots-Tag

jmanico:
I tend to be thumbs-up on this PR. It seems reasonable to want to AI crawlers of the day to leave us alone. :)

mackowski:
@jmanico the problem is that the recommendation is from different header
Screenshot 2025-09-08 at 12 34 56
Under X-Robots-Tag we will have:
'Disable sending this header. To remove the `X-AspNetMvc-Version` header, add the below line in `Global.asax` file.'

mackowski (changes requested):
Update the 'Recommendation' section

jmanico:
I'm with you now @mackowski

mackowski:
@jmanico @szh I updated this PR (with copilot) to fix previous issue

szh:
I'm still not 100% convinced we should outright recommend disallowing crawlers without caveats. There are downsides. For example, I would argue we definitely want AI to ingest the cheat sheets so they will be more likely to give good recommendations for people who will inevitably ask LLMs for security advice.

mackowski:
Agree, we should add this header but think how to describe the recommendation.
By the way - on the pages where I want to disable crawler it is better to implement solid AuthN/Z ;)

jmanico:
Overall we want this recommendation for any data we want to protect from crawlers. It's totally fair that public data does not need this.

mackowski:
@jmanico and @szh I hope this version is better ;-)

Issue #1751, opened 2025-08-12
Update: SAML Security Cheat Sheet


markgamache:
## What is missing or needs to be updated?

SAML trusts between IdPs and RPs rely on x509 certificates. These trusts are usually explicit, mapping a cert to an IdP, by the RP.  There has always been a bit of a debate as to whether it makes sense to use CA signed certificates or self-signed. Very few RPs bother to check typical certificate properties and simply treat the certificate as a container for a public key. Rarely is expiration checked, not chain of trust, nor certificate revocation or the 10 other things that matter in a certificate.  Rarely is their a debate about CA signed certificates coming from public CAs vs various private CAs. Private CAs have no clear governance, and are often a dumpster fire. If private CAs were to be trusted, that would probably necessitate a private PKI cheat sheet that is a mile long. 

There WAS no clear winner in the debate, until now. IMO.  The webPKI industry, driven by the CABF, is making changes that make public CA signed certificates risky to the point of dangerous.  Public CA certificates are on a path to only being good for [47 days](https://www.digicert.com/blog/tls-certificate-lifetimes-will-officially-reduce-to-47-days).  As the cert trust if often done manually, rather than by an RP consuming an IdP's key metadata URL, this means nearly constant change of certificates.  It is already painful for SAML cert rotation, as all RPs have to update their config, at the same time, when the IdP changes its certificate.

## How should this be resolved?

I am proposing that we officially state the self-signed are preferred. That public CA certificates are too risky for uptime. This is in line with the current OWASP guidance to no pin certificates. 

Also, suggesting that the use of private PKI is too risky and should be avoided. We could call out that webTrust (etc) audited private PKI so good. This is a high bar, but makes the point that private PKI is not inherently bad, just rarely executed on well. 

This would land us on using self signed certs. On this topic, we'd add some clear guidance on this, in terms of cert construct, max life key types and sizes and also call out strong private key protection, which is not in the current cheat sheet.

jmanico:
Thank you for this issue submission!

Thoughts:

**You’re recommending self-signed as preferred — but:**
- This can weaken trust bootstrapping if the exchange mechanism is insecure.
- You should outline the risks (e.g., MITM during initial trust establishment, stale keys, weak crypto) and how to mitigate them (e.g., out-of-band key exchange, FIPS 140-3 compliant storage).

**“Avoid private PKI” is too absolute**
[...]

PR #1754, opened 2025-08-15
changes for issue #1751

markgamache:
# You're A Rockstar

This PR fixes issue #1751.


jmanico (commented):
This is an AI review foe accuracy:

“Self-signed is the clear winner”

* This is mostly true in practice, but needs nuance:
* Self-signed certs are only “best” if metadata exchange is automated and the SP validates over authenticated TLS.
* If metadata distribution isn’t automated, self-signed increases operational risk (manual copy = chance of SP trusting the wrong cert).
* In short: Self-signed + metadata URL = best practice. Self-signed + manual distribution = risky.

Certificate lifetime guidance

* You recommend 2 years max. NIST SP 800-57 (Part 1, Rev. 5) actually allows up to 3 years for RSA 2048 in many contexts, but given revocation challenges, 2 years is a safer bound.
* If not HSM-protected, 1 year or less is arguably more appropriate.

CRL vs OCSP

* You only mention CRL. Many IdPs/SPs don’t check revocation at all.
* OCSP (RFC 6960) is more widely used than CRL in WebPKI, though often disabled in SAML deployments. Worth mentioning for completeness.

SAML XML signature vs certificate signature

* You note the difference but it could be emphasized:
* The cert’s signature algorithm (e.g., SHA-256 with RSA) protects the cert chain.
* The SAML XML signature algorithm (e.g., RSA-SHA256 on the <Assertion> element) is what actually protects the assertion.
* Many misunderstand this — worth stressing.

CABF ballot impact

* The 47-day lifetime ballot (SC081v3) only applies to TLS server certificates at present. It does not apply to SAML use cases directly, but would affect any IdP using a public CA for SAML certs. That’s why it’s impractical, not because CABF governs SAML.

Private CA audits

* Correct that WebTrust/ETSI cover public CAs. For private CAs, SOC 2 Type II or ISO 27001 style controls are more common. Strict WebTrust audit for an internal IdP CA is rare.

markgamache:
before I update the PR, some thoughts. @jmanico 

> This is an AI review foe accuracy:
> 
> “Self-signed is the clear winner”
> 
> * This is mostly true in practice, but needs nuance:
> * Self-signed certs are only “best” if metadata exchange is automated and the SP validates over authenticated TLS.
> * If metadata distribution isn’t automated, self-signed increases operational risk (manual copy = chance of SP trusting the wrong cert).
> * In short: Self-signed + metadata URL = best practice. Self-signed + manual distribution = risky.
I'd suggest that even if the import is manual, if the place to get metadata or cert is protected by TLS and some official site, it is still secure, but less available, and self signed or CA signed doesn't change the availability or integrity level.
> 
> Certificate lifetime guidance
> 

Agreed on all of this. will spell it out
> * You recommend 2 years max. NIST SP 800-57 (Part 1, Rev. 5) actually allows up to 3 years for RSA 2048 in many contexts, but given revocation challenges, 2 years is a safer bound.
> * If not HSM-protected, 1 year or less is arguably more appropriate.
> 
> CRL vs OCSP
> 
OCSP is kind of going away, due the privacy risk. Let's Encrypt, for instance, just abandoned it. I will call it out as a possibly and mention the risk and varying support. 
> * You only mention CRL. Many IdPs/SPs don’t check revocation at all.
> * OCSP (RFC 6960) is more widely used than CRL in WebPKI, though often disabled in SAML deployments. Worth mentioning for completeness.
> 
> SAML XML signature vs certificate signature
> 
will do. 
> * You note the difference but it could be emphasized:
> * The cert’s signature algorithm (e.g., SHA-256 with RSA) protects the cert chain.
> * The SAML XML signature algorithm (e.g., RSA-SHA256 on the  element) is what actually protects the assertion.
> * Many misunderstand this — worth stressing.
> 
> CABF ballot impact
> 
The key here is the term public CA is a bit vague, as many of the providers also have private CAs. That is using the definition of public as the common CAs at CCADB. The stores there, are all governed by CABF. as you point out, there is body governing SAML certs, but if you want a public cert, this is really the only choice. Maybe I call all of this out?  CABF/CCADB certs are the closest match to the term public, but they don't care about SAML?

> * The 47-day lifetime ballot (SC081v3) only applies to TLS server certificates at present. It does not apply to SAML use cases directly, but would affect any IdP using a public CA for SAML certs. That’s why it’s impractical, not because CABF governs SAML.
> 
> Private CA audits
> 
It looks like SOC 2 Type II is pretty solid, for CA, but ISO 27001 seems a bit too broad and vague.  I agree that most CAs aren't audited with something like WebTrust, and that is kind of my point. WebTrust/ETSI will do audits for private CAs, FWIW. That is the gold standard and what you need, if you really want to trust a 3rd party. I lean to adding SOC 2 Type II, but not ISO 27001. Thoughts? 

> * Correct that WebTrust/ETSI cover public CAs. For private CAs, SOC 2 Type II or ISO 27001 style controls are more common. Strict WebTrust audit for an internal IdP CA is rare.

markgamache:
@jmanico I think I got all of your thoughts in.

jmanico (approved):
Thank you very kindly for this work!

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -43,14 +43,14 @@ Vulnerabilities in SAML implementations due to XML Signature Wrapping attacks we
 
 The following recommendations were proposed in response ([Secure SAML validation to prevent XML signature wrapping attacks](https://arxiv.org/pdf/1401.7483v1.pdf)):
 
-- Always perform schema validation on the XML document prior to using it for any security-­related purposes:
+- Always perform schema validation on the XML document prior to using it for any security-related purposes:
```
kwwall: I''m not sure why this line is showing up in a diff as it looks like the before and after are identical to me, but I'm glad it popped up. I'm not sure it gets its point sufficiently gets it's point across.

I think the message that we want to get across here is you can't take short cuts and you need to do this schema validation each and every time, not just (say) once for each IdP / user identity. Yes, I know it says "Always" here, but there will also _always_ be some developers who will read right past that. Developers are prone to take shortcuts, especially it it will shave a few milliseconds off a response. I propose changing this to state:
```plaintext
Without exception, always perform schema validation on the XML document prior to using it for any security-related purposes:
```
or something similar. Am just being anal here? Perhaps. But additional cues not to read over important words never hurt, especially in this age of short attention spans.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -43,14 +43,14 @@ Vulnerabilities in SAML implementations due to XML Signature Wrapping attacks we
...
-    - If possible, inspect schemas and perform schema hardening, to disable possible wildcard ­type or relaxed processing statements.
+    - If possible, inspect schemas and perform schema hardening, to disable possible wildcard type or relaxed processing statements.
 - Securely validate the digital signature:
     - If you expect only one signing key, use `StaticKeySelector`. Obtain the key directly from the identity provider, store it in local file and ignore any `KeyInfo` elements in the document.
     - If you expect more than one signing key, use `X509KeySelector` (the JKS variant). Obtain these keys directly form the identity providers, store them in local JKS and ignore any `KeyInfo` elements in the document.
-    - If you expect a heterogeneous signed documents (many certificates from many identity providers, multi­level validation paths), implement full trust establishment model based on PKIX and trusted root certificates.
+    - If you expect a heterogeneous signed documents (many certificates from many identity providers, multilevel validation paths), implement full trust establishment model based on PKIX and trusted root certificates.
```
kwwall: Perhaps we should be clearer to tell them when to expect this in non-technical terms? We could state that this is what you would normally expect when you are federating identities _across_ companies, but if you're just using SAML for internal SSO, this is less likely to come up (unless you have multiple IdPs internally, which often happens after your company has been involved in M&A).

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
 
 ## Identity Provider and Service Provider Considerations
 
-The SAML protocol is rarely the vector of choice, though it's important to have cheatsheets to make sure that this is robust. The various endpoints are more targeted, so how the SAML token is generated and how it is consumed are both important in practice.
+The SAML protocol is rarely the attack vector of choice, though it's important to have cheatsheets to make sure that this is robust. The various endpoints are more targeted, so how the SAML token is generated and how it is consumed are both important in practice.
```
kwwall, line 108: There's a couple reasons _why_ SAML protocol is infrequently the attack vector of choice:

1. The frequently involves cryptographic attacks and cryptographic attacks are generally low on the totem pole because they're often noisy and will show up on SIEM alerts.
2. Often they are only exposed internally or from certain endpoints. For instance, when my AppSec team was depingloy a federated identity management product, it usually was B2B and we didn't open the SAML endpoints to the public (because of DoS concerns; verifying the digital signatures in SAML assertions is CPU intensive, but it's pretty trivial to forge fake ones). Instead we usually required the endpoints be accessed over a VPN.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+### X.509 Certificate Considerations
+
+Typically the security association between the Identity Provider (IdP) and Service Provider (SP) is created when the SP explicitly chooses to trust the IdP's X.509 signing certificate. Exactly how this occurs can have a strong bearing on overall security posture. How the certificate is generated, what the contents of the certificate are, and how the certificate's corresponding private key is protected all have strong bearing on security posture. e.g. if an attacker has access to use the IdP’s signing key, they can mint SAML responses containing any assertion they wish.
```
kwwall: I couple of suggested changes here:

- s/what the contents of the certificate are/what  certificate extensions should be present and which of those should be considered _critical_ extensions, what the contents of the certificate are/
- s/e.g. if an attacker/E.g., if an attacker/

Also, somewhere, I would note that these things should be ironed out in terms of the IdP / SP contract negotiations so that they are legally binding, rather just punting the technical details to the InfoSec nerds at a later TBD time.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+In many cases, the method of manually setting the association is akin to [Certificate Pinning](https://cheatsheetseries.owasp.org/cheatsheets/Pinning_Cheat_Sheet.html), which is not ideal. Depending on the IdP and SP software, or various design considerations, this may be unavoidable.
+
+Keep in mind that the certificate's key type and signing can be different than that of the XML document signing. The certificate's key is the only key that can be use to sign the XML document, but the signing algorithm is chosen at the IdP's discretion. The most commonly supported signing algorithm is SHA-256.
```
kwwall: Okay. I know you know better than this Mark, so I'm going to say you intentionally put this in as a honey trap to see if I was awake.

"The certificate's key" is the _public_ key. We do not _sign_ with the public key. We sign with the _private_ key. The public key in the certificate is used for verifying the signature.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+#### Certificate Contents
+
+In the context of SAML signing and encryption, X.509 certificates are most often treated simply as a wrapper to hold a public key that is used to verify a signature, or to wrap a symmetric key for SAML encryption. Nonetheless, a certificate can contain attributes that can be used to further enhance security.
```
kwwall, line 144: "... a public key that is used to verify a signature" :+1: 

See. I knew that you knew. :grin:

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+##### Keys and Signing Algorithms
+
+The key pair size and type and signing algorithm choice has strong bearing on security and interoperability.  Not all IdP and SP software packages or libraries support all combinations of options. As one IdP often has many SPs associated with it, the only option is to pick the least secure option that is still considered secure. The most supported and currently secure combination is using RSA 2048 bit keys and SHA-256 signing. This is referring to the certificate’s signing algorithm and not the SAML XML signing.
```
kwwall: Maybe we should also note that this situation is only likely to get worse as post-quantum cryptography gains more traction and starts getting rolled out, but at different rates by different companies.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+At least one major vendor [Microsoft Entra](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/how-to-connect-fed-saml-idp) doesn’t support ECC keys.
+
+###### Signing Algorithms
+
+No IdP should use a certificate signed with SHA-1. SHA-256 is the minimum bar. That said, if possible moving to larger hash algorithms like SHA-384 or SHA-512 means you are better future-proofing your service. In this context we are talking about the certificate’s signing algorithm and not the one used to sign the SAML response XML.
+
+##### Certificate Lifetime
+
+Certificates contain a NotBefore and NotOnorAfter attribute. Most IdPs ignore these in favor of guaranteeing uptime if certificate rotation does not happen on time. The SAML certificate lifetime should be handled well enough that ignoring these is not needed. While [NIST SP 800-57 (Part 1, Rev. 5)](https://csrc.nist.gov/pubs/sp/800/57/pt1/r5/final) allows RSA 2048 bit keys to last for 3 years, the maximum lifetime of a SAML signing certificate should be two years. If the private key is not well protected, such as in a Hardware Security Module (HSM), that may be too long to be safe.
```
kwwall: Regarding the statement:
```plaintext
Most IdPs ignore these in favor of guaranteeing uptime if certificate rotation does not happen on time.
```

Just because they do, doesn't mean they should. One ought to plan in advance and be aware when certs are about to expire. The NotBefore and NotOnorAfter attribute for the new certificate can be adjusted to allow some overlap with the current one to allow sufficient time for roll-out.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+##### Extended Key Usage (EKU) and Key Usage (KU)
+
+[EKU](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.12) describes a specific use case that the certificate is intended for. These are use cases like server authentication, client authentication, and code signing. There is no widely accepted EKU for SAML signing, but [RFC 9336](https://www.rfc-editor.org/rfc/rfc9336.txt) defines one that is ideal, id-kp-documentSigning (1.3.6.1.5.5.7.3.36). IdPs and SPs may consider standardizing on this EKU.
```
kwwall: I'd say that code-signing should never be used for SAML. SAML Assertions are not code.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+[KU](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.3) describes the underlying cryptographic operations that the private key is meant for. There are things like digitalSignature, nonRepudiation, keyEncipherment, etc. IdPs and SPs may consider requiring digitalSignature and further, disallowing certificates that have other KUs, as certificates should only be used for one use case. e.g. The IdP's TLS server certificate must never be the SAML signing certificate.
```
kwwall: s/e.g. The IdP's/E.g., the IdP's/

And amen to that last statement!

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+##### Online Certificate Status Protocol (OCSP)
+
+[OCSP](https://datatracker.ietf.org/doc/html/rfc6960) is another way of checking to see if a certificate  is revoked. The OCSP URL is embedded in the certificate , like a CRL, and should be reacable  over HTTP. The response is signed, so MITM attacks are not an integrity concern. OCSP is becoming less favored, as the exchange creates privacy concerns. The caller's IP address can be seen and the certificate that is being used is disclosed. This is less of a concern for SAML, as this does not disclose a destination website, use overall has declined. If an OCSP URL is present on any certificate in the chain, it should be used to check if the certificate is revoked.
```
kwwall: CRLs and OCSP are intended to addressed a possibly compromised private key, but in the content when there are lots of clients (specifically client browsers) that are using some TLS endpoint.

I'd argue that the SAML certificate that the IdP uses to sign the SAML assertion likely has several of orders of magnitude less clients and further more the IdP should know who they all are, namely the SPs that they interact with. Therefore in the event of a compromise, it would be much quicker for an IdP to just notify the SPs directly and make them aware of the situation. That is, I don't think that either CRLs or OCSP is crucial for SAML certificates. Can it work? Perhaps, but only if CRLs or OCSP URLs are checked when the Assertion is verified, and I doubt there are a lot of SAML products or libraries that do that.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+#### Certificate Hierarchy
+
+The SAML signing certificate can be signed by one of three things. The certificate can be self-signed, public CA signed, or private CA signed. Each has pros and cons, however, given the current state of WebPKI (public) and private PKI, using self-signed SAML certificates are the clear winner when proper precatuions are taken for exchanging the certificates.
+
+##### Certificate Issuer
+
+All X.509 certificates are signed, using a private key, by an authority known as an [Issuer](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.4). This may be a CA or in the case of a self-signed certificate the certificate's private key. In the case of a CA signed certificate, the signer may also have a certificate that has an Issuer, and so on. This is called chain, or path, and should terminate in a Root CA (which is self-signed by definition). The issuer should be inspected. If the issuer is a CA, its attributes, such as EKU, KU, and CRLs, may also be validated. This should happen for each certificate in the [path](https://datatracker.ietf.org/doc/html/rfc5280#section-3.2) all the way to the root.
```
kwwall: I think the term "certificate's private key" is a bit misleading. Technically it is the associated private key that corresponds to the public key embodied by the certificate.  I'm not trying to nitpick here, but because terms like this have been broadly tossed around in the past, there are many developers whom I've ran across (I know because I spent 6 years teaching this in a master's level CS class on Computer Security) who think that certificates contain private rather than public keys. And while both often are kept together in the same keystore, I think we do a disservice to future generations if OWASP continues to reinforce this confusion by imprecise wording.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+##### Private CA Signed
+
+As most IdPs and SPs treat the X.509 certificates as an explicit trust, private CAs and PKI could be used. How private CAs are designed, built, and run varies wildly and ultimately running CAs well is very costly. In order to trust a third-party's CAs, one would need to clearly understand the lifecycle of the CA. There are two audit types that would cover this, both of which are very costly, on top of building and running the CAs. If you rely on third-party CAs, they should be [WebTrust](http://www.webtrust.org/), [ETSI](http://www.etsi.org/technologies-clusters/technologies/security/certification-authorities-and-other-certification-service-providers), or [SOC 2 Type II](https://www.aicpa-cima.com/topic/audit-assurance/audit-and-assurance-greater-than-soc-2) audited.
+
+Trusting third-party CAs, if done improperly, could result in unintended over trust, for things such as TLS and code signing. If you choose to trust third-party CAs, make sure they are only trusted for the process of IdP signature validation.
```
kwwall: Probably should be "unintended over-trust" rather than "unintended over trust". When I first read it, I thought "what the heck does he mean by putting 'unintended' at more importance than then 'trust'.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+If third-party CAs are used they still should not issue SAML signing certificates where the lifetime of the certificate exceeds that of the underlying key pair, based on guidance from a standards organization such as [NIST, NSA, etc.](https://www.keylength.com/en/). If using the strongest private key types, this puts the upper limit at two years.
```
kwwall, line 209: Almost all of this section seems to imply that a "Private CA" equates to a "third-party CA", but I've seldom found this to the case. Usually it is an internal CA and many companies already have this set up for TLS certificates for their own company intranet websites. (And trust me, they have no intention to 47-day validity period endorsed by CABF.)

I think if a company already has this internal CA infrastructure set up (including DevSecOps processes for it's care and feeding), its reasonable to tweak it a bit and consider it for SAML signing certs. Then if you wish, I think that more easily allows you to issue separate certs for separate SPs.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+##### Self-Signed
+
+Due to the explicit nature of most SAML security associations, self-signed certificates are ideal for the use case. The contents of the certificate and lifetime are not constrained by the policy or process of the issuing CA, be it public or private. As rotating SAML certificates can be painful and labor intensive, setting the certificate lifetime as long as safely possible is key. Few CAs allow long enough lifetimes, due to their focus on the TLS threat model.
```
kwwall, line 213: Internal CAs don't have the "long enough lifetime issue". However, I think that if you don't have the current internal private CA infrastructure in place, this makes good sense as long as you don't insist on each SP getting a different self-signed cert. Because that could grow chaotically if you end up dealing with more than a handful of SPs.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+
+###### Creating a Self-Signed SAML Certificate
+
+If you are using a Hardware Security Module (HSM), follow the vendor's instructions. This process uses openssl.
+
+1. Generate a Private Key:
+openssl genrsa -out private.key 2048
```
kwwall, line 220: Um, earlier you weer endorsing ECC over RSA, so why not here?

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+x509_extensions = v3_ca
+prompt = no
+
+\[req_distinguished_name\]
+C = US
+ST = California
+L = San Francisco
+O = MyOrganization
+OU = MyUnit
+CN = SAML Signing
```
kwwall, line 237: Some companies, especially during transition, might be moving off FOSS (e.g., Shibboleth) to a COTS solution (e.g., Okta, Ping Identity, etc.), so in those cases there probably should be separate certs and thus different (i.e., more specific) CNs.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -106,14 +106,128 @@ Unsolicited Response is inherently [less secure](https://www.identityserver.com/
...
+basicConstraints = CA:FALSE
+keyUsage = digitalSignature
+extendedKeyUsage = 1.3.6.1.5.5.7.3.36
+
+3. Generate the Self-Signed Certificate:
+openssl req -x509 -new -nodes -key private.key -sha256 -days 365 -out certificate.crt -config cert.cnf -extensions v3_ca
+
+#### Certificate Metadata URLs
+
+Many IdPs publish a metadata URL that contains basic configuration information including the SAML signing certificate. Many SPs can consume the data from the IdP, updating the Signing certificate information in near real-time. Using these options is ideal. This model matches exactly the intent of the [Certificate and Public Key Pinning](https://owasp.org/www-community/controls/Certificate_and_Public_Key_Pinning) when pinning must be used.
```
kwwall, line 249: :+1:

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -125,12 +239,12 @@ The SAML protocol is rarely the vector of choice, though it's important to have
 - Level of granularity in setting authorization context when consuming SAML token (do you use groups, roles, attributes)
 - Ensure each Assertion or the entire Response element is signed
 - [Validate Signatures](#validate-signatures)
-- Validate if signed by authorized IDP
-- Validate IDP certificates for expiration and revocation against CRL/OCSP
+- Validate if signed by an authorized IDP
+- Validate IDP certificates for revocation against CRL/OCSP if they are present
```
kwwall, line 276: s/IDP/IdP/

kwwall (changes requested):
@markgamache - This is awesome content. I only found one serious glitch, but humbly offer up the other suggestions as well.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: At this time, unless you know your SPs and their capabilities, or can demand they meet a min bar, RSA 2048 is the most interoperable while still secure. I'd say we endorse either but we get the biggest bang for the buck on this.

markgamache: adding an OR and syntax for ecc as well.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: Not sure I follow. Each SP does have it's own cert as the are separate entities. FWIW, this doc does not fully cover all the certs and exactly how they are used. I clearly should but in a section on that. RPs use their private key and cert to sign the SAML req going into the IdP. SAML enc certs can be used to enc so that the SAML reqs and resps can't be read, even "in the browser"

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: oh man. I forgot the audience on this and lost my mind. Fixing.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: I added bit more harsh language around it.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: updated words

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: I am going to add a bit about this.  A couple projects/vendors do now have the ability to check revocation. I haven't had the time to tell if they do CRL, OCSP, or both.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: I love the attention to detail here. Hopefully you like my fix.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: hmm, and wow. I realize all the language and ideas I am using here assume that the SP to IdP relationship is that of B2B, so one's private PKI is the other's "third-party".  I will call this out in a new section. If it is all your company's PKI , then the threat model and risk acceptance may change.

markgamache: I have added a section on that. It editing each section to cover int SSO vs B2B would be messy. let me know what you think

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: there is nothing in that space for the users of SAML to know or care about yet, but I am going to put in something for the folks writing the libs/tools.

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: I have been using should vs must carefully, as one of the reasons I am doing the PR is that some organizations refuse to look at the intent of the guidance and only the letter. I guess the flip side is, it's hard to use the guidance to force change when it is a should an not a must. Is there some middle ground, like "OWASP strongly suggests"  :-P

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: Are you looking for me to call this out some, or is this just for the PR, in case anyone ever researches the decisions on the wording?

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: I didn't edit this at all, I just removed a hidden char. After reading it a few times, I have removed the line, as it is confusing and covered in better depth below.

markgamache:
@kwwall  @jmanico  updates per kwwall

jmanico (dismissed):
There is still one issue pending, but I will approve this so this goes live when Kevin is ready to approve of it as well

markgamache:
@jmanico  thx. that was my intention. I want to make sure Kevin loves the edit.

cheatsheets/SAML_Security_Cheat_Sheet.md
kwwall: That's up to you decide (since to this point, you touched it last), but I mostly put it there for the latter reason.

cheatsheets/SAML_Security_Cheat_Sheet.md
kwwall: Doesn't IETF use "recommends" for that purpose? So "OWASP recommends ..." would work IMO.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -135,29 +160,31 @@ No IdP should use a certificate signed with SHA-1. SHA-256 is the minimum bar. T
...
+Certificates contain a NotBefore and NotOnorAfter attribute. Most IdPs ignore these in favor of guaranteeing uptime if certificate rotation does not happen on time. The SAML certificate lifetime should be handled well enough that ignoring these is not needed. Ignoring the certificate's validity period is fundamentally a bad idea. a  While [NIST SP 800-57 (Part 1, Rev. 5)](https://csrc.nist.gov/pubs/sp/800/57/pt1/r5/final) allows RSA 2048 bit keys to last for 3 years, the maximum lifetime of a SAML signing certificate should be two years. If the private key is not well protected, such as in a Hardware Security Module (HSM), that may be too long to be safe.
```
kwwall: s/a bad idea. a  While/a bad idea. While/

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -135,29 +160,31 @@ No IdP should use a certificate signed with SHA-1. SHA-256 is the minimum bar. T
...
 [Certificate Revocation List (CRL) Distribution Point](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.13) are a list of certificates that the CA says should no longer be trusted. They are most often delivered over HTTP and the CRL URLs are generally embedded in each CA issued certificate. The [CRL](https://www.rfc-editor.org/rfc/rfc5280#appendix-C.4) is signed by the CA, so a man-in-the-middle attack against the HTTP cannot harm the integrity of the list, other than to tamper with, and thus invalidate, it. That is, it can't be altered by an attacker. Only CA signed certificates can have a CRL. If the SAML certificate has a CRL listed, it should be reachable by the validating party and the party should [validate](https://www.rfc-editor.org/rfc/rfc5280#section-6.3) it.
 
+Considering the level or risk, if a private key is compromised, and the smaller scale of IdP to SP relationships, parties in a SAML system should establish a plan with contact lists to notify and rotate certificates rapidly in case of an incident. Many SAML products and libraries don’t support revocation checking, and simply revoking the certificate, without coordinated replacement, means there is an outage.
```
kwwall: Probably meant "level **of** risk" here. The rest all looks good.

cheatsheets/SAML_Security_Cheat_Sheet.md
kwwall: I think this is good enough for now. If it turns out that readers have questions about it or want to see more details about SAML for internal SSO use, we can cross that bridge when we come to it. But let's not let the perfect become the enemy of the good here.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -110,11 +109,35 @@ The SAML protocol is rarely the attack vector of choice, though it's important t
...
 
 In many cases, the method of manually setting the association is akin to [Certificate Pinning](https://cheatsheetseries.owasp.org/cheatsheets/Pinning_Cheat_Sheet.html), which is not ideal. Depending on the IdP and SP software, or various design considerations, this may be unavoidable.
 
-Keep in mind that the certificate's key type and signing can be different than that of the XML document signing. The certificate's key is the only key that can be use to sign the XML document, but the signing algorithm is chosen at the IdP's discretion. The most commonly supported signing algorithm is SHA-256.
+Keep in mind that the certificate's signature type can be different from that of the XML document signing type. The certificate's corresponding private key is the only key that can be used to sign the XML document, but the signing algorithm is chosen at the IdP's discretion. The most commonly supported signing algorithm is SHA-256.
```
kwwall: > The most commonly supported signing algorithm is SHA-256.

SHA-256 is not a _signing_ algorithm, it's a message digest! .(If we were discussing HMAC-SHA-256 here, I might forgive someone for referring to it as a signing algorithm as it conceptually would be a little closer, but still not precise.) And while message digests are almost always involved in digital signatures (messages are first hashed, then the hash is signed, but that's done for efficiency, not necessity), calling them a signing algorithm is totally wrong.

I'm guessing someone just made a brain-fart here and probably meant RSA, not SHA-256.

cheatsheets/SAML_Security_Cheat_Sheet.md
```diff
@@ -110,11 +109,35 @@ The SAML protocol is rarely the attack vector of choice, though it's important t
...
+
+This is the certificate that an SP uses to validate an IdD’s SAML response. The IdP signs that response with the certificate’s corresponding private key. This is often the most important certificate and private key, as this protects the identity assertions being sent to the SP.
+
+##### IdP SAML Encryption
+
+This is less commonly used, as this is used when the SP wants to protect the SAML request sent to the IdP, not just from tampering, but from information disclosure. There should be no sensitive information in the SAML request, so it is less commonly used. If used, the certificate and private key must be different from that of the SAML signing certificate.
+
+##### SP Signing
+
+It is a best practice, though not required, that SPs and IdPs not allow IdP Initiated SSO. This means that the caller starts their SAML flow at the SP, which produces and signed SAML request intended for the IdP.
```
kwwall: s/produces and signed/produces a signed/

cheatsheets/SAML_Security_Cheat_Sheet.md
markgamache: added the langue in 'Certificate Contents'

kwwall (approved):
I'm good with the Mark's latest changes. If Mark learned one thing from this is that that Kevin Wall guy is a nitpicky PITA! :stuck_out_tongue_winking_eye:

jmanico (approved):

markgamache:
> I'm good with the Mark's latest changes. If Mark learned one thing from this is that that Kevin Wall guy is a nitpicky PITA! 😜

I can't tell you how much I appreciate the attention to detail. . Your rock!

jmanico:
> If Mark learned one thing from this is that that Kevin Wall guy is a nitpicky PITA! 😜

Kevin - that's your best characteristic! This is what we love around here! Keep it up! :)

Thank you Mark! Very well done :)

PR #1765, opened 2025-08-26
Draft: Security Definitions Cheat Sheet (Encoding vs Escaping, Encryption vs Signature).

Prasad-JB:
This PR introduces the initial draft of the Security Definitions Cheat Sheet.

Included entries:
- Encoding vs Escaping.
- Encryption vs Signature.

Each entry goes beyond just definitions and provides:
- Authoritative references (OWASP, CNCF, NIST, etc.)
- Why the term matters in security.
- Common pitfalls and real-world vulnerabilities.
- Best practices developers should follow.
- Example snippets for clarity.

The goal of this draft is to validate the structure and content style before expanding to additional terms. Feedback on the format and level of detail is welcome.

### DIFF
--- a/cheatsheets/Security_Definitions_CheatSheet.md
+++ b/cheatsheets/Security_Definitions_CheatSheet.md
@@ -1,0 +1,41 @@
+# Security Definitions Cheat Sheet (Draft)
+
+This cheat sheet provides enriched definitions of common security terminology.  
+Each entry includes authoritative references, context on why it matters, real-world pitfalls, and best practices.  
+
+---
+
+## Encoding vs Escaping
+
+**Definition (trusted sources):**  
+- [OWASP: Data Validation](https://owasp.org/www-community/data_validation)  
+- [CNCF Glossary: Encoding](https://glossary.cncf.io/)  
+
+**Why this matters:**  
+Developers often mix up encoding and escaping, but they are not the same thing.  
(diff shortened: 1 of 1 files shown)


mackowski:
This PR is creating 3 files instead of one and is not following Contributing Guide

Prasad-JB:
Thanks for the feedback. I understand that the PR currently creates multiple files instead of one, and I’ll align it with the contributing guide. The goal of this draft was to establish a clear format for security term definitions, with references, real-world context, and examples. I’ll consolidate the content into a single file and make sure it follows the repo standards.
and I’ll update the PR shortly

jmanico (commented):
I'm really confused why we have three files in this commit.

And I strongly disagree with escaping vs encoding. They are used for the same defensive purpose and giving them separate definitions is confusing. I would rather lump these two definitions together.

The rest seems decent work.

But please, smaller PR's are appreciated, we want to analize your many changes in smaller chunks.

And I do appreciate your enthusiasm for this project!

Prasad-JB:
Thanks for the feedback, I understand the concern about multiple files and the overlapping definitions. My intention was to provide clear distinctions for learning purposes, but I see how combining encoding and escaping makes more sense from a defensive perspective. I’ll work on smaller, consolidated PRs in the future and align with the contributing guidelines. Appreciate the guidance and the opportunity to contribute

PR #1766, opened 2025-08-26
Fix: Add secure vs insecure Java LDAP query example.

Prasad-JB:
This PR adds an example in the LDAP Injection Prevention Cheat Sheet:

- ❌  Insecure example: constructing an LDAP filter by string concatenation (vulnerable to injection).  
- ✅  Secure example: using parameterized filters with placeholders to safely include user input.

This helps readers clearly understand the difference between unsafe and safe LDAP query construction in Java.


jmanico:
This PR includes 4 files, including the new definitions. 

Can you give us the changes to the LDAP cheatsheet to be just that one file?

Prasad-JB:
Thanks for catching that, I’ll split this up so the LDAP Injection Prevention Cheat Sheet PR only contains the secure vs insecure Java example. I’ll move the Security Definitions content into its own PR so the scope stays clean.

Prasad-JB:
Closing this PR since it included unrelated Security Definitions changes. Please see the new clean PR with only the LDAP example.

Issue #1741, opened 2025-08-05
New CS proposal: Security Definitions


jmanico:
I find that even senior developers struggle with some security terminology. Encoding vs escaping vs serialization? Signatures vs encryption? Etc.

I suggest we build a cheatsheet that goes over security terminology that is critical to understanding other projects like the ASVS standard. It assumes developers have a basic security understanding.

mackowski:
Love it!

szh:
I think this is a great idea as well! Just to double check, are there any other OWASP projects that already have something like this? I know theres the [CNCF Glossary](https://glossary.cncf.io/) which may be a good reference for this effort.

Prasad-JB:
Hi @jmanico,

This cheatsheet idea sounds great — I can help compile and organize the glossary, referencing OWASP and CNCF sources to keep it consistent. Could you please assign me this issue so I can start drafting?

Thanks!

jmanico:
> This cheatsheet idea sounds great — I can help compile and organize the glossary, referencing OWASP and CNCF sources to keep it consistent. Could you please assign me this issue so I can start drafting?

Done!

ajayojha:
@jmanico I have some suggestions.

> I find that even senior developers struggle with some security terminology.

I agree. I think that providing only simple terminology definitions may not fully solve the problem. Developers can already get basic definitions from sources like ChatGPT by asking ex. “**what is encoding**”.,  But by asking a simple question they may not get below information:
I believe this below information would make this cheatsheet uniquely valuable is **adding context around each term**:

* Why the terminology is crucial in security.
* How misunderstandings or incorrect usage can lead to vulnerabilities.
* Real-world examples or common vulnerabilities reported due to misuse.
* Where developers should focus in their projects to apply the concept correctly.
* Practical use cases and best practices.

This is just a quick thought, if this is not align with your expectation then please ignore :).

kwwall:
> I think this is a great idea as well! Just to double check, are there any other OWASP projects that already have something like this? I know theres the [CNCF Glossary](https://glossary.cncf.io/) which may be a good reference for this effort.

Not sure if there are other OWASP projects doing this, but speaking from experience (I attempted to do this first for a security class I taught and then later with an AppSec team at a former company until it became a maintenance headache), there are other good public security glossaries out there. I know that NSA has one and I think NIST does as well. And IEEE has one, but it's not limited to just security terminology.

[...]

PR #1770, opened 2025-08-28
Add Encoding vs Escaping entry to Security Definitions Cheat Sheet

Prasad-JB:
This PR introduces the first draft entry ("Encoding vs Escaping") for the new Security Definitions Cheat Sheet (#1741).

Format includes:
- Baseline definition with external references
- Why it matters in security
- Real-world example
- Best practices
- Small code snippet

Once structure is validated, I’ll add more terms (e.g., Encryption vs Signature) in separate PRs as discussed.

### DIFF
--- a/cheatsheets/Security_Definitions_Cheat_Sheet.md
+++ b/cheatsheets/Security_Definitions_Cheat_Sheet.md
@@ -1,29 +1,35 @@
 # Security Definitions Cheat Sheet
 
-## Encoding vs Escaping
+## Encoding vs Escaping (and Sanitization)
 
 **Baseline Definition:**  
-- **Encoding**: Transforming data into another format using a publicly available scheme (e.g., Base64, URL encoding).  
-- **Escaping**: Adding special characters to data so that it is treated as plain text, not as executable code (e.g., HTML entity escaping).  
-- Sources: [OWASP](https://owasp.org), [CNCF Glossary](https://glossary.cncf.io)
+- **Encoding (Output Encoding):** Converting data into a safe representation for a specific output context (e.g., `<` → `&lt;` in HTML). Often referred to as *output encoding* in OWASP materials.  
+- **Escaping:** Adding a special character before a control character so it is treated as data, not code (e.g., escaping quotes in SQL or JSON). In practice, escaping and output encoding are sometimes used interchangeably.  
+- **Sanitization:** Modifying or stripping unsafe input to enforce a security policy (e.g., removing `<script>` tags).  
 
 **Why it Matters in Security:**  
-- Misunderstanding these terms can cause developers to use the wrong defense mechanism.  
-- Encoding does **not** protect against injection attacks, while escaping is often critical for preventing XSS.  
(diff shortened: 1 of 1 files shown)


jmanico (commented):
So far so good - thank you!

cheatsheets/Security_Definitions_Cheat_Sheet.md
```diff
@@ -0,0 +1,29 @@
...
+## Encoding vs Escaping
+
+**Baseline Definition:**  
+- **Encoding**: Transforming data into another format using a publicly available scheme (e.g., Base64, URL encoding).  
+- **Escaping**: Adding special characters to data so that it is treated as plain text, not as executable code (e.g., HTML entity escaping).  
+- Sources: [OWASP](https://owasp.org), [CNCF Glossary](https://glossary.cncf.io)
+
+**Why it Matters in Security:**  
+- Misunderstanding these terms can cause developers to use the wrong defense mechanism.  
+- Encoding does **not** protect against injection attacks, while escaping is often critical for preventing XSS.
```
kwwall: While generally true, this is likely to send a lot of developers into a panic because we commonly refer to these as `Encoders` not `Escapers`. E.g., https://javadoc.io/static/org.owasp.encoder/encoder/1.3.1/org/owasp/encoder/Encode.html and https://javadoc.io/static/org.owasp.esapi/esapi/2.7.0.0/org/owasp/esapi/Encoder.html. And while one could argue that perhaps both of these classes are misnamed, neither of them are going to get renamed at this point.

Not only that, we have a lot of other places where we probably refer to these as "output encoding" not "output escaping" and I'm not convinced this is a hill I'd want to die on, The old terminology is too far ingrained.

cheatsheets/Security_Definitions_Cheat_Sheet.md
```diff
@@ -0,0 +1,29 @@
+# Security Definitions Cheat Sheet
+
+## Encoding vs Escaping
```
kwwall: I almost think we should throw in "sanitization" into this grouping as well. Thoughts?

Prasad-JB:
Thanks for the thoughtful feedback 

Terminology (Encoding vs Escaping): I agree that the industry and OWASP docs frequently use "output encoding" instead of "output escaping". My initial intent was to draw a clear distinction for newcomers, since "encoding" alone is often misunderstood as a security control. I’ll adjust the wording to acknowledge both terms ("output encoding" a.k.a. escaping in certain contexts) so we don’t cause confusion.

Adding Sanitization: That’s a great suggestion. It often comes up in the same discussions, so including it alongside Encoding vs Escaping could help readers understand how the three fit together. I can add a subsection or a follow-up PR focused on “Sanitization vs Encoding vs Escaping.”

I’ll make these adjustments so the definitions are consistent with existing OWASP usage while still clarifying the security implications for developers.

mackowski:
@Prasad-JB you created couple PRs for 'Security Definitions CS' - please create one but well written

PR #1771, opened 2025-08-28
Fix: Add secure vs insecure Java LDAP query example

Prasad-JB:
This PR updates the LDAP Injection Prevention Cheat Sheet by adding a clear example:

❌ Insecure: constructing LDAP filters by string concatenation (vulnerable to injection).
✅ Secure: using parameterized filters with placeholders to safely include user input.

This makes the cheat sheet easier for developers to understand safe vs unsafe practices in Java.


Prasad-JB:
This PR is now scoped only to the LDAP Injection Prevention Cheat Sheet with the secure vs insecure Java example, as requested. Thanks for the guidance!

jmanico (approved):

mackowski (approved):

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -105,6 +105,25 @@ occur before the validation step (source: [Prevent LDAP injection](https://wiki.
 
 For further information visit [OWASP ESAPI Java Encoder Project which includes encodeForLDAP(String) and encodeForDN(String)](https://owasp.org/www-project-java-encoder/).
```
kwwall, line 106: Just noticed this from the context.  OWASP ESAPI is a different library than the OWASP Java Encoder Project. The 2 methods that are mentioned are only in ESAPI. Instead, this should say:

> For further information visit [OWASP ESAPI](https://owasp.org/www-project-enterprise-security-api/) which includes [encodeForLDAP(String)](https://javadoc.io/static/org.owasp.esapi/esapi/2.7.0.0/org/owasp/esapi/Encoder.html#encodeForLDAP-java.lang.String-) and [encodeForDN(String)](https://javadoc.io/static/org.owasp.esapi/esapi/2.7.0.0/org/owasp/esapi/Encoder.html#encodeForDN-java.lang.String-)

kwwall (commented):
Your current changes look good, but I noticed a mistake in the immediate context before your changes. So, if you would be so kind as to fix this while you are at it, I would appreciate it greatly. Thanks.

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
kwwall: Ah; never mind. Didn't notice this was already merged. Getting ready to hit the pillow. @mackowski - If I remember I'll create a issue and maybe a PR, depending on if I can find the time.

Prasad-JB:
Thanks for catching that, @kwwall You’re right. ESAPI and Java Encoder are separate projects. I can open a follow-up PR to correct the reference if you’d like, unless you’re already on it.

kwwall:
Prasad, I woke up sick this morning, so I would appreciate if you would
handle this. Thanks, -kevin

--
Blog: https://off-the-wall-security.blogspot.com/    | GitHub: @kwwall
<https://github.com/kwwall>| OWASP ESAPI Project
<https://owasp.org/www-project-enterprise-security-api/>co-lead | OWASP and
ACM lifetime member
NSA: All your crypto bit are belong to us.

Prasad-JB:
Got it, Kevin. Thanks for clarifying. I’ll open a follow-up PR to correct the ESAPI vs Java Encoder reference in the cheat sheet. Wishing you a speedy recovery. take care!

Issue #1716, opened 2025-06-28
New CS proposal: WebSocket Security Best Practices


maheshkukreja:
## What is the proposed Cheat Sheet about?

A security guide for WebSockets, a real-time communication technology commonly used in web apps (like chat, gaming, trading apps), addressing their unique security risks.

## What security issues are commonly encountered related to this area?

- Cross-Site WebSocket Hijacking (CSWSH)
- Lack of transport encryption (`ws://` instead of `wss://`)
- Weak or missing authentication
- Injection and cross-site scripting (XSS) through message payloads
- Insufficient input validation/sanitization
- Denial-of-service via WebSocket connections/messages
- Inadequate logging and monitoring for WebSocket interactions

## What is the objective of the Cheat Sheet?

- Help developers build secure WebSocket implementations
- Prevent common security mistakes unique to WebSockets
- Provide actionable best practices, including specific considerations across major programming languages and frameworks
- Guide compliance with security standards (PCI-DSS, HIPAA, GDPR)
- Improve WebSocket security testing capabilities

## What other resources exist in this area?

**OWASP currently has:**

- [HTML5 Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html#websockets) (briefly covers WebSockets - should we remove it from that cheatsheet and link to the new cheatsheet instead?)
- [Web Security Testing Guide](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/11-Client_Side_Testing/10-Testing_WebSockets) (basic coverage of WebSocket testing)
- 

**What’s missing:**

- No comprehensive, standalone guide focused specifically on WebSocket security best practices
- Framework-specific implementation guidance is minimal

**Why we need this:**

WebSockets are widely used in production but lack comprehensive, authoritative security guidance from OWASP. This cheat sheet would be OWASP’s first complete resource dedicated to WebSocket security best practices.

jmanico:
Great idea!

mackowski:
@maheshkukreja do you want to create draft of this cheatsheet?

maheshkukreja:
Hi @mackowski - I can, yes. But I'll need a couple of weeks to get to it.

maheshkukreja:
First draft [here,](https://github.com/OWASP/CheatSheetSeries/pull/1780) looking forward to the feedback.

PR #1780, opened 2025-09-03
add WebSocket_Security_Cheat_Sheet.md

maheshkukreja:
This PR fixes issue #1716.

As a follow up, we'll need to clean up the [HTML5 Security Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/HTML5_Security_Cheat_Sheet.md) to remove websocket related text in there.

### DIFF
--- a/cheatsheets/WebSocket_Security_Cheat_Sheet.md
+++ b/cheatsheets/WebSocket_Security_Cheat_Sheet.md
@@ -1,0 +1,251 @@
+# WebSocket Security Cheat Sheet
+
+## Introduction
+
+WebSockets enable real-time, bidirectional communication between clients and servers, powering applications like chat systems, live trading platforms, and collaborative tools. Unlike traditional HTTP requests, WebSocket connections remain open and allow continuous data exchange.
+
+However, WebSockets introduce security challenges that differ from standard web application security:
+
+- **Cross-Site WebSocket Hijacking (CSWSH)**: Attackers hijack authenticated connections from malicious websites
+- **Authentication bypass**: No built-in authentication makes access control easy to forget
+- **Injection attacks**: WebSocket messages can carry XSS, SQL injection, and other malicious payloads
+- **Denial-of-service**: Persistent connections enable new DoS attack vectors like connection exhaustion
+- **Monitoring gaps**: Traditional HTTP logs only capture the initial upgrade request, missing all message traffic
+
+**Real-world vulnerabilities:**
+
+- **Gitpod CSWSH (2023)**: [Insufficient origin validation](https://github.com/advisories/GHSA-f53g-frr2-jhpf) allowed full account takeover via hijacked WebSocket connections
+- **Spring RCE vulnerability**: [CVE-2018-1270](https://spring.io/security/cve-2018-1270) let attackers execute code through crafted STOMP messages
+
+## Primary Defenses
+
+### Transport Security
+
+#### Always Use WSS (WebSocket Secure)
+
+Never use unencrypted `ws://` connections in production. Unencrypted `ws://` connections allow eavesdropping and tampering.
+
+```javascript
+// Secure - always use this
+const socket = new WebSocket('wss://app.example.com/socket');
+
+// Insecure - never use in production
+// const socket = new WebSocket('ws://app.example.com/socket');
+```
+
+See the [Transport Layer Security Cheat Sheet](Transport_Layer_Security_Cheat_Sheet.md) for more details.
+
+#### WebSocket Protocol Configuration
+
+**Use modern protocol versions:**
+
+Only support [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455) (the current WebSocket standard). Drop backward compatibility for outdated versions like [Hixie-76](https://datatracker.ietf.org/doc/html/draft-hixie-thewebsocketprotocol-76) and [hybi-00](https://datatracker.ietf.org/doc/html/draft-ietf-hybi-thewebsocketprotocol-00) which have known security vulnerabilities.
+
+**Compression security:**
+
+Disable `permessage-deflate` compression unless specifically needed. Compression can introduce security vulnerabilities similar to CRIME/BREACH attacks where compression combined with secret data can leak information.
+
+```javascript
+// Node.js - disable compression for security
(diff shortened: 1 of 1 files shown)


randomstuff:
I don't have much data to back it up but I would think that memory(/resource) exhaustion caused by the lack of control flow/backpressure support in many WebSocket implementations may an important security considerations to add in there?

cheatsheets/WebSocket_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,251 @@
...
+CSWSH allows attackers to hijack authenticated WebSocket connections from malicious websites:
+
+1. User logs into your application (session cookie established)
+2. User later visits a malicious website
+3. Malicious site opens WebSocket to your application, browser sends cookies automatically
+4. Server accepts the connection → attacker gets live, authenticated WebSocket access
+
+#### Origin Header Validation
+
+The primary CSWSH defense is validating the `Origin` header during handshake. Browsers include this header and malicious JavaScript cannot override it.
```
randomstuff: Maybe:

* explicitly say that (in most cases), you want to validate the `Origin`
* explicitly say to use an allowlist (and not a denylist)?

cheatsheets/WebSocket_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,251 @@
...
+
+// Dangerous - enables code execution
+// const message = eval('(' + data + ')');
+```
+
+See the [Input Validation Cheat Sheet](Input_Validation_Cheat_Sheet.md) for more details.
+
+### Service Tunneling Risks
+
+Don't tunnel TCP services (VNC, FTP, SSH) through WebSockets. If your application has XSS vulnerabilities, attackers could access these services directly from victims' browsers.
```
randomstuff: (Rarely, that may be what you actually want to do though …)

cheatsheets/WebSocket_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,251 @@
...
+    return true;
+  }
+});
+```
+
+**Important:** Be specific with origins. Don't use wildcards or substring matching.
+
+#### Additional CSWSH Protections
+
+Beyond Origin validation, use **SameSite cookies** (`SameSite=Lax` or `Strict`) to prevent cross-site cookie transmission. Consider **token-based authentication** using one-time tokens instead of relying solely on cookies. For applications already using CSRF protection, include **CSRF tokens** in WebSocket handshakes.
```
randomstuff: Would to the cookie thing fit in the Session Management section?

cheatsheets/WebSocket_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Service Tunneling Risks
+
+Don't tunnel TCP services (VNC, FTP, SSH) through WebSockets. If your application has XSS vulnerabilities, attackers could access these services directly from victims' browsers.
+
+### Denial-of-Service Protection
+
+Persistent WebSocket connections increase DoS risk.
+
+**Limit connections and resources** by restricting total connections and implementing per-IP limits. Set **message size limits** (typically 64KB or less) and implement **rate limiting** to prevent message flooding - 100 messages per minute is a common starting point.
```
randomstuff: "per-IP" is possibly dubious (especially with IPv6…). In many cases, wouldn't it be better to use per-user limits instead?

cheatsheets/WebSocket_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,251 @@
...
+});
+```
+
+**Important:** Be specific with origins. Don't use wildcards or substring matching.
+
+#### Additional CSWSH Protections
+
+Beyond Origin validation, use **SameSite cookies** (`SameSite=Lax` or `Strict`) to prevent cross-site cookie transmission. Consider **token-based authentication** using one-time tokens instead of relying solely on cookies. For applications already using CSRF protection, include **CSRF tokens** in WebSocket handshakes.
+
+#### Session Management
```
randomstuff, line 104: Shall we add this as a subsection of an "Authentication" section? In many cases, you might be authenticated using a token which is not passed as a cookie. Because, the browser-based WebSocket API does not allow passing `Authorization` headers, you are often left with passing the token through query strings. This introduces a security issue: the token will leak into the access logs.

Another solution is to pass the token as part of a WebSocket message. (But this is not always possible, especially when you implement an well known WebSocket-based protocol).

maheshkukreja:
> I don't have much data to back it up but I would think that memory(/resource) exhaustion caused by the lack of control flow/backpressure support in many WebSocket implementations may an important security considerations to add in there?

Thanks for the feedback, @randomstuff - I think I addressed your concerns, let me know if there's anything else.

cheatsheets/WebSocket_Security_Cheat_Sheet.md
maheshkukreja: `Session Management` is already under `Authentication` at the moment. Added a `Token Based Authentication` section [here](https://github.com/OWASP/CheatSheetSeries/pull/1780/files#diff-e641453e754952e358a45de67862fc4c7452dfb18471865cba6b67d0e3f43c06R125) to cover other concerns.

jmanico:
> As a follow up, we'll need to clean up the [HTML5 Security Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/HTML5_Security_Cheat_Sheet.md) to remove websocket related text in there.

Can you kindly submit this as a separate issue?

maheshkukreja:
> > As a follow up, we'll need to clean up the [HTML5 Security Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/HTML5_Security_Cheat_Sheet.md) to remove websocket related text in there.
> 
> Can you kindly submit this as a separate issue?

https://github.com/OWASP/CheatSheetSeries/issues/1783

andrzejsydor (approved):

mackowski (approved):
Awesome work! Thanks @maheshkukreja

Issue #1782, opened 2025-09-04
Update: Authentication Cheat Sheet


markbastion:
## What is missing or needs to be updated?
Under the Implement Proper Password Strength Controls section on the Authentication Cheat Sheet 

Where it states:
Minimum length for passwords should be enforced by the application. Passwords shorter than 8 characters are considered to be weak ([NIST SP800-63B](https://pages.nist.gov/800-63-3/sp800-63b.html)).

## How should this be resolved?
The NIST guidance is actually a bit more specific 
https://pages.nist.gov/800-63-4/sp800-63b.html#passwordver

Where is states:
Verifiers and CSPs SHALL require passwords that are used as a single-factor authentication mechanism to be a minimum of 15 characters in length. Verifiers and CSPs MAY allow passwords that are only used as part of multi-factor authentication processes to be shorter but SHALL require them to be a minimum of eight characters in length.

So passwords should be at least 15 characters and can only be 8 characters if MFA is enabled according to NIST.

szh:
This is very interesting. I hadn't heard this - has this recommendation of 15 chars been implemented in the wild much?

jmanico:
This suggestion from NIST is very recent and I support it. It's also very logical. If you give us a PR I support it.

We are also making this change at ASVS 5.0.

PR #1785, opened 2025-09-06
Update to reflect latest change from NIST on password minimum length

markbastion:
Update to the Authentication Cheat Sheet to reflect latest change from NIST on password minimum length depending on if MFA is enabled or not.
Latest NIST guidance: https://pages.nist.gov/800-63-4/sp800-63b.html#passwordver

This PR addresses issue #1782.


mackowski (approved):
thanks @markbastion

Issue #1372, opened 2024-04-08
Update: LDAP_Injection_Prevention_Cheat_Sheet


einhirn:
## What is missing or needs to be updated?

After a casual reading, the Cheat Sheet seems to suggests that even passwords should always be escaped. When this is taken at face value, it can cause all kinds of issues, mainly stuff not working as expected any more (see https://github.com/dexidp/dex/issues/3433).

The `Safe Java Escaping Example` wrongly assumes that a password check against LDAP is done via LDAP filter and escapes the password, when actually password checking against an ldap server is (or at least should be) exclusively done via BIND request which safely accepts the password in unfiltered plaintext. 

## How should this be resolved?

The usual methods for password authentication against LDAP should be mentioned:
* Single-Step Auth:
  - BIND-Request with DN constructed from template like `uid=${USERID},ou=people,dc=example,dc=com`. Whatever is inserted in that template should be escaped as needed. Password from user input is given to BIND call as is.
* Three-Step-Auth:
  - BIND with credentials limited to searching - these would be configured in the app.
  - `SEARCH` LDAP for users DN like `(&(uid=${USERID})(deactivated=0))` or something. Again, anything inserted into that query should be escaped as needed
  - BIND with DN from search result above, if any. Password from user input is given to BIND call as is.
 
These two both assume the LDAP server to require (non-anonymous) BIND before allowing access to user data, which I think is (or should be) default or at least best practice. So `Enabling Bind Authentication` should be moved up in the document's focus, i.e. "To prevent LDAP Injection, first the LDAP server needs to be configured properly...". Even if it's out of scope of this document, proper safety precautions should be mentioned - "Simple BIND transmits plain text passwords to the server, so it should only be used with an encrypted connection" or "If you **really** need to use unencrypted connection, make sure to only use SASL BIND with challenge protocols etc. to prevent password sniffing."

The `Safe Java Escaping Example` should be updated to reflect either one of the above methods and no longer use the password in a search filter.

jmanico:
This cheatsheet can use a lot of changes. I like your ideas so far. Would you care to submit PR's?

einhirn:
I'll look into it and see when I might find some time to do that... I tend to overthink what I write on a word for word level, so it takes a lot of time to write something, especially something public in the field of security 😬

Jeymz:
> ## What is missing or needs to be updated?
[...]

PR #1797, opened 2025-09-12
feat(1372): LDAP Authentication vs Filter Clarity

Jeymz:
- Added a section regarding safe LDAP Authentication
- Updated an incorrect RFC link that I discovered during PR prep.

**Note:**
> I'm not an LDAP Injection wizard nor am I a Java Expert. Please validate the example prior to PR Approval.
> I just want to ensure we aren't misleading developers into thinking they can validate passwords using LDAP search filters.

This PR fixes issue #1372 .

### DIFF
--- a/cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
@@ -127,7 +127,7 @@ NamingEnumeration<SearchResult> results =
 
 [.NET AntiXSS](https://blogs.msdn.microsoft.com/securitytools/2010/09/30/antixss-4-0-released/) (now the Encoder class) has LDAP encoding functions including `Encoder.LdapFilterEncode(string)`, `Encoder.LdapDistinguishedNameEncode(string)` and `Encoder.LdapDistinguishedNameEncode(string, bool, bool)`.
 
-`Encoder.LdapFilterEncode` encodes input according to [RFC4515](https://tools.ietf.org/html/rfc4515) where unsafe values are converted to `\XX` where `XX` is the representation of the unsafe character.
+`Encoder.LdapFilterEncode` encodes input according to [RFC4515](https://datatracker.ietf.org/doc/html/rfc4515) where unsafe values are converted to `\XX` where `XX` is the representation of the unsafe character.
 
 `Encoder.LdapDistinguishedNameEncode` encodes input according to [RFC2253](https://tools.ietf.org/html/rfc2253) where unsafe characters are converted to `#XX` where `XX` is the representation of the unsafe character and the comma, plus, quote, slash, less than and great than signs are escaped using slash notation (`\X`). In addition to this a space or octothorpe (`#`) at the beginning of the input string is `\` escaped as is a space at the end of a string.
 


szh:
Hey @Jeymz! 👋

Jeymz:
> Hey @Jeymz! 👋

Why hello there captain! Hope things are well.

szh:
RFC link update looks good. Any Java experts here who can validate the example?

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -105,6 +105,54 @@ occur before the validation step (source: [Prevent LDAP injection](https://wiki.
...
+    throw new AuthenticationException("User not found");
+}
+
+SearchResult result = results.next();
+String userDN = result.getNameInNamespace();
+
+// Step 2: Bind with the DN and provided password
+Hashtable<String, String> env = new Hashtable<>();
+env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
+env.put(Context.PROVIDER_URL, "ldap://example.com:389");
```
kwwall: With 'simple' authN bind type, one really needs to use "ldaps://" which generally is on port 636. You probably can get away with using vanilla LDAP on port 389 for certain variations of SASL (like GSSAPI, SCRAM-SHA-256, etc.) but even then, I wouldn't generally recommend it.

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -105,6 +105,54 @@ occur before the validation step (source: [Prevent LDAP injection](https://wiki.
...
+✅ **Revised Secure Example (Java search + bind)**
+
+```java
+// Step 1: Look up the user DN with a parameterized search
+String searchFilter = "(&(uid={0})(objectClass=person))";
+SearchControls controls = new SearchControls();
+controls.setSearchScope(SearchControls.SUBTREE_SCOPE);
+
+NamingEnumeration<SearchResult> results = ctx.search(
+    "ou=users,dc=example,dc=com",
```
kwwall: This of course implies where you know to look and assumes the objectClass will be 'person'. That may not be a valid assumption for things like service accounts that are not associated with people.

Another approach that I've seen back when I worked for a telecommunications company was to do an anonymous subtree search first for the UID (or sometimes, a CN) and then matching DNs (usually only one) would be returned, and the BIND operation would be done on that. YMMV.

Jeymz: I've tried to take a stab and using your approach for the example. Thanks for the keen eye on this. I appreciate reviewers that add valuable insights like this to help make implementing these patterns easier for real world scenarios. Really wish I knew more Java.

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
Jeymz: I've pushed an update to address this. Can't believe I missed that in the example. Stupid Jeymz... Good thing we got great members like you to filter out my stupidity 😅

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -105,6 +105,63 @@ occur before the validation step (source: [Prevent LDAP injection](https://wiki.
...
+
+NamingEnumeration<SearchResult> results = anonCtx.search(
+    "ou=users,dc=example,dc=com",
+    searchFilter,
+    new Object[]{ username },
+    controls
+);
+
+if (!results.hasMore()) {
+    throw new AuthenticationException("User not found"); // Obscure to invalid username and/or password client side
```
kwwall: Not critical, but UIDs are generally considered not considered sensitive data within a company, where this is likely to be used, so I would advise liksting the uid that you were searching for and did not find here. If the exceptions are eventually logged to a SIEM, that information can be heplful.

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -105,6 +105,63 @@ occur before the validation step (source: [Prevent LDAP injection](https://wiki.
...
+anonEnv.put(Context.SECURITY_AUTHENTICATION, "none");
+
+DirContext anonCtx = new InitialDirContext(anonEnv);
+
+String searchFilter = "(uid={0})"; // No assumption about objectClass
+SearchControls controls = new SearchControls();
+controls.setSearchScope(SearchControls.SUBTREE_SCOPE);
+
+NamingEnumeration<SearchResult> results = anonCtx.search(
+    "ou=users,dc=example,dc=com",
```
kwwall: My comment about doing the anonymous search was more aimed at LDAP directory services where UIDs could be in _multiple_ trees (e.g, perhaps "ou=service_accounts" besides "ou=people"), so to illustrate this, you could redo your LDAP search to look in both of those OUs where accounts are stored (preferred) or anchor the search at just "dc=example,dc=com" (a bit less preferred. But if you know all your accounts are directly under "ou=people" it's less necessary.

kwwall (commented):
I'm not quite ready to approve this. While overall, I think the code looks fine (despite my 2 comments), I still have concerns about whether or not this code example actually addresses the points raised by GitHub issue #1372. I almost think that this code could potentially distract from the 2 main point raised in the issue that "BIND should be used for authentication via LDAP and that passwords should not be escaped as part of that because passwords should not be part of any LDAP filter. Instead of showing additional code and raising additional issues such as future requests to show additional examples using a SASL mechanism or if we would show an example that uses (say) the [Mozilla LDAP Java SDK](https://www-archive.mozilla.org/directory/javasdk) or Spring's LdapTemplate, or a C# .NET example, I'm wondering if we should just put the main points that @einhirn originally brought up in the GitHub issue #1372. And if we really want some examples of LDAP authentication in this CS, rather than using up space, we should just point to a couple of classic examples such as one or mohttps://docs.oracle.com/javase/8/docs/technotes/guides/jndi/jndi-ldap.htmlre of these:

- https://docs.oracle.com/javase/8/docs/technotes/guides/jndi/jndi-ldap.html
- https://www.geeksforgeeks.org/advance-java/spring-security-with-ldap-authentication/
- https://www.baeldung.com/java-ldap-auth
- Etc.
If we do decide to include an LDAP Bind code snippet, I think we should at least drop it into a collapsible section since LDAP authentication is not the main purpose of this cheat sheet.

Jeymz:
@kwwall I think I may have been trying to stretch too far to justify the example above. Instead let's just remove the insecure aspects of that example to clarify that we should never use an LDAP search to AuthN a user account and focus only on why allow lists are a must when dealing with unsafe LDAP Query construction.

Jeymz:
@szh Is there any way we can either get some eyes on this to either complete or close this PR?

szh:
@kwwall does it look ok to you now without the example?

cheatsheets/LDAP_Injection_Prevention_Cheat_Sheet.md
```diff
@@ -128,7 +127,7 @@ NamingEnumeration<SearchResult> results =
 
 [.NET AntiXSS](https://blogs.msdn.microsoft.com/securitytools/2010/09/30/antixss-4-0-released/) (now the Encoder class) has LDAP encoding functions including `Encoder.LdapFilterEncode(string)`, `Encoder.LdapDistinguishedNameEncode(string)` and `Encoder.LdapDistinguishedNameEncode(string, bool, bool)`.
 
-`Encoder.LdapFilterEncode` encodes input according to [RFC4515](https://tools.ietf.org/search/rfc4515) where unsafe values are converted to `\XX` where `XX` is the representation of the unsafe character.
+`Encoder.LdapFilterEncode` encodes input according to [RFC4515](https://tools.ietf.org/html/rfc4515) where unsafe values are converted to `\XX` where `XX` is the representation of the unsafe character.
```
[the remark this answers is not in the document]
Jeymz: The link `https://tools.ietf.org/html/rfc4515` redirects to `https://datatracker.ietf.org/doc/html/rfc4515`. Either way should work, I only chose to use the former since it was closest in similarity to the original link used prior to my update. @szh which direction would you like me to take with this?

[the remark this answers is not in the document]
mackowski: I think we can link directly to https://datatracker.ietf.org/doc/html/rfc4515

mackowski:
@jmanico @szh @kwwall - are you ok with the this PR?

PR #1866, opened 2025-10-24
Fix argon2id parameter selection advice

JakeGinesin:
OWASP recommends argon2id parameters that are lower than what is recommended by the standardizing RFC, RFC9106. 

Also, the text "These configuration settings provide an equal level of defense, and the only difference is a trade off between CPU and RAM usage" is wrong; modifying the memory parameter affects how parallelizable brute-force attacks are on the given hash digest, thereby affecting security.

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -106,7 +106,7 @@ Rather than a simple work factor like other algorithms, Argon2id has three diffe
 - m=2097152 (2 GiB), t=1, p=1; for "a default setting for all environments"
 - m=65536 (64 MiB), t=3, p=1; for "memory-constrained environments"
 
-These configuration settings provide different levels of defense, as increasing the memory parameter makes brute-force attacks harder to parallelize due to higher RAM requirements per hash. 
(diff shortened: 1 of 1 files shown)


andrzejsydor (approved):

jmanico:
My take:

* 2 GiB per hash means that authenticating even a few concurrent users (say, 20–50) could easily consume tens of gigabytes of RAM.
* That’s suitable for high-security offline key derivation (e.g., password managers, master key encryption), but not for online login APIs.

Our resident expert here ( @Sc00bz ) may have other thoughts on this matter.

randomstuff:
FWIW, this is what is indicated in ASVS 5.0:

| KDF        | Reference | Required Parameters | Status |
| ---------- | --------- | ------------ |:-:|
| argon2id | [RFC 9106](https://www.rfc-editor.org/info/rfc9106) | t = 1: m ≥ 47104 (46 MiB), p = 1 | A |
|          |                                                     | t = 2: m ≥ 19456 (19 MiB), p = 1 | A |
|          |                                                     | t ≥ 3: m ≥ 12288 (12 MiB), p = 1 | A |

jmanico:
Unless @Sc00bz wishes to reopen this, I am politely closing this out. We already have done a great deal of research on this and I feel the suggestions here are not applicable to scalable password storage.

Issue #1803, opened 2025-09-19
Update: Cross-Site Request Forgery Prevention Cheat Sheet - add Fetch Metadata


mkhanas:
## What is missing or needs to be updated?
The **CSRF Cheat Sheet** currently lacks any mention of **Fetch Metadata (`Sec-Fetch-*`) headers.** `Sec-Fetch-Site` (and related `Sec-Fetch-*` headers) let servers determine the *context* of a request (e.g. `same-origin`, `same-site`, `none`, `cross-site`) and therefore provide a reliable way to block obvious cross-site requests before they reach application logic.

While modern browsers provide these headers and many sources recommend using them as a CSRF mitigation:

1. Other OWASP cheat sheets (for example, *Cookie Theft Mitigation* and *XS-Leaks*) already mention `Sec-Fetch-*`. The CSRF Cheat Sheet should likewise document this technique as a **defense-in-depth** option.
2. Public guidance ([[MDN](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/CSRF)](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/CSRF), [[web.dev](https://web.dev/articles/fetch-metadata)](https://web.dev/articles/fetch-metadata)) and active community discussion (for example, Go issue [[#73626](https://github.com/golang/go/issues/73626)](https://github.com/golang/go/issues/73626)) highlight growing adoption and interest in using Fetch Metadata to reduce CSRF risk.
3. Early adopters and libraries (e.g., [[tower-sec-fetch](https://github.com/matteojoliveau/tower-sec-fetch)](https://github.com/matteojoliveau/tower-sec-fetch)) demonstrate practical implementations and demand for canonical guidance.

developers following the OWASP CSRF Cheat Sheet do not currently have OWASP-backed guidance on how to apply Fetch Metadata (`Sec-Fetch-*`) checks safely and correctly, or how to combine them with existing defenses (SameSite, tokens, Origin/Referer checks).

## How should this be resolved?
Add a subsection **“Fetch Metadata (`Sec-Fetch-*`) — an additional defense-in-depth control”** under the *Defense-in-Depth Techniques* section. The subsection should:

1. **Explain the concept**
    - What `Sec-Fetch-Site`, `Sec-Fetch-Mode`, `Sec-Fetch-Dest`, and `Sec-Fetch-User` are and why `Sec-Fetch-Site` is most relevant for CSRF protection.
2. **Describe recommended policy**
[...]

PR #1875, opened 2025-11-02
Update Fetch Metadata positioning

mkhanas:
As agreed in https://github.com/OWASP/CheatSheetSeries/issues/1803 :
- Updated Fetch Metadata positioning
- Extended core guidance to mention Fetch Metadata request headers (Sec-Fetch-*) as an alternative to CSRF tokens for state-changing requests.
- Added clarification that developers can use CSRF tokens or Fetch Metadata depending on project needs and client compatibility.

- [x] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](../templates/New_CheatSheet.md).

Please select one of the following options:

- [x] I have NOT used any AI tool to generate the contents of this PR.
    and the prompt used is `[your prompt here]`. [Feel free to add more details if needed]

This PR fixes issue #1803.

### DIFF
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -13,7 +13,7 @@ In short, the following principles should be followed to defend against CSRF:
 **IMPORTANT: Remember that Cross-Site Scripting (XSS) can defeat all CSRF mitigation techniques!** While Cross-Site Scripting (XSS) vulnerabilities can bypass CSRF protections, CSRF tokens are still essential for web applications that rely on cookies for authentication. Consider the client and authentication method to determine the best approach for CSRF protection in your application.
 
 - **See the OWASP [XSS Prevention Cheat Sheet](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) for detailed guidance on how to prevent XSS flaws.**
-- **First, check if your framework has [built-in CSRF protection](#use-built-in-or-existing-csrf-implementations-for-csrf-protection) and use it**
+- **First, check if your framework has [built-in CSRF protection](#built-in-or-existing-csrf-implementations-for-csrf-protection) and use it**
 - **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state-changing requests (requests that cause actions on the site) and validate them on the backend.**
 - **If your software targets only modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) together with the fallback options described below to block cross-site state-changing requests.**
 - **Stateful software should use the [synchronizer token pattern](#synchronizer-token-pattern)**


andrzejsydor (approved):

nickchomey:
Thanks for taking the lead on this! You'll definitely want a review by someone more knowledgeable (eg. @FiloSottile), but here's some feedback.

I believe it was agreed in https://github.com/OWASP/CheatSheetSeries/issues/1803 that "Fetch metadata is a complete and robust fix for CSRF, not just defense in depth." However, this PR reads unnecessarily precautionary. 
* First, i think it would be easier for people to review if you didn't move the Fetch Metadata headers section. The whole thing is a diff, rather than just specific wording. If it should be moved, that could be done before merging the PR, after approval of the language.
* Various places call it a "primary" protection, which still suggests it is defense in depth. Such language is not used with respect to the CSRF tokens sections
* The concerns about browser compatibility are vastly overstated. It has [97.6% coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site), which is practically 100%. Only some ancient Chrome, IE, and 2+ year old Safari (which surely will be updated soon enough) are unsupported. The real security risk is anyone *using* these browsers - you'd be doing them a favour by returning a message saying "update your browser if you want to use this site". Moreover, Origin headers - which are used by the Go implementation - have effectively 100% coverage
* non-browser (eg curl) request support is a non-issue
* It is referred to as "relatively new", "simple" etc - this undermines its robustness. 
* Using Origin headers as the first fallback should be strongly/mandatorily recommended. In fact, it could link to the [Go CrossOriginProtection source code](https://go.dev/src/net/http/csrf.go) as a model implementation.
* The limitations and gotchas section was not modified at all. Again, browser compatibility is not an issue. There's no substantiation of the concerns about blocking legitimate CORS or 3rd party flows - either these concerns should be listed, along with mitigations, or not mentioned at all. I dont see how it is a limitation that it needs to work with https - that's simply good practice. In fact, HSTS should be encouraged.

mkhanas:
Hi, @nickchomey! Thanks a lot for the thorough review and for the pointers, here some thoughts

> this PR reads unnecessarily precautionary

You’re right — that’s on me. The phrasing ended up sounding more precautionary than intended; it reflects my personal stance on the topic

> First, i think it would be easier for people to review if you didn't move the Fetch Metadata headers section. The whole thing is a diff, rather than just specific wording. If it should be moved, that could be done before merging the PR, after approval of the language.

I moved the Fetch Metadata section because we’re proposing it be treated as one of the primary mitigations. If we leave it buried, readers may miss that change.

> Various places call it a "primary" protection, which still suggests it is defense in depth. Such language is not used with respect to the CSRF tokens sections

I understand your point — my intent was to reflect the current state of adoption and confidence rather than to imply it’s inherently secondary.

As I mentioned in the issue discussion, Fetch Metadata is still relatively new compared to traditional CSRF mitigations. While it’s an elegant and highly effective mechanism for modern browsers, it hasn’t yet gone through the same long cycle of real-world testing, widespread deployment, and maturity that CSRF tokens have. In security, mechanisms often take time to earn broad trust — a good parallel might be the SameSite cookie attribute, which was proposed years before browsers began enforcing it by default and before most frameworks integrated it as a standard protection.

That said, I’m open to rephrasing. Which parts stand out to you most?

> The concerns about browser compatibility are vastly overstated.

My point wasn’t to question the coverage but rather to acknowledge that, as a community, we can’t dictate every project’s goals or constraints.

If the software targets modern browsers, then Fetch Metadata is clearly the way to go. However, as we discussed, for environments where these headers aren’t yet supported, developers will still need to rely on CSRF tokens — even if that feels less practical.

> non-browser (eg curl) request support is a non-issue

Agree

> The limitations and gotchas section was not modified at all.

Regarding the note on blocking legitimate CORS or third-party flows — I intentionally kept it, just to highlight that these areas need extra care, since enabling Fetch Metadata protections without considering such flows could break them. The relevant mitigations are already covered in section 3.2 (“How to treat Fetch Metadata headers on the server-side”) section.

> In fact, HSTS should be encouraged.

Good catch! I hadn’t considered mentioning it explicitly, but that’s a great recommendation and would indeed address the “potentially trustworthy” issue as well. I’ll incorporate that.

nickchomey:
I can't help but think that your response was written almost completely by AI... It makes me wary of collaborating any further on this as its not clear to me that a thoughtful human is actually on the other side...

Perhaps incorporate my suggestions into a new commit and we can see where we're at then

mkhanas:
I don’t like where this is headed, especially with things starting to feel a bit personal. 

It feels like the focus is mainly on getting Fetch Metadata headers recognized as the primary CSRF protection and moving away from CSRF tokens entirely, which doesn’t sit well with me. That’s just my personal view, but I worry it could lead to CSRF tokens falling out of use altogether, since most developers would naturally choose the simpler Fetch Metadata approach.

I get your point though, and as I mentioned in my previous comment, this section reflects my own perspective. It’s admittedly an awkward position — I don’t have solid evidence that Fetch Metadata isn’t as robust as you’d like to present it, but at the same time, “lack of evidence doesn’t mean lack of existence” isn’t a strong argument either. I’m just not sure how to balance both sides — treating Fetch Metadata as “the best” while still keeping CSRF tokens relevant.

mkhanas:
Feel free to open your own PR — I’m not comfortable positioning Fetch Metadata as THE MAIN CSRF defense.

nickchomey:
These last 2 comments, and those from the previous issue, have no hint of AI.

I don't think anyone was ever advocating for fetch metadata being THE primary csrf protection. Just that it is suitable as A Primary/standalone protection, if a few niche caveats and associated mitigation are presented.

>  I worry it could lead to CSRF tokens falling out of use altogether, since most developers would naturally choose the simpler Fetch Metadata approach.

Though, this is *precisely* what people are ultimately advocating for - CSRF tokens are a *headache* (and therefore much more likely to implemented poorly due to human error). They also make caching extremely difficult. 

I see no conflict or even issue with effectively saying "these are two worthy options. Take your pick". Most would, indeed, eventually pick fetch metadata, and tokens would eventually be forgotten. What's wrong with that? That's just another example of good technical progress for the web platform.

mkhanas:
> These last 2 comments, and those from the previous issue, have no hint of AI.

I’m not a native speaker, so in the first comment I was just trying to politely make my point. When I saw you didn’t like my style, I just dropped it :)

> I see no conflict or even issue with effectively saying "these are two worthy options. Take your pick". Most would, indeed, eventually pick fetch metadata, and tokens would eventually be forgotten.

I know, and that makes perfect sense. I was just trying to play it safe and leave room for discussion, something like:
“If your app is intended to be used only on newer versions (provide versions that support Fetch Metadata headers), go with Fetch Metadata. If for some reason you still need to support older versions (shame on you), use CSRF tokens.”

Let’s do it this way: I’ll fix the language according to the suggestions in your first message, add the missing bits about HSTS, keep the Metadata section at the beginning (so we don’t have to open another PR to move it later), and then ask for a review from the rest — especially the Go folks, where this all started.

mkhanas:
Hi, @nickchomey 
Tried to incorporate most of your suggestions. Most changes are clarifications, wording improvements, and notes about compatibility and HSTS. Feel free to take a look when you have a chance

jmanico:
> I’m not comfortable positioning Fetch Metadata as THE MAIN CSRF defense

As the founder of the cheatsheet series and project lead, I agree 100%. 

Fetch Metadata s a very useful first line defense against CSRF, but it is not safe to rely on it alone for all production workloads. Use it as part of defense-in-depth. Combine it with SameSite cookies, origin checks, and per-request CSRF tokens.

The spec’s notion of “user-initiated” navigation and some navigation flows (top-level navigations, prerender/prefetch, PaymentRequest-like flows) can result in values that permit requests you’d expect to be blocked; attackers can sometimes craft flows to exploit those behaviors. There is research showing odd corner cases.

So let's revisit this in a year, but for now, Fetch Metadata is one defense with limitations and is not the only be-all defense for CSRF.

Anywhere we say this, I want it softened. And again, we can revisit this in a year.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
 
+## Fetch Metadata headers
+
+Fetch Metadata request headers provide extra context about how an HTTP request was made, and how the resource will be used, enabling servers to reject suspicious cross-site requests. Servers can use these headers — most importantly `Sec-Fetch-Site` — as a lightweight and reliable method to block obvious cross-site requests. See the [Fetch Metadata specification](https://www.w3.org/TR/fetch-metadata/) for details.
+
+Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) using the `Origin` and `Referer` headers **is a mandatory requirement** for any Fetch Metadata implementation.
```
artis3n:
```suggestion
Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) using the `Origin` and `Referer` headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/?search=sec-fetch-site) in all desktop browsers since March 2023 (with the exception of IE as it no longer receives updates and does not support `Sec-Fetch-*` headers) and all mobile browsers since October 2025.
```

artis3n: "legacy" and "modern" are imprecise to describe browser versions, and their meanings will change often. We should be precise about what browsers we're considering "modern" here (all web browsers since 2023 and all* mobile browsers since 2025). *`Sec-Fetch-User` is not yet supported by Safari on web or mobile, but it also isn't used in the Sec-Fetch-Site scheme so not sure how to capture that.

If we only discuss `Sec-Fetch-Site` since that's all that is needed for CSRF protection in the `Sec-Fetch-*` group then we don't need to worry about this caveat.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+The Fetch Metadata request headers are:
```
artis3n, line 164: Should this section be focused on the subset of Sec-Fetch-* headers that are used to protect against CSRF? (Just `Sec-Fetch-Site` and Origin/Host header validation, and link off to MDN or something "if you'd like to learn more about what other Sec-Fetch-* headers are available for other purposes, see..."

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+    }
+    ```
+
+    1.3. Allow none for user-driven top-level navigations (bookmarks, typed URLs, explicit form submits) where appropriate.
+
+2. If `Sec-Fetch-*` headers are absent: choose a fallback based on risk and compatibility requirements:
+    2.1. Fail-safe (recommended for sensitive endpoints): treat absence as unknown and block the request.
+    2.2. Fail-open (compatibility-first): fallback to other security measure ([standard origin verification](#using-standard-headers-to-verify-origin), CSRF tokens, and/or require additional validation).
+
+3. Additionall options
```
artis3n:
```suggestion
3. Additional options
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+        req.method === 'GET' &&
+        req.get('Sec-Fetch-Dest') !== 'object' &&
+        req.get('Sec-Fetch-Dest') !== 'embed') {
+      return true; // Allow this request
+    }
+    ```
+
+    3.2 Whitelist explicit cross-origin flows. If certain endpoints intentionally accept cross-origin requests (CORS JSON APIs, third-party integrations, webhooks), explicitly exempt those endpoints from the global Sec-Fetch deny policy and secure them with proper CORS configuration, authentication, and logging.
+
+### Things to consider
```
artis3n: How about a section for Requirements / Pre-Conditions/ etc. to indicate what must already be present in order to rely on this method? (Must enforce HTTPS only or be using localhost, must not use GET requests for state-changing requests, recommended to set HSTS).

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+### Browser compatibility
+
+Fetch Metadata request headers are supported in most modern browsers on both desktop and mobile (Chrome, Edge, Firefox, Safari 16.4+, and even in webviews on both iOS and Android), with [over 97% global coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site). For compatibility detail, see the [browser support table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).
+
+For the rare cases of outdated or embedded browsers that lack `Sec-Fetch-*` support, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) should provide the required coverage. If this is acceptable for your project, consider prompting users to update their browsers, as they are running on outdated and potentially insecure versions.
+
+### How to treat Fetch Metadata headers on the server-side
```
artis3n, line 181: I'd recommend we update this section. Maintaining language agnosticity, I'd like to see the pattern outlined in https://words.filippo.io/csrf/#protecting-against-csrf-in-2025 personally.

1. Validate Origin header against an allowlist (for the JS example, recommend configuring as a `new Set([domains])` for an easy `set.has(Origin)` as this validation step
2. Check if Sec-Fetch-Site header is present
    a. If present, allow request if value is `same-origin` or `none`, else deny
3. If Sec-Fetch-Site and Origin headers are both missing, pass request through
4. Validate Origin header against Host header and pass request if they match, else reject (do this at the end and not with initial Origin validation due to step 3 support of significantly legacy browsers pre-2020)

Also the pattern described, basically, in https://web.dev/articles/fetch-metadata#how_to_use_fetch_metadata_to_protect_against_cross-origin_attacks .

artis3n: And I'm partial since it is the language I am usually in but would be nice to note with the examples that Go developers can follow the above pattern by just using https://pkg.go.dev/net/http@go1.25rc2#CrossOriginProtection in the standard library as of Go 1.25.

Perhaps similar to https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-automatically-including-csrf-tokens-as-an-ajax-request-header , a section for `Go: Use net/http CrossOriginProtection middleware` with a link out to those docs.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
 
 Though the Naive Double-Submit Cookie method is simple and scalable, it remains vulnerable to cookie injection attacks, especially when attackers control subdomains or network environments allowing them to plant or overwrite cookies. For instance, an attacker-controlled subdomain (e.g., via DNS takeover) could inject a matching cookie and thus forge a valid request token. [This resource](https://owasp.org/www-chapter-london/assets/slides/David_Johansson-Double_Defeat_of_Double-Submit_Cookie.pdf) details these vulnerabilities. Therefore, always prefer the _Signed Double-Submit Cookie_ pattern with session-bound HMAC tokens to mitigate these threats.
 
+## Fetch Metadata headers
```
artis3n, line 158: Perhaps a larger reordering of the content. We have `Signed Double-Submit Cookie (RECOMMENDED)` in the middle of the table of contents (the most complex developer option, I might add), which is already hard to find as there are a _lot_ of headers on this page.

Perhaps instead:

```markdown
## Recommended Mitigation Patterns
### Signed Double Submit Cookie
### Sec-Fetch-Site Header
## Discouraged Mitigation Patterns
### Token-Based Mitigations
#### Naive Double-Submit Cookie Pattern
### ...
```

We already have a `Defense In Depth Techniques` section so that handles defense-in-depth approaches for both recommended patterns. This simplifies the wall of content in the table of contents and raises up the actual steps we'd want developers to take to protect their applications.

artis3n (commented):
First time commenter, but I am very motivated by making security simple and I like the Sec-Fetch-Site pattern very much. Added some recommendations about structuring this page.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Hi @artis3n, sorry for the wait on the review

If I understood your comment correctly, you’re suggesting narrowing this section down to just Sec-Fetch-Site and Origin/Host header validation. I agree this would simplify things, but my thinking was that it’s important to keep the broader context of Fetch Metadata headers, since they work as a family and the spec encourages using them together.

> and link off to MDN or something

we already link out to the W3C spec for readers who want deeper detail.

Maybe you can draft a short suggestion of what you’d like this section to look like?

mkhanas: Also, there are examples of Sec-Fetch-Mode, Sec-Fetch-Dest, and Sec-Fetch-User further down in the section. 

Because of that, I think it makes sense to keep the broader description here, so all the related information stays in one place and developers don’t need to jump between different resources to understand how the full set of headers works together.

artis3n: That makes sense if we want to keep them co-located. I do think we should distinguish between the headers you _need_ to use for CSRF protection vs. the ones you _can_ use for additional checks. E.g. mode, dest, and user.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: This PR's aproach was actually based primarily on the approach outlined in https://web.dev/articles/fetch-metadata#how_to_use_fetch_metadata_to_protect_against_cross-origin_attacks, however, I intentionally changed the flow to make it deny-by-default, rather than the more allow-by-default used in both that article and Filippo’s blog.

Both external sources propose flows that can short-circuit early and allow requests without evaluating Fetch Metadata. For example:

> If the Origin header matches an allow-list of trusted origins, allow the request.

In this example, the request never reaches any Fetch Metadata checks. That effectively makes the Origin header the primary defense.

> These requests are not from (post-2020) browsers, and can’t be affected by CSRF.

I’m not fully comfortable relying on that assumption. My interpretation is that it’s based on the earlier point that older (pre-2020) browsers sometimes omitted the Origin header on POST. The wording could also be misread as “pre-2020 browsers can’t be affected by CSRF,” which I’m sure isn’t the intended meaning.

if as a community we agree that Fetch Metadata is the stronger and more reliable signal in modern browsers, then it seems logical to evaluate Sec-Fetch-Site first and use standard headers (Origin/Referer/Host) only as a fallback for legacy or non-compliant clients.

mkhanas: > a section for Go: Use net/http CrossOriginProtection middleware with a link out to those docs.

Regarding your second suggestion, that’s a great idea. We should definitely add a Go section to the “CSRF Prevention in Modern Frameworks” area. I can write a short part for it and include a reference to the CrossOriginProtection middleware along with a link to the documentation.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Agree, this originally started as a “things to consider” section, but during the PR review we already narrowed it down to notes about trustworthy URLs and HSTS. We can definitely rephrase it as requirements rather than optional considerations.

Regarding “must not use GET requests for state-changing requests,” there’s already a dedicated section for that, but we can briefly mention it here as well.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: First of all, YES! With the Fetch Metadata section added, the document has become a bit harder to navigate. I agree that restructuring the table of contents might be great, but I think this deserves a separate issue and PR once we finish resolving all the Fetch Metadata related changes.

Regarding the structure you’re suggesting:
I don’t think we should introduce a “Discouraged Mitigation Patterns” top-level section. In the current document, the DISCOURAGED subsection under the double-submit pattern is useful specifically as an explicit example of what not to implement, especially since, as you mentioned, it’s the most complex developer-controlled option. But creating a standalone section about DISCOURAGED approaches feels too strong, better not have them at all)

Instead, I believe the structure could look more like this:
```
## Disallowing Simple Requests
## Token-Based Mitigation
...
## Fetch Metadata Headers
## Dealing with Client-Side CSRF Attacks
## Defense In Depth Techniques
```

the first recommendation should be disallowing simple requests, and only after that should we list the available protection mechanisms.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: mode, dest, and user are still part of the overall Fetch Metadata–based CSRF protection strategy, the Sec-Fetch-Site header is just the primary signal we rely on. But I get your point, I’ll try to update the wording to make that distinction clearer.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
artis3n: That makes sense!

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -14,7 +14,8 @@ In short, the following principles should be followed to defend against CSRF:
 
 - **See the OWASP [XSS Prevention Cheat Sheet](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) for detailed guidance on how to prevent XSS flaws.**
 - **First, check if your framework has [built-in CSRF protection](#use-built-in-or-existing-csrf-implementations-for-csrf-protection) and use it**
-- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state changing requests (requests that cause actions on the site) and validate them on the backend**
+- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state changing requests (requests that cause actions on the site) and validate them on the backend, or validate [Fetch Metadata headers](#fetch-metadata-headers) on the backend for all state-changing requests.**
+- **If your software is intended to be used only on modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) to block cross-site state-changing requests**
```
nickchomey: I'm happy to see this here, up front. But, in this case, I'll be the stickler! Even if you intend to only serve modern browsers, obviously old browsers may still connect to your site. So, perhaps we should add something to the effect of 

```suggestion
- **If your software is intended to be used only on modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) to block cross-site state-changing requests, so long as you use the fallback options detailed below.**
```

Or is the point here to be brief?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
 
+## Fetch Metadata headers
+
+Fetch Metadata request headers provide extra context about how an HTTP request was made, and how the resource will be used, enabling servers to reject suspicious cross-site requests. Servers can use these headers — most importantly `Sec-Fetch-Site` — as a lightweight and reliable method to block obvious cross-site requests. See the [Fetch Metadata specification](https://www.w3.org/TR/fetch-metadata/) for details.
+
+Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) using the `Origin` and `Referer` headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/?search=sec-fetch-site) in all desktop browsers since March 2023 (with the exception of IE as it no longer receives updates and does not support `Sec-Fetch-*` headers) and all mobile browsers since October 2025.
```
nickchomey:
```suggestion
Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) using the `Origin` and `Referer` headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/?search=sec-fetch-site) in all browsers since March 2023 (with the exception of IE as it no longer receives updates and does not support `Sec-Fetch-*` headers).
```

When saying that Mobile compatibility only happened in Oct 2025, i suspect you got duped by caniuse's weird peculiarity that it usually/always? just lists the latest Android Chromium version for compatibility. Sec-Fetch-Site has been available there since 2019 - same as desktop.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
nickchomey: I agree that making it clear that sec-fetch-site is the primary signal would be important, so as to not overcomplicate/overwhelm people. The others can be included for reference/deep dive.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +152,96 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+If any of headers above contain values not listed in the specification, in order to support forward-compatibility, servers should ignore those headers.
+
+### Ease of use
+
+Unlike [synchronizer tokens](#synchronizer-token-pattern) or [double-submit patterns](#alternative-using-a-double-submit-cookie-pattern) — which require additional client/server coordination and are easy to misimplement — Fetch Metadata checks are straightforward to implement correctly. They typically require only a small amount of server-side logic (inspect Sec-Fetch-Site, optionally refine with Sec-Fetch-Mode/Sec-Fetch-Dest) and no client changes. That simplicity reduces complexity, making the approach attractive for many applications.
+
+### Browser compatibility
+
+Fetch Metadata request headers are supported in most modern browsers on both desktop and mobile (Chrome, Edge, Firefox, Safari 16.4+, and even in webviews on both iOS and Android), with [over 97% global coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site). For compatibility detail, see the [browser support table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).
```
nickchomey:
```suggestion
Fetch Metadata request headers are supported in most modern browsers on both desktop and mobile (Chrome, Edge, Firefox, Safari 16.4+, and even in webviews on both iOS and Android), with [over 98% global coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site). For compatibility detail, see the [browser support table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).
```
Its 98.02 now

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
nickchomey: While the blog post uses Origin first, the actual [Go source code](https://go.dev/src/net/http/csrf.go#:~:text=func%20(c%20*CrossOriginProtection)%20Check(req%20*Request)%20error%20%7B) checks sec-fetch-site first:

I have to assume this concern was raised while implementing the feature, and thus appropriately uses origin as a fallback.

nickchomey (changes requested):
sorry i couldnt look at this til now. I made a few small suggestions

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: No discussion needed here

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Yeah, my mistake, thanks for pointing this out!

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: good for us)

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Thanks for bringing this up, I went straight for the blog. 

As we discussed, we’ll definitely add a reference to the Go implementation in the Frameworks section at the end of the document. Do you feel the current “### How to treat Fetch Metadata headers on the server-side” section needs any updates? @nickchomey

mkhanas:
@nickchomey Another great idea, we should definitely add that, otherwise it’ll get lost. My first thought was to put it under the “Things to consider” subsection, but we already discussed converting that into Requirements. 

Now I’m thinking that at the end of the Fetch Metadata section we should have three subsections: Concerns (including @jmanico’s remaining points), Requirements (trusted URLs, localhost, HSTS), and the existing “Rollout & testing recommendations”

nickchomey:
That sounds fine to me

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Updated

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Added this as part of "### Use Built-In Or Existing CSRF Implementations for CSRF Protection"

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: created a separate issue for this one - https://github.com/OWASP/CheatSheetSeries/issues/1892,  resolving the conversation now.

cc  @artis3n

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Updated, feel free to resolve the conversation

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -14,7 +14,8 @@ In short, the following principles should be followed to defend against CSRF:
 
 - **See the OWASP [XSS Prevention Cheat Sheet](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) for detailed guidance on how to prevent XSS flaws.**
 - **First, check if your framework has [built-in CSRF protection](#use-built-in-or-existing-csrf-implementations-for-csrf-protection) and use it**
-- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state changing requests (requests that cause actions on the site) and validate them on the backend**
+- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state changing requests (requests that cause actions on the site) and validate them on the backend, or validate [Fetch Metadata headers](#fetch-metadata-headers) on the backend for all state-changing requests.**
+- **If your software is intended to be used only on modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) to block cross-site state-changing requests, so long as you use the fallback options detailed below.**
```
nickchomey: I don't think the 2nd half of the first bullet should have been added. It is completely redundant with the 2nd bullet (and also has a lot of repetitive wording with the rest of the 1st bullet.


```suggestion
- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state-changing requests (requests that cause actions on the site) and validate them on the backend.**
- **If your software is intended to be used only on modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) to block cross-site state-changing requests, so long as you use the fallback options detailed below.**
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,95 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
 
+## Fetch Metadata headers
+
+Fetch Metadata request headers provide extra context about how an HTTP request was made, and how the resource will be used, enabling servers to reject suspicious cross-site requests. Servers can use these headers — most importantly `Sec-Fetch-Site` — as a lightweight and reliable method to block obvious cross-site requests. See the [Fetch Metadata specification](https://www.w3.org/TR/fetch-metadata/) for details.
+
+Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) using the `Origin` and `Referer` headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/?search=sec-fetch-site) in all browsers since March 2023 (with the exception of IE as it no longer receives updates and does not support `Sec-Fetch-*` headers).
```
nickchomey: Is referer actually necessary, or even a good idea? I've seen it mentioned in various places that it has some issues. And origin is essentially 100% browser coverage. Hence that's all Go 1.25 uses. I don't think we mention referer elsehwre, either.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,95 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+For the rare cases of outdated or embedded browsers that lack `Sec-Fetch-*` support, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) should provide the required coverage. If this is acceptable for your project, consider prompting users to update their browsers, as they are running on outdated and potentially insecure versions.
+
+### How to treat Fetch Metadata headers on the server-side
+
+`Sec-Fetch-Site` is the most useful Fetch Metadata header for blocking CSRF-like cross-origin requests and should be the primary signal in a Fetch-Metadata-based policy. Use other Fetch Metadata headers (`Sec-Fetch-Mode`, `Sec-Fetch-Dest`, `Sec-Fetch-User`) to further refine or tailor policies to your application (for example, allowing navigate mode top-level requests or permitting specific Dest values for resource endpoints).
```
nickchomey: is navigate mode defined anywhere?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,95 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+**Policy (high level)**
+
+1. If `Sec-Fetch-Site` is present:
+    1.1. Treat cross-site as untrusted for state-changing actions. By default, reject non-safe methods (POST / PUT / PATCH / DELETE) when `Sec-Fetch-Site: cross-site`.
```
nickchomey: This isn't really a comment about the doc, but a related question. 

In wordpress, Ajax requests are often made with GET to make mutations. This is, of course, wrong in many ways, but does happen.

Though, wp encourages the use of their nonce mechanism for Ajax, and doesn't use fetch metadata at all. Still, I will likely put in some effort to see if sec fetch can replace the wp nonces in a drop-in manner (just turn off/ignore nonces if they're there and use sec fetch)

So, I wonder what can be done to handle state-changing GET requests?

Should anything be mentioned here for the benefit of other frameworks or sites that use GET incorrectly?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,95 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+- Your application must be served over trustworthy URLs. Fetch Metadata request headers are only sent to [potentially trustworthy URLs](https://www.w3.org/TR/secure-contexts/#is-url-trustworthy). In practice, this includes `https`, `wss`, `file`, and `localhost` (including `127.0.0.0/8` and `::1/128`). See the [W3C Secure Contexts spec](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) for full details.
+- HTTPS must be enforced across the entire application. This ensures consistent inclusion of Fetch Metadata headers. Enabling [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) helps achieve this by automatically upgrading all HTTP requests to HTTPS.
+- [Safe HTTP methods](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) must not be used for state-changing requests.
+
+### Rollout & testing recommendations
+
+- Include an appropriate `Vary` header, in order to ensure that caches handle the response appropriately. For example, `Vary: Accept-Encoding, Sec-Fetch-Site`. See more [Fetch Metadata specification](https://w3c.github.io/webappsec-fetch-metadata/#vary).
```
nickchomey: Can we add any info on what the issue is here that is being addressed? It isn't obvious to me, and surely most others.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
nickchomey: Seems OK to me.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
jmanico: It's not obvious to me either

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
jmanico: State changing GET requests are very popular in Node since Express allow GET's with a payload by default.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
jmanico: A lot of legacy control docs say use referrer when origin is missing, like it was in Firefox for a long time.

But for 2025 and beyond, I think removing referrer references is more than reasonable.

mkhanas: I think I'll completely remove the  "`Origin` and `Referer`" part.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Accordingly to https://developer.mozilla.org/en-US/docs/Web/API/Request/mode there are 4 modes: `cors`, `no-cors`, `same-origin`, and `navigate`. But allow me to rephrase this into:
to further refine or tailor policies to your application's needs (for example, allowing top-level navigation requests or permitting specific Dest values for resource endpoints).

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,95 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+- Sec-Fetch-Mode, Sec-Fetch-Dest, Sec-Fetch-User — additional headers that provide context about the request (such as the request mode, destination type, or whether it was triggered by a user navigation). More details are available in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Glossary/Fetch_metadata_request_header).
+
+If any of headers above contain values not listed in the specification, in order to support forward-compatibility, servers should ignore those headers.
+
+### Ease of use
+
+Unlike [synchronizer tokens](#synchronizer-token-pattern) or [double-submit patterns](#alternative-using-a-double-submit-cookie-pattern) — which require additional client/server coordination and are easy to misimplement — Fetch Metadata checks are straightforward to implement correctly. They typically require only a small amount of server-side logic (inspect Sec-Fetch-Site, optionally refine with Sec-Fetch-Mode/Sec-Fetch-Dest) and no client changes. That simplicity reduces complexity, making the approach attractive for many applications.
```
smlx:
```suggestion
Unlike [synchronizer tokens](#synchronizer-token-pattern) or [double-submit patterns](#alternative-using-a-double-submit-cookie-pattern) — which require additional client/server coordination and are difficult to implement correctly — Fetch Metadata checks are much more straightforward. They typically require only a small amount of server-side logic (inspect Sec-Fetch-Site, optionally refine with Sec-Fetch-Mode/Sec-Fetch-Dest) and no client changes. That simplicity reduces complexity, making the approach attractive for many applications.
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: Good call! I guess we can easily add a note like:
“If your application relies on safe HTTP methods (GET, HEAD, or OPTIONS) for state‑changing actions, you should explicitly reflect that in your policy – e.g., by requiring a Fetch‑Metadata header review for requests to those endpoints. This can be enforced with a policy rule like the following example.”

And the policy example:
```js
const SAFE_METHODS = new Set(['GET','HEAD','OPTIONS']);
const SENSITIVE_ENDPOINTS = new Set([
  '/user/profile',
  '/account/details',
]);

const site = req.get('Sec-Fetch-Site');
const path = req.path;

// Block if cross-site + sensitive endpoint
if (site === 'cross-site' && SENSITIVE_ENDPOINTS.has(path)) {
  return false;
}

// Block if cross-site + unsafe method
if (site === 'cross-site' && !SAFE_METHODS.has(req.method)) {
  return false;
}
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: First of all, this recommendation comes directly from the [Fetch Metadata specification](https://w3c.github.io/webappsec-fetch-metadata/#vary) 

> If a given endpoint’s response depends upon the values the client delivers in a [Fetch metadata header](https://w3c.github.io/webappsec-fetch-metadata/#fetch-metadata-headers), developers should be careful to include an appropriate Vary header...

Regarding what is the issue here, I believe that James Kettle explained it best in the https://portswigger.net/research/exploiting-cors-misconfigurations-for-bitcoins-and-bounties: 

> Mostly things just break. However, in the right circumstances it can enable some quite serious attacks.

To make this clearer , we can add a note like:

> If the server responds differently based on HTTP headers (e.g., `Sec-Fetch-Site`, `Origin`), caches must vary on those headers. Without this, a CDNs or Proxies may reuse a response generated for a different context, causing broken behavior or contributing to cache-poisoning scenarios. Adding the appropriate `Vary` header ensures caches keep these responses separate

I'll also update `Vary: Accept-Encoding, Sec-Fetch-Site` to `Vary: Sec-Fetch-Site, Origin`

mkhanas: Updated

mkhanas:
Hey @nickchomey , as discussed above, I’ve added a Concerns section covering speculative navigation and intermediary behavior.

@jmanico, I wasn’t able to find the specific issues you referenced in your comments, so I kept the section more general without particular examples. If you think it would be useful to add those references, feel free to bring them here

jmanico (approved):

mackowski:
Looks good for me but I will also would like to wait for @nickchomey feedback

andrzejsydor (approved):

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -14,7 +14,8 @@ In short, the following principles should be followed to defend against CSRF:
 
 - **See the OWASP [XSS Prevention Cheat Sheet](Cross_Site_Scripting_Prevention_Cheat_Sheet.md) for detailed guidance on how to prevent XSS flaws.**
 - **First, check if your framework has [built-in CSRF protection](#use-built-in-or-existing-csrf-implementations-for-csrf-protection) and use it**
-- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state changing requests (requests that cause actions on the site) and validate them on the backend**
+- **If the framework does not have built-in CSRF protection, add [CSRF tokens](#token-based-mitigation) to all state-changing requests (requests that cause actions on the site) and validate them on the backend.**
+- **If your software is intended to be used only on modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) to block cross-site state-changing requests, so long as you use the fallback options detailed below.**
```
nickchomey:
```suggestion
- **If your software targets only modern browsers, you may rely on [Fetch Metadata headers](#fetch-metadata-headers) together with the fallback options described below to block cross-site state-changing requests.**
```

The original "so long as" language came from me, but I wonder whether this edit is cleaner and the style more appropriate for docs?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -25,13 +26,16 @@ In short, the following principles should be followed to defend against CSRF:
 - **Do not use GET requests for state changing operations.**
 - **If for any reason you do it, protect those resources against CSRF**
 
-## Token-Based Mitigation
+### Use Built-In Or Existing CSRF Implementations for CSRF Protection
```
nickchomey:
```suggestion
## Built-In Or Existing CSRF Implementations for CSRF Protection
```

Should this be a level 2 header, commensurate with the other alternatives, rather than embedded within the level 2 introduction? And the "Use" seems awkward for a header and not consistent with the other options.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
 
 Though the Naive Double-Submit Cookie method is simple and scalable, it remains vulnerable to cookie injection attacks, especially when attackers control subdomains or network environments allowing them to plant or overwrite cookies. For instance, an attacker-controlled subdomain (e.g., via DNS takeover) could inject a matching cookie and thus forge a valid request token. [This resource](https://owasp.org/www-chapter-london/assets/slides/David_Johansson-Double_Defeat_of_Double-Submit_Cookie.pdf) details these vulnerabilities. Therefore, always prefer the _Signed Double-Submit Cookie_ pattern with session-bound HMAC tokens to mitigate these threats.
 
+## Fetch Metadata headers
+
+Fetch Metadata request headers provide extra context about how an HTTP request was made, and how the resource will be used, enabling servers to reject suspicious cross-site requests. Servers can use these headers — most importantly `Sec-Fetch-Site` — as a lightweight and reliable method to block obvious cross-site requests. See the [Fetch Metadata specification](https://www.w3.org/TR/fetch-metadata/) for details.
```
nickchomey:
```suggestion
Fetch Metadata request headers provide extra information about the context from which an HTTP request was made. Servers can use these headers — most importantly `Sec-Fetch-Site` — as a lightweight and reliable method to block obvious cross-site requests. See the [Fetch Metadata specification](https://www.w3.org/TR/fetch-metadata/) for details.
```

I think this is more concise, as well as more in-line with what [mdn says](https://developer.mozilla.org/en-US/docs/Glossary/Fetch_metadata_request_header):

> header that provides additional information about the context from which the request originated. This allows the server to make decisions about whether a request should be allowed based on where the request came from and how the resource will be used

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+Because some legacy browsers may not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/?search=sec-fetch-site) in all browsers since March 2023 (with the exception of IE as it no longer receives updates and does not support `Sec-Fetch-*` headers).
```
nickchomey:
```suggestion
Because some legacy browsers do not send `Sec-Fetch-*` headers, a fallback to [standard origin verification](#using-standard-headers-to-verify-origin) headers **is a mandatory requirement** for any Fetch Metadata implementation. `Sec-Fetch-*` [is supported](https://caniuse.com/mdn-http_headers_sec-fetch-site) in all major browsers since March 2023.
```

The legacy browsers in concern DO not send, rather than MAY not. Alternatively, we could remove "some" and keep "may". I prefer the suggested edit.

Fixed the link to be direct rather than a search result 

Also, do we really need to provide ANY information about IE? As noted in the removed text, Microsoft itself doesn't even support it anymore, so it's no longer a legitimate browser. If OWASP isn't already making it explicit elsewhere that using IE is a significant red flag, they should be. Anyone using it still should know what they're doing and can see it's incompatibility in the link.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+The Fetch Metadata request headers are:
+
+- Sec-Fetch-Site — the primary signal for CSRF protection. It indicates relationship between request initiator’s origin and it's target's origin: `same-origin`, `same-site`, `cross-site`, or `none`.
```
nickchomey:
```suggestion
- Sec-Fetch-Site — the primary signal for CSRF protection. It indicates the relationship between the request initiator’s origin and its target's origin: `same-origin`, `same-site`, `cross-site`, or `none`.
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+If any of headers above contain values not listed in the specification, in order to support forward-compatibility, servers should ignore those headers.
+
+### Ease of use
+
+Unlike [synchronizer tokens](#synchronizer-token-pattern) or [double-submit patterns](#alternative-using-a-double-submit-cookie-pattern) — which require additional client/server coordination and are difficult to implement correctly — Fetch Metadata checks are much more straightforward. They typically require only a small amount of server-side logic (inspect Sec-Fetch-Site, optionally refine with Sec-Fetch-Mode/Sec-Fetch-Dest) and no client changes. That simplicity reduces complexity, making the approach attractive for many applications.
+
+### Browser compatibility
+
+Fetch Metadata request headers are supported in most modern browsers on both desktop and mobile (Chrome, Edge, Firefox, Safari 16.4+, and even in webviews on both iOS and Android), with [over 98% global coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site). For compatibility detail, see the [browser support table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).
```
nickchomey:
```suggestion
Fetch Metadata request headers are supported in all modern browsers on both desktop and mobile (Chrome, Edge, Firefox, Safari 16.4+, and even in webviews on both iOS and Android), with [over 98% global coverage](https://caniuse.com/mdn-http_headers_sec-fetch-site). For compatibility detail, see the [browser support table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site#browser_compatibility).
```
As discussed re:IE, if they aren't supported, they aren't really modern...

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+    
+    // Block if cross-site + sensitive endpoint
+    if (site === 'cross-site' && SENSITIVE_ENDPOINTS.has(path)) {
+      return false;
+    }
+    
+    // Block if cross-site + unsafe method
+    if (site === 'cross-site' && !SAFE_METHODS.has(req.method)) {
+      return false;
+    }
```
nickchomey:
```suggestion
    // Block if cross-site + unsafe method OR cross-site + sensitive endpoint
    if (site === 'cross-site' && (! SAFE_METHODS.has(req.method) || SENSITIVE_ENDPOINTS.has(path))) {
      return false; // forbid this request
    }
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+1. If `Sec-Fetch-Site` is present
+
+   1.1. Treat cross-site as untrusted for state-changing actions. By default, reject non-safe methods (POST / PUT / PATCH / DELETE) when `Sec-Fetch-Site: cross-site`.
+
+    ```JavaScript
+    const SAFE = new Set(['GET','HEAD','OPTIONS']);
+    const site = req.get('Sec-Fetch-Site'); // e.g. 'cross-site','same-site','same-origin','none'
+    
+    if (site === 'cross-site' && !SAFE.has(req.method)) {
```
nickchomey:
```suggestion
    const SAFE_METHODS = new Set(['GET','HEAD','OPTIONS']);
    const site = req.get('Sec-Fetch-Site'); // e.g. 'cross-site','same-site','same-origin','none'
    
    if (site === 'cross-site' && !SAFE_METHODS.has(req.method)) {
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+    1.3. Allow `same-origin`. Treat `same-site` as allowed only if your threat model trusts sibling subdomains; otherwise handle `same-site` conservatively (for example, require additional validation).
+
+    ```JavaScript
+    const trustSameSite = false; // set true only if you trust sibling subdomains
+    
+    if (site === 'same-origin') {
+      return true;
+    } else if (site === 'same-site') {
+      // handle same-site separately so the subcondition is clearly scoped to same-site
+      if (!trustSameSite && !SAFE.has(req.method)) {
```
nickchomey:
```suggestion
      if (!trustSameSite && !SAFE_METHODS.has(req.method)) {
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+      }
+      return true;
+    }
+    ```
+
+    1.4. Allow none for user-driven top-level navigations (bookmarks, typed URLs, explicit form submits) where appropriate.
+
+2. If `Sec-Fetch-*` headers are absent: choose a fallback based on risk and compatibility requirements:
+    2.1. Fail-safe (recommended for sensitive endpoints): treat absence as unknown and block the request.
+    2.2. Fail-open (compatibility-first): fallback to other security measure ([standard origin verification](#using-standard-headers-to-verify-origin), CSRF tokens, and/or require additional validation).
```
nickchomey:
```suggestion
    2.2. Fail-open (compatibility-first): fallback to ([standard origin verification](#using-standard-headers-to-verify-origin), CSRF tokens, and/or require additional validation).
```

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+### Requirements
+
+- Your application must be served over trustworthy URLs. Fetch Metadata request headers are only sent to [potentially trustworthy URLs](https://www.w3.org/TR/secure-contexts/#is-url-trustworthy). In practice, this includes `https`, `wss`, `file`, and `localhost` (including `127.0.0.0/8` and `::1/128`). See the [W3C Secure Contexts spec](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) for full details.
+- HTTPS must be enforced across the entire application. This ensures consistent inclusion of Fetch Metadata headers. Enabling [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) helps achieve this by automatically upgrading all HTTP requests to HTTPS.
+- [Safe HTTP methods](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) must not be used for state-changing requests.
```
nickchomey:
```suggestion
- [Safe HTTP methods](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP) should not be used for state-changing requests.
```

They shouldn't be, but they often are - as discussed above. But, if we change `must` to `should`, should we even have this "Requirement", since we discussed how to handle GET-based mutations?

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -151,6 +155,127 @@ Since an attacker is unable to access the cookie value during a cross-site reque
...
+
+### Concerns
+
+- Prerender/prefetch and other [speculative navigation](https://developer.mozilla.org/en-US/docs/Web/Performance/Guides/Speculative_loading) may send `Sec-Fetch-*` values that don’t match the final navigation, and browser-initiated flows (e.g., [PaymentRequest](https://developer.mozilla.org/en-US/docs/Web/API/Payment_Request_API)) could generate requests without predictable fetch-metadata headers. These behaviors are still being refined, so header propagation isn’t fully stable across all navigation types.
+- Intermediaries (proxies, gateways, load balancers) can remove or modify `Origin` and `Sec-*` headers — whether due to privacy filters, network optimizations, or simple misconfiguration — which can break fetch-metadata-based protections. This kind of header stripping is a problem on its own, but it still occurs surprisingly often in large enterprise networks, healthcare, ICS environments.
```
nickchomey:
```suggestion
- Intermediaries (proxies, gateways, load balancers) may remove or modify `Origin` and `Sec-*` headers — whether due to privacy filters, network optimizations, or simple misconfiguration — which can break fetch-metadata-based protections. This kind of header stripping is problematic, but common.
```

I don't think the extra information is useful or appropriate here.

nickchomey:
I think we're very close! I made various small suggestions for edits.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: I believe this section is tightly connected to the introduction, so promoting it to a level-2 header would break that flow. I'll remove the "Use" part

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: For me, it was a bit unusual to include this as a requirement, since it's more of a standalone recommendation and is already discussed elsewhere in the document. However, since it was specifically requested, I didn’t see any harm in adding it.

Let's keep it for now with `should`

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
mkhanas: I think everyone would agree that using IE is already a major concern on its own, and in the suggested wording “in all major browsers” is used, I personally wouldn’t consider IE a major browser anymore:)

I have no objections to removing this part.

mkhanas:
Accepted most of the suggestions, should be clear now.

cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
nickchomey: We could also use "modern" in place of "major", which I believe is a term used elsewhere in this doc (and probably others)

mkhanas:
Hi , @mackowski ! It looks like all questions and suggestions from the discussion have been resolved. When you have a moment, could you please take another look?

mackowski:
Thanks @mkhanas @nickchomey @jmanico @artis3n

Issue #1895, opened 2025-11-17
Update: Kubernetes Security Cheat Sheet


KadirArslan:
## What is missing or needs to be updated?
Kubernetes security cheat sheet is not mention the supply chain attacks. Which is very common. 

## How should this be resolved?
I'll fix this. Also I've additional improvement notes.

PR #1896, opened 2025-11-17
Update Kubernetes_Security_Cheat_Sheet.md

KadirArslan:
Kubernetes security cheat sheet is not mentioning one of the top 10 vulnerabilities for k8s, which is supply chain attacks / preventation mechanism. 
 
I fixed it. 

This PR fixes issue #1895.

### DIFF
--- a/cheatsheets/Kubernetes_Security_Cheat_Sheet.md
+++ b/cheatsheets/Kubernetes_Security_Cheat_Sheet.md
@@ -238,7 +238,7 @@ During the build phase, you should secure your Kubernetes container images by bu
 
 ### What is a container image?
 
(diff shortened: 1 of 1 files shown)


KadirArslan:
Also @jmanico @mackowski current format is OK, but I'd prefer seeing all top 10 vulnerability topics in the bullet-point, quick-reference style. IMO raw information that can be scanned quickly rather than lengthy explanations.

I can refactor this cheat sheet but I need time to spend on it. I've added this to my plans. 

wdyt on this? Should I spent some time on it? Refactor all of the CS to be more readable/actionable

szh (approved):
LGTM!

PR #1897, opened 2025-11-17
Update Zero_Trust_Architecture_Cheat_Sheet.md

KadirArslan:
I've added a new section about important Zero Trust practices:

- Policy-as-Code (PaC) 
- Continuous Verification
- Telemetry Signals

I think these are important things that we need to cover on CS.


mackowski:
Thanks @KadirArslan

PR #1960, opened 2026-01-12
Revise bcrypt work factor recommendation to 12

Xerodium:
Updated bcrypt work factor recommendation from 10 to 12 for legacy systems.

Most libraries by default set work factor to 12, and it is not effective anymore to rely on 1,024 iterations on a work factor of 10 as minimum

# You're A Rockstar

Please select one of the following options:

    and the prompt used is `[your prompt here]`. [Feel free to add more details if needed]

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -77,7 +77,7 @@ In this strategy, a password is hashed as usual using a password hashing algorit
 
 ### Using Work Factors
 
- The work factor is the number of iterations of the hashing algorithm that are performed for each password (usually, it's actually `2^work` iterations). The work factor is typically stored in the hash output. It makes calculating the hash more computationally expensive, which in turn reduces the speed and/or increases the cost for which an attacker can attempt to crack the password hash.
+The work factor is the number of iterations of the hashing algorithm that are performed for each password (usually, it's actually `2^work` iterations). The work factor is typically stored in the hash output. It makes calculating the hash more computationally expensive, which in turn reduces the speed and/or increases the cost for which an attacker can attempt to crack the password hash.
 
 When you choose a work factor, strike a balance between security and performance. Though higher work factors make hashes more difficult for an attacker to crack, they will slow down the process of verifying a login attempt. If the work factor is too high, the performance of the application may be degraded, which could be used by an attacker to carry out a denial of service attack by exhausting the server's CPU with a large number of login attempts.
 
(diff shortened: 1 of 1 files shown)


mackowski:
@jmanico what do you think?

jmanico:
I suggest a no here, but CC'ing in the expert behind this body of work @Sc00bz

Xerodium:
I'll leave the final verdict to you experts however my justification here is although 10 is a good number of iterations, I think the rise in available compute coupled with usually weaker passwords hashed when it comes to legacy systems makes me think 12 is a better standard and future proofs a bit more (not that people are out there cracking bcrypt easily right now) :)

Sc00bz:
> I think the rise in available compute

This is how it's already done. As GPUs get faster the number goes up. Why 12 and not 11? I find it odd that everyone suggests an even number when not backed by benchmarks. Even me, well before the current metrics. I think I kept my recommendation at cost 8 because it would soon be cost 8.

Anyway these are minimums such that a ~$1000 GPU\* gets <10 kH/s. Although "10 kH/s" was arbitrarily picked.

RTX 5090 benchmark of bcrypt cost 5 is [304.8 kH/s](https://gist.github.com/Chick3nman/09bac0775e6393468c2925c1e1363d5c#file-rtx_5090_cuda_v6-2-6-851-benchmark-L524-L528). An RTX 5090 is considered 2 "GPUs" because of cost. `304.8/2 = 152.4 kH/s/GPU`. Converting that to a cost that gets 10 kH/s you solve for `x` in `152.4 * (1+2*2^5) / (1+2*2^x) = 10`. You get `x ≈ 8.95`. So bcrypt cost 9 is the current minimum. The cheat sheet has cost 10 because of liking even numbers, but fixing this now is pointless because next gen GPUs will likely bump it to cost 10 (ie >157.41\*\* kH/s/GPU for cost 5 an increase of 3.290%).

P.S. You might be wondering about `1+2*2^cost`. That's the number of calls to Blowfish's key schedule. Once before the iterations and twice for every iteration. This ignores the encrypting of 512 bytes (8 bytes, 64 times) at the end which is 64/521 ("512/4168") the cost of a key schedule. So technically it should be `585/521+2*2^cost` then `x ≈ 8.95325` instead of `x ≈ 8.95070`.

\* Based on current GPUs and prices. It's possible that this will increase due to RAM shortages. It's historically been a high-end GPU but not super high-end because of diminishing returns. That changed with the RTX 40 series and the RTX 4090 as 1.5 "GPUs".

\*\* Used the more accurate `585/521+2*2^cost`. For future reference:
```
Minimum: Benchmarked "GPU" with cost 5
cost 10:    >157.41 kH/s
cost 11:    >314.66 kH/s
cost 12:    >629.14 kH/s
cost 13:  >1,258.10 kH/s
cost 14:  >2,516.03 kH/s
cost 15:  >5,031.89 kH/s
cost 16: >10,063.62 kH/s
```
It would be better to benchmark the GPU with the target cost because this will increase the speed a little because it wont need to generate as many passwords. But this is good enough.

Xerodium:
Wow @Sc00bz what a writeup. Tbh I think that is me shown and looks like 10 is fine based on that math. I'll close this thank you all!

jmanico:
@Sc00bz is a legend. He sword is **_math_** and his shield is **_detailed explanations_** and we are all very lucky to have his support.

Issue #1742, opened 2025-08-06
Update: Password_Storage_Cheat_Sheet


Sc00bz:
## What is missing or needs to be updated?
https://github.com/OWASP/CheatSheetSeries/blob/9f9424ae0237ae3f21ab24e8ea98c9ef243bb01d/cheatsheets/Password_Storage_Cheat_Sheet.md?plain=1#L118

> the minimum CPU/memory cost parameter (N), the blocksize (r) and the degree of parallelism (p)

This should be changed to either:

"the minimum CPU/memory cost parameter (N), the blocksize (r) and the CPU cost/parallelism (p)"

or

"the minimum memory cost parameter (N), the blocksize (r) and the degree of parallelism (p)"

Since computational cost are proportional to both N and p. So either add "CPU" to p or remove it from N. There's likely a better way to word "the CPU cost/parallelism (p)".

https://github.com/OWASP/CheatSheetSeries/blob/9f9424ae0237ae3f21ab24e8ea98c9ef243bb01d/cheatsheets/Password_Storage_Cheat_Sheet.md?plain=1#L126

> provide an equal level of defense

Maybe change to "provide a minimal level of defense" or maybe "provide a similar minimal level of defense", but I feel "similar" is redundant.

jmanico:
Steve, I trust your instinct here. Would you like me to take this on or would you like to submit a PR?

Prasad-JB:
Hi @Sc00bz, I’d like to work on this issue.
I’ve reviewed the section in Password_Storage_Cheat_Sheet.md and understand the wording changes needed for (N), (r), and (p).
Could you please assign me so I can submit a PR?

Thanks!

jmanico:
> Hi [@Sc00bz](https://github.com/Sc00bz), I’d like to work on this issue. I’ve reviewed the section in Password_Storage_Cheat_Sheet.md and understand the wording changes needed for (N), (r), and (p). Could you please assign me so I can submit a PR?

Done!

Shubb07:
So is this issue been solved or is it yet to be solved , you can assign it to me if it isn't .

gamila-wisam:
Hi @jmanico,
I’ve created a PR to address this issue by updating the scrypt section in Password_Storage_Cheat_Sheet.md:
- Clarified scrypt parameters (N, r, p) on line 118 for accuracy.
- Updated defense level wording on line 126 to "provide a similar minimal level of defense" and clarified the CPU/RAM trade-off.

[...]

PR #2002, opened 2026-02-09
Update scrypt section wording in Password_Storage_Cheat_Sheet.md for …

gamila-wisam:
This PR updates the scrypt section in Password_Storage_Cheat_Sheet.md for clarity and technical accuracy:

- Line 118: Clarified scrypt parameters (N, r, p) to reflect proper memory and parallelism usage.
- Line 126: Updated wording about defense level to "similar minimal level of defense" and clarified parallelism/RAM trade-off.

This PR fixes issue #1742.

I have NOT used any AI tool to generate the contents of this PR.

### DIFF
--- a/cheatsheets/Password_Storage_Cheat_Sheet.md
+++ b/cheatsheets/Password_Storage_Cheat_Sheet.md
@@ -123,7 +123,7 @@ Like [Argon2id](#argon2id), scrypt has three parameters that can be configured: 
 - N=2^14 (16 MiB), r=8 (1024 bytes), p=5
 - N=2^13 (8 MiB), r=8 (1024 bytes), p=10
 
-These configuration settings provide a similar minimal level of defense, with the main trade-off between CPU and RAM usage.
+These configuration settings provide a similar minimal level of defense, with the main trade-off between parallelism and RAM usage.
 
 ### bcrypt
 


jmanico (approved):
Nice fix and clarification

jmanico:
cc @Sc00bz

gamila-wisam:
Thank you @jmanico!

Sc00bz:
This is better that it was.

There is still the "... trade-off between CPU and RAM usage." In an ideal world, they'd all use the same amount of compute. The only minor difference in compute is due to the settings needing to be integers. For `N=2^13 (8 MiB), r=8 (1024 bytes), p=10`, let's say `p` is actually 9.5 (round up thus 10). The next one, the memory is doubled and `p` halved so 4.75 (thus 5). Then 2.375 (3), 1.1875 (2), 0.59375 (1).

I guess if it was "... trade-off between parallelism and RAM usage." But one could say "CPU" includes CPU cores thus parallelism. When I read CPU in that sentence, I think compute. Maybe that's just me. Anyway 👍.

gamila-wisam:
HI @Sc00bz, Thanks its Updated now!

mackowski (approved):

Issue #2008, opened 2026-02-14
Update: HTTP_Headers_Cheat_Sheet


Brandon-Johns:
## What is missing or needs to be updated?

This relates to issue https://github.com/OWASP/CheatSheetSeries/issues/1401

The section `Adding HTTP Headers in Different Technologies` > `Apache` states that the `always` option should be used because "without the always option, the header will only be sent for certain status codes".

However, the [Apache documentation](https://httpd.apache.org/docs/2.4/mod/mod_headers.html#header) also states that

> always is not a superset of onsuccess with respect to existing headers

The documentation then elaborates more on this point with detailed description and an example.

The documentation also states that it can sometimes make sense to repeat the `Header set` command with both `onsuccess` (default, can be omitted) and `always`. I'm not particularly experianced in Apache configuration, but I can't see any downside to doing this for as a general rule for the security related headers. It seems that it would be good practice as a means to cover the edge cases.

## How should this be resolved?

I suggest that the cheat sheet be updated
- to warn that `always` is not a superset of `onsuccess`
- to suggest repeating the header set / unset commands both with and without `always`

The example would then become

```apache
<IfModule mod_headers.c>
Header set X-Frame-Options "DENY"
Header always set X-Frame-Options "DENY"
</IfModule>
```

himasree9:
Hi @Brandon-Johns, 
I would like to work on this update.
I can review the Apache documentation carefully and propose a clarified section including warnings about always vs onsuccess.
Please let me know if I can proceed.

Brandon-Johns:
Hi @HimasreeKolathur24 
I'm not a manager of this repository, but I assume that your work on this would be most welcome.
Please feel free to proceed.
As first step, we should discuss it here.

My progress on it so far:

I found the [MDN apache configuration recommendations](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Server-side/Apache_Configuration_htaccess#remove_the_x-powered-by_response_header), which uses the `always` option in a few places.

Having looked at it and thought about it more, I currently suspect the meaning to be
[...]

PR #2012, opened 2026-02-17
Clarify Apache Header always vs onsuccess behavior to prevent duplication (#2008)

HimasreeKolathur24:
## Summary

This PR clarifies the behavior of `Header set` (default `onsuccess`) and `Header always set` in Apache `mod_headers`.

Previously, the section mentioned the `always` option but did not explain that it operates on a separate internal header table. On 2xx responses, both `onsuccess` and `always` directives may apply, which can result in duplicate headers if the same header is configured in both contexts.

This update:
- Adds a short clarification about the separation of header tables.
- Warns about potential header duplication.
- Recommends a safe configuration pattern using `Header unset` before `Header always set`.

The goal is to prevent misconfiguration while keeping the guidance concise and practical.

This PR fixes issue #2008.

---

## AI Tool Usage Disclosure

- [x] I have used AI tools to generate the contents of this PR. I have verified the contents and I affirm the results. The LLM used is `ChatGPT (GPT-5.2)` and the prompt used was assistance in clarifying Apache mod_headers behavior and improving documentation wording.

### DIFF
--- a/cheatsheets/HTTP_Headers_Cheat_Sheet.md
+++ b/cheatsheets/HTTP_Headers_Cheat_Sheet.md
@@ -300,11 +300,20 @@ header("X-Frame-Options: DENY");
 
 ### Apache
 
-Below is an `.htaccess` sample configuration which sets the `X-Frame-Options` header in Apache. Note that without the `always` option, the header will only be sent for certain status codes, as described in [the Apache documentation](https://httpd.apache.org/docs/2.4/mod/mod_headers.html#header).
+Below is an `.htaccess` sample configuration which sets the `X-Frame-Options` header in Apache.
+
+As described in the [Apache documentation](https://httpd.apache.org/docs/2.4/mod/mod_headers.html#header), `Header set` (default `onsuccess`) and `Header always set` operate on separate internal header tables.
+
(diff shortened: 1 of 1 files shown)


Brandon-Johns:
Thank you for working on this.
It's good.

I think it could be slightly improved with
- Restore the link to the Apache documentation to enable further reading, in case that the solution does not meet the needs of the reader (if they need to apply different values of the headers for different status codes).
- Add a note that to unset a header, it should be unset for both `onsuccess` and `always`.
- Given the edge cases, as I discussed in the issue thread, I suggest to alter the sentance
    - From: `On 2xx responses, both may apply, which can result in duplicate headers if the same header is configured in both contexts.`
    - To: `In some cases, both header tables may be used, which can result in duplicate headers if the same header is configured in both contexts.`

himasree9:
@Brandon-Johns,
Thank you for the feedback. I’ve updated the section to restore the documentation link, adjust the wording, and clarify unsetting in both contexts. Please let me know if anything else should be refined.

Brandon-Johns (approved):
This it very clear and well stated.
I fully approve of the changes in this form
Thank you for your work on it

jmanico (approved):

mackowski (approved):

PR #2051, opened 2026-03-05
Add Node.js example for login throttling

YK-03:
# You're A Rockstar

- [ ] In case of a new Cheat Sheet, you have used the Cheat Sheet template.
- [x] All the markdown files do not raise any validation policy violation, see the policy.
- [x] All the markdown files follow these format rules.
- [ ] The CI build of your PR pass, see the build status here.

## Description

This PR adds a small Node.js / Express example demonstrating how login rate limiting can be implemented under the **Login Throttling** section of the Authentication Cheat Sheet.

The goal is to provide developers with a practical implementation example showing how rate limiting can help mitigate brute-force and credential stuffing attacks.

The example is framework-agnostic in concept while using Express only as a demonstration.

- [x] I have used AI tools to assist with drafting the example and formatting the contribution. The contents were reviewed and validated before submission.

### DIFF
--- a/cheatsheets/Authentication_Cheat_Sheet.md
+++ b/cheatsheets/Authentication_Cheat_Sheet.md
@@ -283,6 +283,21 @@ Rather than implementing a fixed lockout duration (e.g., ten minutes), some appl
 
 When designing an account lockout system, care must be taken to prevent it from being used to cause a denial of service by locking out other users' accounts. One way this could be performed is to allow the use of the forgotten password functionality to log in, even if the account is locked out.
 
+##### Example Implementation (Node.js / Express)
+
+A simple example of login rate limiting can be implemented using middleware such as `express-rate-limit`.
+
+```javascript
+const rateLimit = require("express-rate-limit");
+
+const loginLimiter = rateLimit({
+  windowMs: 15 * 60 * 1000,
+  max: 10,
+  message: "Too many login attempts. Please try again later."
+});
+
+app.post("/login", loginLimiter, loginHandler);
+
 #### CAPTCHA
 
 The use of an effective CAPTCHA can help to prevent automated login attempts against accounts. However, many CAPTCHA implementations have weaknesses that allow them to be solved using automated techniques or can be outsourced to services that can solve them. As such, the use of CAPTCHA should be viewed as a defense-in-depth control to make brute-force attacks more time-consuming and expensive, rather than as a preventative.


jmanico (approved):

jmanico:
I accepted it because its a clean PR, but can other reviewers consider if we want a proof of concept code sample like this?

YK-03:
Thanks for the review and approval!

Happy to adjust the example if needed. If reviewers prefer a more generic or language-agnostic example instead of Node.js specifically, I can update the section accordingly.

randomstuff:
There are caveats which are worth highlighting.

This does not work so well when using more than once instance (or more than one Node.js worker eg. when using the cluster module). You might want to use some Redis, KeyDB and friends?

Alternatively, consider implementing rate limiting at the reverse proxy / WAF?

randomstuff:
(Also, it does not work so well against [highly distributed credentials stuffing attacks](https://datadome.co/threat-research/tale-of-a-heavily-distributed-credential-stuffing-attack/) but this is difficult to protect against without risking locking the user out of his account.)

jmanico:
I am politely closing this PR. I will take starting over to make this code sample work.

kwwall:
@jmanico wrote:
> I am politely closing this PR. I will take starting over to make this code sample work.

I looked at this a bit the other day and unless one were to perhaps separate the login page into two separate HTTP requests as some sites do (e.g., first request to enter username, and 2nd to enter creds), so unless I'm missing something I'm not even sure it would this rate limiting approach would be very effective. If the username and credentials all are entered via a single form, what's to say that a potential attacker just wouldn't forge the source IP address since for a DoS attack to lockout a user to succeed, they needn't wait for any response. (I suppose if account  is written such that successive failures much come from the same IP address, it might be a bit effective, but most account lockup configurations that I've reviewed don't ever consider that.) It least if one were to separate the form into 2 steps, this would slow things down a bit because they would have to supply a legitimate source IP to wait until the response came back to enter the credentials (typically a password, or maybe a password and TOTP token). My guess is that the express-rate-limiting was designed for other things like REST APIs where account lockout is less likely to apply but you don't want the API itself to be DoS'd.

Just my $.02. I have admittedly not bothered to look at the express-rate-limiting code.

YK-03:
Thanks for the detailed feedback and discussion.

I appreciate the insights around multi-instance deployments, distributed credential stuffing, and the limitations of simple rate limiting approaches. It makes sense that a more carefully designed example would be needed for the cheat sheet.

Looking forward to the revised version and happy to help review or contribute further if needed.

Issue #2037, opened 2026-02-27
Email Validation and Verification in Identity Systems Cheat Sheet


jeremyrayjewell:
Problem Statement:
Developers frequently mis-handle email validation, canonicalization, and verification flows in identity systems, leading to account takeover and enumeration risks.

Why This Needs Its Own Cheat Sheet:
Authentication CS is large.
Input Validation CS does not address identity specific email handling risks.

Proposed Scope:
Email format validation pitfalls
Canonicalization and normalization
Unicode and IDN concerns
Case sensitivity handling
Email ownership verification flows
Password reset threat model
Email change workflows
Anti enumeration controls
Temporary email abuse
Email as weak factor risks

jeremyrayjewell:
Requesting NEW_CS label for this proposal. Thank you.

mackowski:
@jeremyrayjewell this is great idea!
@jmanico @szh what do you think, should we have new cheatsheet or adjust Authentication CS? 

Also question to all, if we go with new CS, do we see other cheatsheets that should reference this new one?

jmanico:
I vote new cheatsheet! :)

mackowski:
@jeremyrayjewell go for it :) we are waiting for PR

jeremyrayjewell:
PR submitted: #2072

Initial draft of the Email Validation and Verification Cheat Sheet based on the agreed scope.

Happy to iterate based on feedback.

mackowski:
Closing as resolved — the Email Validation and Verification Cheat Sheet was created and merged in PR #2072. Thanks for the proposal!

PR #2072, opened 2026-03-19
Add Email Validation and Verification Cheat Sheet (initial draft)

jeremyrayjewell:
# Email Validation and Verification in Identity Systems Cheat Sheet

Initial draft based on accepted proposal #2037.

This cheat sheet covers:
- Email canonicalization and normalization
- Validation pitfalls
- Unicode and IDN concerns
- Verification flows (signup, reset, change)
- Anti-enumeration controls
- Email as a weak authentication factor

This PR fixes issue #2037.

---

## Checklist

- [x] In case of a new Cheat Sheet, you have used the Cheat Sheet template.
- [x] All the markdown files do not raise any validation policy violation.
- [x] All the markdown files follow format rules.
- [x] All your assets are stored in the assets folder.
- [x] All the images used are in PNG format.
- [x] You verified/tested the effectiveness of your contribution.
- [ ] The CI build of your PR pass.

---

## AI Tool Usage Disclosure

  the contents and I affirm the results. The LLM used is `GPT-5.3`.

### DIFF
--- a/cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md
+++ b/cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md
@@ -2,13 +2,9 @@
 
 ## Introduction
 
-Email addresses are widely used as primary identifiers in authentication
-and account recovery workflows. Improper handling of email validation,
-normalization, and verification can lead to account takeover, user
-enumeration, and identity confusion.
+Email addresses are widely used as primary identifiers in authentication and account recovery workflows. Improper handling of email validation, normalization, and verification can lead to account takeover, user enumeration, and identity confusion.
 
-This cheat sheet provides guidance on securely handling email addresses
-within identity systems.
+This cheat sheet provides guidance on securely handling email addresses within identity systems.
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md
```diff
@@ -0,0 +1,154 @@
...
+
+Monitoring email-related flows helps detect abuse.
+
+### Recommendations
+
+- Log verification attempts and failures
+- Monitor password reset activity
+- Detect abnormal patterns (e.g., high-frequency requests)
+
+## References
```
mackowski, line 148: Can you make all references to be a links?

jmanico:
This is a useful start, but I see a few issues that should be addressed before merge:

1. cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md:68 recommends treating the entire email address as case-insensitive for identity purposes. That is too broad as written. The domain part is case-insensitive, but the local part can be case-sensitive under SMTP, and internationalized local-parts introduce additional edge cases. A blanket rule here can create account-collision or mistaken-account behavior, especially when identities are imported or federated. I recommend preserving the original address, defining an explicit comparison policy, and only folding the local-part when the system fully owns that behavior.
2. cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md:145 covers logging and monitoring, but it does not mention the risk of logging full email addresses, reset URLs, or verification/reset tokens. That is an important omission for this topic because it can create both PII leakage and account-takeover exposure through logs.
 3. Can you convert all references into links, per mackowski’s feedback?
4. Please fix the lint errors.
5. I also think there are too many hard line breaks in the document. In several places, the text wraps where I would expect the sentence or paragraph to continue naturally. Please take another pass on formatting and smooth that out.

This is headed in the right direction, but I’d like to see those items cleaned up before approval.

jeremyrayjewell:
Updated in 63456128.

I converted the references to links, fixed the nested-list indentation issues causing markdown lint failures, added guidance to avoid logging full email addresses and never log verification/reset tokens or URLs, and revised the case-sensitivity section to preserve the original address and require an explicit local-part comparison policy instead of treating the entire address as case-insensitive.

I also cleaned up the formatting to reduce unnecessary hard line breaks.

`npm run lint-markdown -- cheatsheets/Email_Validation_and_Verification_Cheat_Sheet.md` now passes locally.

PR #2085, opened 2026-03-27
Add guidance for secure handling of user file downloads

### LINKED ISSUE #2082: Update: HTTP_Headers_Cheat_Sheet
Serving user-provided files, such as attachments, without special headers can lead to various security vulnerabilities, including Cross-Site Scripting (XSS) and Content Security Policy (CSP) bypass. This is because browsers execute these files by default. Currently, the cheat sheet only covers the `X-Content-Type-Options` header and partially the `Content-Type` header.

1. Include the `Content-Disposition` header with a recommendation to use `Content-Disposition: attachment` for serving user-provided files. According to the [RFC](https://httpwg.org/specs/rfc6266.html#rfc.section.4.2), `attachment` should prompt the user to save the response locally, rather than process it normally. Without the header, a browser may default to [`inline` mode](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body), which typically processes the content.
2. Suggest using `Content-Type: application/octet-stream` for serving user-provided files. This may prevent a browser from executing a file. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#applicationoctet-stream):
> application/octet-stream
> This is the default for binary files. As it means unknown binary file, browsers usually don't execute it, or even ask if it should be executed. They treat it as if the [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition) header was set to attachment, and propose a "Save As" dialog.
> ...
> Most web servers send unrecognized resources as the application/octet-stream MIME type. For security reasons, most browsers do not allow setting a custom default action (like "Open in Word") for such resources, forcing the user to save it to disk to use it.

Using these headers with `X-Content-Type-Options` should prevent a browser from processing user-provided files.

ritesh-singh-1595:
This PR adds a small section on securely serving user-provided files.

It includes recommendations for using Content-Disposition, Content-Type, and X-Content-Type-Options headers to prevent unintended execution and reduce security risks.

The goal is to improve clarity around safe file download practices.

Closes #2082

### DIFF
--- a/cheatsheets/HTTP_Headers_Cheat_Sheet.md
+++ b/cheatsheets/HTTP_Headers_Cheat_Sheet.md
@@ -288,6 +288,16 @@ The HTTP `Public-Key-Pins` response header is used to associate a specific crypt
 
 This header is deprecated and should not be used anymore.
 
+## Secure File Download Headers
+
+When serving user-provided files, proper headers should be used to prevent unintended execution.
+
+- Use `Content-Disposition: attachment` to force download instead of inline rendering.
+- Use `Content-Type: application/octet-stream` for unknown or binary files.
+- Ensure `X-Content-Type-Options: nosniff` is set to prevent MIME type sniffing.
+
+These headers help reduce risks such as Cross-Site Scripting (XSS) and unintended file execution.
+
 ## Adding HTTP Headers in Different Technologies
 
 ### PHP


mackowski:
Please follow https://github.com/OWASP/CheatSheetSeries/blob/master/CONTRIBUTING.md

PR #2088, opened 2026-03-29
Add Cache-Control guidance for secure handling of sensitive content

### LINKED ISSUE #2086: Update: HTTP_Headers_Cheat_Sheet
The `Cache-Control` header, as its name implies, regulates the caching behavior of shared caches (intermediate nodes, such as proxies) and private caches (endpoints, such as browsers). The old [RFC 7234](https://datatracker.ietf.org/doc/html/rfc7234#section-3) and the new [RFC 9111](https://datatracker.ietf.org/doc/html/rfc9111#section-3) define three primary conditions that prevent a response from being stored in a cache: the `no-store` cache directive (applicable to any cache), the `private` cache directive (applicable to shared caches), and the `Authorization` header (applicable to shared caches). Otherwise, under certain conditions, a cache may heuristically classify a response as cacheable. For instance, if a private endpoint is protected by a measure that differs from the `Authorization` header (such as cookies), a response without proper `Cache-Control` settings may be stored in a cache. This poses three risks: storing sensitive content in a shared cache, unauthorized access to protected data via a cache, and cache poisoning.
Additionally, one should not confuse the `no-cache` with `no-store` extension, since `no-cache`, despite its naming, permits caching, but requires revalidation of a response.

Add a description for the `Cache-Control` header, suggest using `no-store` or `private` for protected content.

ritesh-singh-1595:
This PR adds a new section for the Cache-Control header in the HTTP Headers cheat sheet.

It includes guidance on using no-store and private directives when handling sensitive or user-specific data, along with a reference to Mozilla documentation.

The goal is to improve clarity around caching-related security risks and make the cheat sheet more complete.

Closes #2086

### DIFF
--- a/cheatsheets/HTTP_Headers_Cheat_Sheet.md
+++ b/cheatsheets/HTTP_Headers_Cheat_Sheet.md
@@ -75,21 +75,22 @@ Although it is recommended to always set the `Content-Type` header correctly, it
 - *NOTE:* the `charset` attribute is necessary to prevent XSS in **HTML** pages
 - *NOTE*: the `Content-Type` can be any of the possible [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
 
-## Cache-Control
+### Cache-Control
 
 The `Cache-Control` header defines how responses are cached by browsers and intermediate caches.
 
-### Recommendation
+#### Recommendation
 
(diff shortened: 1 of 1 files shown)


ritesh-singh-1595:
@maintainers I’ve addressed all review comments and updated the PR. Could you please re-review and approve the workflows? Thanks!

PR #449, opened 2020-07-21
Issue #438 - Improve XSS cheatsheet to address escape/encode confusion

cdwijayarathna:
I have changed the XSS prevention cheatsheet based on the discussion we had at issue #438 .

Basically, I removed word 'escape' where it does not follow the definition at  https://owasp.org/www-project-proactive-controls/v3/en/c4-encode-escape-data. 

I left a few references to 'escape' as well where I felt it follows the above definition.

Open for discussion

This PR covers issue #438

PR #2112, opened 2026-04-13
feat: add OpenCRE reference to Denial of Service Cheat Sheet

Abhijeet2409:
This PR adds a reference to the OpenCRE project in the Denial of Service Cheat Sheet.

### Context
The Denial of Service Cheat Sheet does not include a reference to the OpenCRE project, which prevents it from being mapped and discovered via OpenCRE.

### Changes
Added a link to the relevant CRE (Denial of Service protection - CRE 623-550) under the Introduction section, following the format used in other cheat sheets (e.g., Secrets Management).

### Result
This enables proper mapping in OpenCRE and improves discoverability.

Related issue: https://github.com/OWASP/OpenCRE/issues/449

### DIFF
--- a/cheatsheets/Denial_of_Service_Cheat_Sheet.md
+++ b/cheatsheets/Denial_of_Service_Cheat_Sheet.md
@@ -4,6 +4,8 @@
 
 This cheat sheet describes a methodology for handling denial of service (DoS) attacks on different layers. It also serves as a platform for further discussion and analysis, since there are many different ways to perform DoS attacks.
 
+See [the Open CRE project on denial of service](https://www.opencre.org/cre/623-550) for additional technical requirements and mappings.
+
 ### Fundamentals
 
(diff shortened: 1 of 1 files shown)


Bornunique911 (commented):
LGTM 👍 — this is a useful addition for OpenCRE traceability.

Suggestion: per OpenCRE Issue [#449](https://github.com/OWASP/OpenCRE/issues/449) references should be placed close to relevant controls rather than after the Introduction. Could we move this to a section like **Network Architecture/Design** or **Rate Limiting**, similar to how it’s done in the Secrets Management Cheat Sheet?

Optionally, we can still keep it in Related Articles as well.

Abhijeet2409:
Hi @jmanico, could you please take a look at this PR when you have time?

jmanico:
Folk, references do not go into the top of cheatsheets, and I do not see this as a high value or clear resource. We really want to keep our cheatshets as lean as possible. I'm sorry but I think its best to close this.

Abhijeet2409:
Thanks for the clarification, @jmanico — that makes sense. I understand your point about keeping the cheatsheets lean and focused.
I appreciate the guidance!

Issue #1101, opened 2023-03-08
Update: CSRF prevention cheat sheet to offer more detail on SameSite cookie limitations


madelson:
## What is missing or needs to be updated?

With regards to SameSite cookies, the cheat sheet says this:

> It is important to note that this attribute should be implemented as an additional layer defense in depth concept. This attribute protects the user through the browsers supporting it, and it contains as well 2 ways to bypass it as mentioned in the following [section](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7.1). This attribute should not replace having a CSRF Token. Instead, it should co-exist with that token in order to protect the user in a more robust way.

The 2 ways to bypass described in the linked document are:

> 1.  Attackers can still pop up new windows or trigger top-level
       navigations in order to create a "same-site" request (as described in [section 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-2.1)), which is only a speedbump along the
       road to exploitation.

>    2.  Features like `<link rel='prerender'>` [[prerendering](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#ref-prerendering)] can be
       exploited to create "same-site" requests without the risk of user
       detection.

However, other commentary I've found (e.g. [this SO post](https://security.stackexchange.com/questions/263135/are-these-samesite-vulnerabilities-still-present-today)) suggests that those bypass approaches might be dated and/or limited in scope.

## How should this be resolved?

It would be nice if the cheat sheet addressed these limitations in more detail. Can SameSite be a first-class mitigation mechanism under certain scenarios (e.g. no domain sharing, no mutating GET requests)? What attacks are still possible?

advename:
To my knowledge:

### `SameSite=Lax`
`Lax` provides a reasonable balance between security and usability for websites that want to maintain a user's logged-in session after the user arrives from an external link. `Lax` will send the cookie in cross-site requests, but only if both of the [following conditions are met](https://portswigger.net/web-security/csrf/bypassing-samesite-restrictions#lax):
- The request uses a ["safe" HTTP method](https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1) (`GET`, `HEAD`, `OPTIONS` and `TRACE`) ([1](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7.1))
[...]

PR #2118, opened 2026-04-16
CSRF: expand SameSite limitations and clarify when it can stand alone

0xBassia:
Addresses #1101

Replaces the short paragraph about "two ways to bypass" `SameSite` with practical guidance on the real-world gaps developers should know about, plus a narrow list of conditions under which `SameSite` can serve as a primary CSRF defense.

### What changed

In the **SameSite (Cookie Attribute)** section, after the existing intro and examples, the final paragraph was replaced with two new subsections:

**`#### Limitations of SameSite`** covers:

- `Lax` only blocks unsafe methods, so any state-changing `GET` endpoint is unprotected (the single most common failure in practice).
- `SameSite` is scoped to the registrable domain, not the origin. Sibling hosts and sub-domains you don't fully control can still issue same-site requests. This also amplifies the impact of subdomain takeovers (with a link to the Subdomain Takeover Prevention CS).
- Top-level navigation and window-opening tricks.
- Uneven browser coverage on legacy/embedded clients.
- `SameSite` does not protect against client-side CSRF.

**`##### When SameSite May Be Sufficient On Its Own`** gives the narrow set of conditions (no shared registrable domain, no state-changing safe-method endpoints, `Strict` or `Lax` + `__Host-` prefix, origin/referer check for defense in depth) under which teams can reasonably rely on `SameSite` without a token, per @jmanico's feedback in the issue discussion.

### Discussion context

@jmanico (core team) confirmed in the issue that the existing two-bypass paragraph was outdated and invited a PR: _"Would you like to give us a PR that addresses your points?"_ The issue author (@madelson) identified the two specific gaps that most deployments hit (shared registrable domain and state-changing GETs), and this PR codifies both plus the adjacent concerns into the cheat sheet.

### Scope

- Single file modified.
- No content removed beyond the outdated paragraph; the token-based recommendations elsewhere in the cheat sheet are unchanged.
- Pre-existing StackOverflow links flag as 403 under the link-checker due to bot detection; none of those are from this change.

### Checklist

- [x] `npm run lint-markdown` clean
- [x] `npm run lint-terminology` clean
- [x] No new dead links introduced
- [x] US English
- [x] Single file, architectural guidance, no code changes

### DIFF
--- a/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
+++ b/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md
@@ -419,7 +419,27 @@ Set-Cookie: JSESSIONID=xxxxx; SameSite=Lax
 
 All modern desktop and mobile browsers support the `SameSite` attribute. The main exceptions are legacy browsers including Opera Mini (all versions), UC Browser for Android, and older mobile browsers (iOS Safari < 13.2, Android Browser < 97). To track the browsers implementing it and know how the attribute is used, refer to the following [service](https://caniuse.com/#feat=same-site-cookie-attribute). Chrome implemented `SameSite=Lax` as the default behavior in 2020, and Firefox and Edge have followed suit. Additionally, the `Secure` flag is required for cookies that are marked as `SameSite=None`.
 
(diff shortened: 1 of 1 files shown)


jmanico (approved):
This is excellent, concise and deep. Perfect.

PR #2210, opened 2026-05-29
Add AML and Sanctions Compliance for AI Agent Payments Cheat Sheet

razashariff:
@mackowski see updated PR -- I had already written this weeks ago, needed to update the links and cross-reference checks against the other OWASP alignments which are now done.

Thanks,
Raza

---

## Summary

This PR adds the cheat sheet proposed in #2125, drafted to the structure approved on 26 April 2026.

The cheat sheet covers Anti-Money Laundering (AML) and sanctions screening controls specifically for AI agent payments, including:

- Cryptographic agent identity verification before screening
- Entity screening against OFAC, UK HMT, EU, and UN sanctions lists
- Agent operator screening and KYB requirements
- Tamper-evident, hash-chained signed audit trails
- Fail-closed enforcement
- Trust-tiered rate limiting against cryptographic identity
- Self-hosted vs hosted screening architecture trade-offs
- Consolidated Do's and Don'ts

## Relationship to existing OWASP material

The cheat sheet is complementary to and consistent with:

- The OWASP MCP Security Cheat Sheet (Section 7 message-level integrity), already published.
- The OWASP MCP Top 10 (MCP01, MCP07, MCP08).
- The OWASP AISVS Chapter 10 MCP Security Requirements (10.2.13, 10.4.11, 10.6.4).
- The OWASP API Security Top 10 API4:2023 (Unrestricted Resource Consumption), used as the basis for trust-tiered rate limiting.

The cryptographic stack used throughout (ECDSA P-256 signing, SHA-256 hash-chained audit entries, message-level signatures) is consistent with the existing OWASP MCP Top 10 Cryptographic Security Layer control to keep a single coherent crypto posture across the OWASP AI-agent material.

## Sources cited

Each H2 section includes at least one primary-source inline citation, per the contribution guidelines tightened in #2174. Sources cited include FinCEN, OFAC, UK OFSI, EU EUR-Lex, OCC, NIST SP 800-53 r5, NIST SP 800-209, OWASP MCP Cheat Sheet, OWASP MCP Top 10, OWASP API Security Top 10, OWASP AISVS, and the IETF Internet-Draft on cryptographic MCP message security.

## Scope and sourcing (required)

- [x] This PR is focused: it modifies a single cheat sheet, or a small coordinated set, and the scope is described in the PR body.
[...]

PR #2209, opened 2026-05-29
New Cheat Sheet: AML and Sanctions Compliance for AI Agent Payments

arian-gogani:
## New Cheat Sheet: AML and Sanctions Compliance for AI Agent Payments

Invited by @mackowski in #2125.

AI agents increasingly initiate financial transactions autonomously. When an agent approves a loan, executes a trade, or processes a payment, the same AML and sanctions screening obligations apply as for human-initiated transactions. This cheat sheet covers the controls needed to make those decisions auditable, compliant, and independently verifiable.

### Sections covered:

- **Pre-transaction compliance checks** (screening before execution, signed receipts proving the check ran)
- **Tamper-evident audit trails** (Ed25519 signed, hash-chained receipts over RFC 8785 canonical JSON)
- **Cross-agent payment accountability** (receipt chains that travel with the transaction across multiple agents)
- **Sanctions list freshness** (staleness controls and version tracking)
- **Suspicious activity reporting** (agent-to-human escalation patterns)
- **Regulatory mapping table** (BSA/AML, EU AMLD6, EU AI Act Art. 12, FATF Rec. 16, OFAC)

This is a first draft. Happy to iterate on structure, depth, and coverage based on maintainer feedback.

PR #38, opened 2019-03-16
CSP CS Refactor

ThunderSon:
```
This PR covers issue #27.
```

PR #2217, opened 2026-06-06
docs(sec-8-11): add action_ref content address, receipt field spec, EU AI Act Article 12 mapping

arian-gogani:
## Summary

Refinements to Sections 8, 10 and 11 of the AML and Sanctions Compliance for AI Agent Payments cheat sheet, following up on PR #2209 and #2210.

Rewritten on 27 August after @mackowski's review, which asked for a disclaimer distinguishing proposed patterns from established standards. Writing that disclaimer turned up three errors in my own text, so this PR now does four things.

**Section 8, the `action_ref` content address.** A stable, recomputable content address for an agent action, so receipts can be correlated across systems without replaying the event log.

**Section 8, a standardised-versus-proposed disclaimer.** RFC 8785 (JCS) and RFC 8032 (Ed25519) are published IETF specifications and the canonicalization and signature guidance follows from them. The `action_ref` construction, the receipt field set, and the correlation pattern are proposed patterns that no RFC, NIST publication, or published audit framework specifies, and no conformance suite tests. The text now says so in the section itself.

**Section 10, a worked receipt** binding sanctions-list freshness to that content address.

**Section 11, the regulatory mapping row**, corrected. See below.

## Corrections to my own earlier text in this PR

- **AST09 does not specify the construction.** I had written that the formula "is specified as normative implementation guidance" in [OWASP Agentic Skills Top 10 AST09](https://github.com/OWASP/www-project-agentic-skills-top-10/blob/main/ast09.md). It is not. That file contains the string `action_ref` exactly once, in the outcome receipt field list, as "content-derived join key, independently recomputable". It names a property. The `HEX(SHA-256(JCS(...)))` construction is mine, and the text now attributes it to me.
- **Article 12 does not say tamper-evident.** The mapping row claimed "automatic tamper-evident logging". [Article 12](https://artificialintelligenceact.eu/article/12/) requires that high-risk systems "shall technically allow for the automatic recording of events (logs) over the lifetime of the system". It requires the record to exist. It does not require it to be signed, tamper-evident, or produced by a party other than the system being logged. The row now quotes it and states what it does not cover.
- **The date was wrong.** The diff said "enforcement December 2, 2027". There is no such date. Most of the Act applies from 2 August 2026 and Article 6(1) from 2 August 2027.
- **The ISO/IEC 42001 Annex A citation is removed.** It is a paid standard, I had not read the clause I was citing, and a clause number nobody in this thread has read is not a citation.

Scope: one file, `cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md`. markdownlint passes with 0 errors.

---

- [ ] In case of a new Cheat Sheet, you have used the Cheat Sheet template. *(not applicable, this edits an existing sheet)*
- [x] All the markdown files do not raise any validation policy violation.
- [x] All the markdown files follow these format rules.
- [ ] All your assets are stored in the **assets** folder. *(not applicable, no assets)*
- [ ] All the images used are in the **PNG** format. *(not applicable, no images)*
- [x] You verified/tested the effectiveness of your contribution.
- [x] The CI build of your PR pass.

- [x] This PR is focused: it modifies a single cheat sheet, and the scope is described in the PR body.
- [x] Every technical claim, recommendation, or threat assertion added in this PR is supported by a primary source linked inline, or is explicitly labelled in the text as a proposed pattern rather than a standard.

On that last box: it would not have been honest to tick it on the earlier version of this PR. The ISO/IEC 42001 Annex A citation was exactly the failure that box is there to prevent, and the AST09 attribution overstated what that document says. Both are fixed, and every remaining citation has been checked against the primary text.

- [x] I have used AI tools to generate the contents of this PR. I have verified the contents and I affirm the results. The LLM used is `Claude Opus (Anthropic)` and the prompt used was, in substance, *"review this open PR against the reviewer's request for a disclaimer, check every citation in it against the primary source, and correct anything that overstates what the source says."* I have independently verified every citation and technical claim against the cited source, including reading `ast09.md` and the text of EU AI Act Article 12 directly rather than relying on a summary. The three corrections listed above are the result of that check.

### DIFF
--- a/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
+++ b/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
@@ -173,6 +173,7 @@ Use the **JSON Canonicalization Scheme (JCS), [RFC 8785](https://www.rfc-editor.
 - Canonicalize every receipt with RFC 8785 (JCS) before signing, and again before verifying.
 - Sign over the hash of the canonical bytes, not raw or pretty-printed JSON.
 - Record the canonicalization, hash, and signature algorithm in the receipt (for example `canon: jcs`, `alg: ecdsa-p256-sha256`) so any verifier can reproduce it.
+- Derive a stable **content address** (`action_ref`) for each action: `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log.
 
 ### Don't
 
@@ -205,6 +206,22 @@ A receipt stating "screened, no match" is meaningless without which version of t
 - Include the sanctions-list source(s), version or publication date, and screening timestamp **inside the signed receipt**.
 - Define a maximum acceptable list age, record it in the receipt, and fail-closed if it is exceeded.
 - Make list freshness auditable after the fact from the receipt alone.
+- Bind the freshness record to the specific agent action using the `action_ref` content address (Section 8), so an auditor can link the screening result to the transaction without relying on log position:
+
+```json
+{
+  "screening_sources": [
+    {
+      "list": "OFAC-SDN",
+      "version": "2026-06-04",
+      "freshness_max_age_hours": 24,
+      "screening_timestamp_ms": 1749081600000
+    }
+  ],
+  "result": "no_match",
+  "action_ref": "a3f7c2..."
+}
+```
 
 ### Don't
 
@@ -223,6 +240,7 @@ The controls in this cheat sheet map to common AML and sanctions obligations. Th
 | Sanctions-list freshness in receipt (Section 10) | Obligation to screen against current lists; sanctions-evasion controls |
 | Fail-closed enforcement (Section 5) | Blocking obligations for sanctioned parties |
 | Trust-tiered limits (Section 6) | Risk-based approach (FATF Recommendation 1); monitoring thresholds |
+| Signed receipt with `action_ref` content address (Sections 4, 8-10) | EU AI Act Article 12 (automatic tamper-evident logging for high-risk AI systems, enforcement August 2, 2026); ISO/IEC 42001 Annex A (AI management system records) |
 
 ### Do
 
--- a/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
+++ b/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
@@ -173,7 +173,7 @@ Use the **JSON Canonicalization Scheme (JCS), [RFC 8785](https://www.rfc-editor.
 - Canonicalize every receipt with RFC 8785 (JCS) before signing, and again before verifying.
 - Sign over the hash of the canonical bytes, not raw or pretty-printed JSON.
 - Record the canonicalization, hash, and signature algorithm in the receipt (for example `canon: jcs`, `alg: ecdsa-p256-sha256`) so any verifier can reproduce it.
-- Derive a stable **content address** (`action_ref`) for each action: `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log.
+- Derive a stable **content address** (`action_ref`) for each action: `action_ref = HEX(SHA-256(JCS({agent_id, action_type, scope, timestamp_ms})))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log. This construct originates from [OWASP Agentic Skills Top 10 AST09](https://github.com/OWASP/www-project-agentic-skills-top-10/blob/main/ast09.md), where it is specified as normative implementation guidance for execution receipts.
 
 ### Don't
 
@@ -240,7 +240,7 @@ The controls in this cheat sheet map to common AML and sanctions obligations. Th
 | Sanctions-list freshness in receipt (Section 10) | Obligation to screen against current lists; sanctions-evasion controls |
 | Fail-closed enforcement (Section 5) | Blocking obligations for sanctioned parties |
 | Trust-tiered limits (Section 6) | Risk-based approach (FATF Recommendation 1); monitoring thresholds |
-| Signed receipt with `action_ref` content address (Sections 4, 8-10) | EU AI Act Article 12 (automatic tamper-evident logging for high-risk AI systems, enforcement August 2, 2026); ISO/IEC 42001 Annex A (AI management system records) |
+| Signed receipt with `action_ref` content address (Sections 4, 8-10) | EU AI Act Article 12 (automatic tamper-evident logging for high-risk AI systems, enforcement December 2, 2027); ISO/IEC 42001 Annex A (AI management system records) |
 
 ### Do
 
--- a/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
+++ b/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
@@ -173,7 +173,15 @@ Use the **JSON Canonicalization Scheme (JCS), [RFC 8785](https://www.rfc-editor.
 - Canonicalize every receipt with RFC 8785 (JCS) before signing, and again before verifying.
 - Sign over the hash of the canonical bytes, not raw or pretty-printed JSON.
 - Record the canonicalization, hash, and signature algorithm in the receipt (for example `canon: jcs`, `alg: ecdsa-p256-sha256`) so any verifier can reproduce it.
-- Derive a stable **content address** (`action_ref`) for each action: `action_ref = HEX(SHA-256(JCS({agent_id, action_type, scope, timestamp_ms})))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log. This construct originates from [OWASP Agentic Skills Top 10 AST09](https://github.com/OWASP/www-project-agentic-skills-top-10/blob/main/ast09.md), where it is specified as normative implementation guidance for execution receipts.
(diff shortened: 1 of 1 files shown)


razashariff:
Arian —Thanks

One point of framing: this is a contribution to the sheet rather than co-authorship — the document has a single author of record and I'd like to keep it that way. The git history will show your commit and that's the right attribution for what it is. OK for you to be credited though.

  @mackowski happy for you to merge if the content passes review.

mackowski (changes requested):
Thanks for the contribution! The `action_ref` pattern and EU AI Act mapping are interesting additions, but before we can merge this we'd need a disclaimer making clear these are proposed patterns rather than established standards — there's no RFC, NIST document, or published audit framework that specifies this construct yet.

razashariff:
Thanks @mackowski. To be clear from my side: the action_ref content-address, the receipt field spec, and the EU AI Act Article 12 mapping here aren't something I authored or verified, and they weren't part of the section I originally contributed. I can't vouch for them as a tested or standardised construct, and as you note, there's no RFC, NIST publication, or published audit framework behind them yet. So please don't read my earlier contribution as backing this one. Happy for you to drop it if it doesn't meet the bar.

cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
```diff
@@ -173,6 +173,7 @@ Use the **JSON Canonicalization Scheme (JCS), [RFC 8785](https://www.rfc-editor.
 - Canonicalize every receipt with RFC 8785 (JCS) before signing, and again before verifying.
 - Sign over the hash of the canonical bytes, not raw or pretty-printed JSON.
 - Record the canonicalization, hash, and signature algorithm in the receipt (for example `canon: jcs`, `alg: ecdsa-p256-sha256`) so any verifier can reproduce it.
+- Derive a stable **content address** (`action_ref`) for each action: `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log.
```
randomstuff: I don't really know what we are talking about here in this section (and more generally in the whole cheat sheet?).

Is the process described here an actual implementation requirements or is it more a suggested implementation for some existing functional requirements (from regulation texts)? In other words, could I instead of all that use a JWT and never have to care about JCS at all?

randomstuff: Is this actually based on https://datatracker.ietf.org/doc/draft-marques-asqav-compliance-receipts/05/ ? Is this explained somewhere?

arian-gogani:
@randomstuff good questions.

on whether this is implementation requirement vs suggested implementation: the cheat sheet is guidance, not a mandate. JCS + Ed25519 is one way to satisfy the underlying requirement (tamper-evident, independently verifiable receipts). you could use a JWT with a deterministic serialization and get the same property. the reason JCS is called out specifically is that JSON serialization is non-deterministic by default (key ordering, whitespace, number formatting), so two parties hashing the same logical object can get different digests. JCS (RFC 8785) closes that gap. if you use JWT with a fixed header and deterministic payload serialization, the verifiability property holds the same way.

on whether this is based on draft-marques-asqav-compliance-receipts: no, this was developed independently. the action_ref content-address pattern (SHA-256 over JCS-canonical fields) predates my awareness of that draft. the convergence is interesting though, it suggests the primitive shape is natural for this problem. the four-field preimage (agent_id, action_type, scope, timestamp_ms) is the minimum set that uniquely identifies an agent action without carrying the full event payload. happy to add a reference to the IETF draft if the reviewers think it adds context.

razashariff:
so the  functional requirement here is regulatory: agent payment and sanctions decisions have to produce tamper-evident, attributable, independently verifiable records (AML record-keeping, EU AI Act Art. 12). JCS is one suggested implementation of that, not a mandate.

  If a single issuer signs a receipt and one party verifies that exact token, a JWS/JWT is enough and you never touch JCS. Canonicalization only earns its place when different systems must independently re-serialize and re-verify the same record, or derive a stable cross-system content address. So unless cross-system correlation is in scope, the
  signed token is fine and you can skip it. Thx

cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
randomstuff: Nitpick, maybe `action_ref = HEX(SHA-256(JCS({agent_id, action_type, scope, timestamp_ms})))`

randomstuff: Would it make sense to say:

> Derive a stable **content address** (`action_ref`) for each action: eg. `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`?

randomstuff: (Also why not base64url?)

randomstuff:
(I don't know anything about the topic which makes reviewing quite difficult without quite a bit work.)

jmanico:
> (I don't know anything about the topic which makes reviewing quite difficult without quite a bit work.)

Just cc me or @mackowski into anything you want our attention on and move on to anything else that interests you.  :)

arian-gogani:
Thanks for the review, @mackowski. An update that changes the context:

Since this PR was opened, the `action_ref` construct was merged as normative implementation guidance in the OWASP Agentic Skills Top 10 (AST09, PRs #35 and #38, merged by project lead @kenhuangus on June 25). The formula is no longer a proposed pattern. It is referenced OWASP guidance with EU AI Act Article 12 relevance noted in the standard itself.

That said, I am happy to add a note clarifying that the construct originates from AST09 and linking to the canonical source, rather than presenting it as independently derived in this document. That would address the provenance concern without understating the status.

On the review comments from @randomstuff:
- HEX encoding: agreed, the formula should specify the output encoding. HEX is the convention in the AST09 text and the reference implementation.
- "content address" framing: yes, that is clearer. Happy to adopt that phrasing.
- base64url: HEX was chosen for human readability in logs and debuggability. No strong objection to base64url as an alternative, but the AST09 reference uses hex, so consistency argues for keeping it.

Shall I push the updates (disclaimer referencing AST09 as source, HEX explicit, "content address" phrasing)?

cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
arian-gogani: Good suggestion. 'Content address' is the more precise term and makes the property clearer: the identifier is derived from the content, not assigned. Happy to update.

arian-gogani: Hex is used because the OWASP AST09 spec and the x402 conformance vectors both use hex-encoded SHA-256. Staying consistent with the normative references avoids a format translation step during verification. Base64url would be equally valid technically, but switching would create a mismatch with the published conformance fixtures.

randomstuff:
(About the HEX thing and all, if you think it makes more sense in this context I'm not in position to argue :))

arian-gogani:
@mackowski I have pushed the disclaimer you asked for. Writing it turned up three things in my own text that were wrong, so those are corrected in the same commit, and I have rewritten the PR body to use the required template, which I had not done.

**The disclaimer**, at the end of Section 8:

> RFC 8785 (JCS) and RFC 8032 (Ed25519) are published IETF specifications, and the canonicalization and signature behaviours described above follow from them directly. The `action_ref` content address, the receipt field set, and the correlation pattern built on them are proposed patterns. No RFC, NIST publication, or published audit framework specifies them, and no conformance suite currently tests them. Treat them as one worked implementation of the underlying requirement rather than as a standard to conform to.

That is the line where you and @razashariff drew it, and I think it is the right place for it.

**Three corrections.**

*AST09 does not specify the construction.* I wrote that the formula "is specified as normative implementation guidance" there. It is not. `ast09.md` contains the string `action_ref` exactly once, in the outcome receipt field list: "content-derived join key, independently recomputable". It names a property and a field. The `HEX(SHA-256(JCS(...)))` construction is mine, and the text now says that.

*Article 12 does not say tamper-evident.* The row claimed "automatic tamper-evident logging". Article 12 says high-risk systems "shall technically allow for the automatic recording of events (logs) over the lifetime of the system". It requires the record to exist. It does not require it to be signed, to be tamper-evident, or to be produced by any party other than the system being logged. The row now quotes it and says what it does not cover.

*The date was wrong.* I had "enforcement December 2, 2027". There is no such date. Most of the Act applies from 2 August 2026 and Article 6(1) from 2 August 2027. My own PR body had said August 2026, so the body and the diff had been contradicting each other since June.

**One citation removed.** The row also cited "ISO/IEC 42001 Annex A (AI management system records)". ISO/IEC 42001 is a paid standard, I had not read the clause I was citing, and a clause number nobody in this thread has read is not a citation. It is gone.

@razashariff, this is the substance of your objection from 19 June, that you could not vouch for these as tested or standardised constructs. On three of them you were right in a way I had not checked, and the document now says which parts rest on published specifications and which do not. If the remaining `action_ref` paragraph still reads as more than a worked example, I am happy to cut it further, or drop it and leave only the JCS and signature guidance, which is the part that does follow from an RFC.

I have also filled in the PR template, including the AI disclosure, which the original body was missing entirely. That is on me and it should have been there in June.

One file, markdownlint clean.

cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md
```diff
@@ -173,6 +173,15 @@ Use the **JSON Canonicalization Scheme (JCS), [RFC 8785](https://www.rfc-editor.
 - Canonicalize every receipt with RFC 8785 (JCS) before signing, and again before verifying.
 - Sign over the hash of the canonical bytes, not raw or pretty-printed JSON.
 - Record the canonicalization, hash, and signature algorithm in the receipt (for example `canon: jcs`, `alg: ecdsa-p256-sha256`) so any verifier can reproduce it.
+- Derive a stable **content address** (`action_ref`) for each action: `action_ref = HEX(SHA-256(JCS({agent_id, action_type, scope, timestamp_ms})))`. This 64-character hex string uniquely and recomputably identifies the action across systems, enabling cross-system receipt correlation without replaying the full event log. [OWASP Agentic Skills Top 10 AST09](https://github.com/OWASP/www-project-agentic-skills-top-10/blob/main/ast09.md) names `action_ref` as a "content-derived join key, independently recomputable" in the outcome receipt. It does not specify a construction. The SHA-256-over-JCS formula above is one way to satisfy that property and is a proposed pattern, not a standardised one.
+
+> **What in this section is standardised, and what is not.** RFC 8785 (JCS) and
```
randomstuff: This is formatted as and citation but this is not a citation. I would maybe drop the question as well.

Also, I think, I would put this before the bullet points. Maybe an intro paragraph such as:

> This subsection proposes an implementation solution for the requirements. This does not describes an existing standard. [... extra info]

randomstuff (changes requested):
Some style/presentation points.

Issue #2235, opened 2026-06-17
Update: [JSON_Web_Token_for_Java_Cheat_Sheet] revocation method is vulnerable to JWT malleability


randomstuff:
## What is missing or needs to be updated?

The document suggests a revocation list / deny list for revoking JWTs. This boils down to:

1. hash the JWT;
2. store the hash in a revocation list.

~~~java
 public void revokeToken(String jwtInHex) throws Exception {
     if (jwtInHex != null && !jwtInHex.trim().isEmpty()) {
         //Decode the ciphered token
         byte[] cipheredToken = DatatypeConverter.parseHexBinary(jwtInHex);

         //Compute a SHA256 of the ciphered token
         MessageDigest digest = MessageDigest.getInstance("SHA-256");
         byte[] cipheredTokenDigest = digest.digest(cipheredToken);
         String jwtTokenDigestInHex = DatatypeConverter.printHexBinary(cipheredTokenDigest);

         //Check if the token digest in HEX is already in the DB and add it if it is absent
         if (!this.isTokenRevoked(jwtInHex)) {
             try (Connection con = this.storeDS.getConnection()) {
                 String query = "insert into revoked_token(jwt_token_digest) values(?)";
                 int insertedRecordCount;
                 try (PreparedStatement pStatement = con.prepareStatement(query)) {
                     pStatement.setString(1, jwtTokenDigestInHex);
                     insertedRecordCount = pStatement.executeUpdate();
                 }
                 if (insertedRecordCount != 1) {
                     throw new IllegalStateException("Number of inserted record is invalid," +
                     " 1 expected but is " + insertedRecordCount);
                 }
             }
         }

     }
 }
~~~

However this method is not correct because JWT are malleable (especially JWTs using ECDSA). An attacker in possession of the revoked JWT can forge another equivalent valid JWT as demonstrated by the script:

~~~python
from typing import Tuple
from joserfc import jwt, jwk
import base64
import string


def decode_b64u(value: str) -> bytes:
    missing_padding = len(value) % 4
    return base64.urlsafe_b64decode(value + missing_padding * "=")


def generate_jwt():
    key = jwk.JWKRegistry.generate_key("EC", "P-256")
    token = jwt.encode({"alg": "ES256"}, {"name": "John"}, key)
    public_key = jwk.JWKRegistry.import_key(key.as_dict(private=False))
    return (public_key, token)


def change_token_signature(token: str) -> str:
    header, body, signature = token.split(".")

[...]

PR #2242, opened 2026-06-21
Fix unsafe raw-token denylist; recommend jti+aud based revocation

shishir-cyber:
# You're A Rockstar

Fixes the JWT revocation guidance described in #2235. The cheat sheet currently recommends hashing the raw JWT (or its ciphered bytes) and storing that digest in a denylist for revocation. This is unsafe, because JWTs do not have a single canonical byte representation — a logically identical token can be transformed into a different byte sequence that still passes signature verification, producing a different hash and silently bypassing the denylist.

This affects revocation in two independent ways:

1. ECDSA signature malleability. For ECDSA-signed JWTs (e.g. ES256), a valid signature (r, s) has a second, equally valid form (r, (-s) mod n), where n is the public order of the curve's generator point. Both verify successfully against the same payload and public key but are different bytes, so they hash differently. This is a structural property of ECDSA, not an implementation bug — see Inherent Malleability of ECDSA Signatures and the formal treatment in True2F: Backdoor-resistant authentication tokens, Section 3. I independently reproduced this for a JWT/ES256 token using the proof-of-concept script from the linked issue: both the original and the derived signature verified successfully against the same public key, while differing in their signature bytes.
2. Non-strict JWT parsing. Many JWT libraries accept non-canonical encodings of the same logical token — extra base64 padding, alternate-but-decodable characters, or trailing bits that don't affect the decoded content. RFC 7515 Errata ID 8508 clarifies that trailing padding characters should be omitted and that decoders accepting padded input do so against the specification's recommendation — yet such leniency exists in real implementations, as demonstrated against cpp-jwt in this issue's discussion. These variants are byte-for-byte different from the original and bypass a hash-based denylist regardless of signing algorithm.

Changes

JSON_Web_Token_for_Java_Cheat_Sheet.md

- Added a note explaining both malleability causes above and why hashing the raw token is unsafe for revocation, with inline citations.
- Replaced the SHA-256 digest-based denylist with one keyed on the jti (JWT ID) claim, combined with aud (audience) to support scoping revocation per relying party. Both claims are embedded in the signed payload, so neither malleability issue can be used to bypass them — tampering with either claim invalidates the signature.
- Updated the SQL schema (composite primary key on jwt_id + aud) and the TokenRevoker Java class (isTokenRevoked / revokeToken) accordingly. Both methods now take an already-verified DecodedJWT rather than a raw token string, since jti/aud are only trustworthy to read after signature verification.
- Documented a deliberate simplification: per RFC 7519, Section 4.1.3, aud may be a single string or an array of strings; the example only checks the first value for readability, noted inline. Removed a leftover sentence describing the old SHA-256 approach that would otherwise have contradicted the new guidance.

REST_Security_Cheat_Sheet.md

- Updated the corresponding denylist reference, which described the same "digest or hash of the JWT" pattern, to match the jti + aud recommendation.
- Fixed a broken anchor link to the Java cheat sheet's revocation section (the anchor did not match the actual heading text).

Verification

I independently verified the technical claims in this PR rather than relying on the issue description alone:

Read RFC 7519, Sections 4.1.3 and 4.1.7, directly to confirm jti/aud claim semantics.
Read RFC 7515 Errata ID 8508 to confirm the base64url canonical encoding recommendation.
Reproduced the ECDSA signature malleability behavior by running a verification script against a freshly generated EC key pair and confirming both the original and derived signatures verify successfully while differing in bytes.

This PR fixes issue #`2235`.

- [x] I have used AI tools to generate the contents of this PR. I have verified the contents and I affirm the results. The LLM used is `Claude (Anthropic)` and the prompt used was an iterative conversation covering: drafting a warning on JWT denylist malleability, rewriting the SHA-256-based revocation example to use jti+aud claims, and fixing a related cross-reference in REST_Security_Cheat_Sheet.md. I have independently verified every citation and technical claim against the cited source (RFC 7519 sections 4.1.3 and 4.1.7 for jti/aud semantics; the ECDSA signature malleability math was verified against the proof-of-concept script in the original issue).

### DIFF
--- a/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
+++ b/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
@@ -207,7 +207,7 @@ Since JWTs are stateless, There is no session maintained on the server(s) servin
 
 Another way to protect against this is to implement a token denylist that will be used to mimic the "logout" feature that exists with traditional session management system.
 
-When the user wants to "logout" then it call a dedicated service that will add the token's identifying claims (`jti` and `aud`) to the denylist resulting in an immediate invalidation of the token for further usage in the application.
+When the user wants to "logout" then it call a dedicated service that will add the token's identifying claims (`jti` and `iss`) to the denylist resulting in an immediate invalidation of the token for further usage in the application.
 
 **Note:**
 
@@ -218,24 +218,23 @@ A denylist keyed on a digest of the raw token (e.g. `SHA-256(token)`) is unsafe,
 - **ECDSA signature malleability.** For JWTs signed with an ECDSA algorithm (e.g. `ES256`), a valid signature `(r, s)` has a second, equally valid form `(r, (-s) mod n)`, where `n` is the order of the curve's generator point. Both signatures verify successfully against the same public key for the same header and payload, but produce different token bytes — and therefore a different digest. An attacker in possession of a revoked token can compute this alternate signature and obtain a token that still authenticates.
 - **Non-strict JWT parsing.** Many JWT libraries tolerate multiple, non-canonical encodings of the same logical token — for example, base64url values with extraneous padding, alternate-but-decodable character substitutions, or trailing bytes with differing unused bits that decode to identical content. These variants are byte-for-byte different from the original token and therefore also bypass a hash-based denylist, regardless of signing algorithm (HMAC, RSA, or ECDSA).
 
-Because of this, the denylist must be keyed on a value that is **stable across these malleable encodings**, not on the token's raw bytes. The recommended approach is to use the `jti` (JWT ID) claim, which is a unique identifier assigned by the issuer at creation time and embedded inside the signed payload — making it immune to the malleability classes above, since any tampering with it invalidates the signature. Combining `jti` with the `aud` (audience) claim allows revocation to be scoped per relying party in multi-audience deployments.
-
-As a complementary hardening measure, JWT libraries should be configured (or chosen) to reject non-canonical encodings outright — strict base64url decoding, no padding tolerance — rather than relying on revocation logic to compensate for parser leniency.
+Because of this, the denylist must be keyed on a value that is **stable across these malleable encodings**, not on the token's raw bytes. The recommended approach is to use the `jti` (JWT ID) claim, which is a unique identifier assigned by the issuer at creation time and embedded inside the signed payload — making it immune to the malleability classes above, since any tampering with it invalidates the signature. Combining `jti` with the `iss` (issuer) claim ensures a globally unique denylist key — `jti` uniqueness is only guaranteed within a single issuer, so a (`jti`, `iss`) pair is required to prevent collisions between tokens from different issuers. Note that this denylist is audience-maintained (operated by the relying party), not by the issuer itself.
 
 #### Implementation Example
 
 ##### Block List Storage
 
-The following example demonstrates a denylist keyed on `jti` and `aud` rather than the raw token, per the recommendation above.
+The following example demonstrates a denylist keyed on `jti` and `iss` rather than the raw token, per the recommendation above.
 
 A database table with the following structure will be used as the central denylist storage.
 
 ``` sql
 create table if not exists revoked_token(
   jwt_id varchar(255) not null,
-  aud varchar(255) not null,
+  iss varchar(255) not null,
+  expires_at timestamp not null,
   revocation_date timestamp default now(),
-  primary key (jwt_id, aud)
+  primary key (jwt_id, iss)
 );
 ```
 
@@ -246,13 +245,17 @@ Code in charge of adding a token to the denylist and checking if a token is revo
 ``` java
 /**
  * Handle the revocation of the token (logout).
- * Revocation is keyed on the token's "jti" and "aud" claims rather than
+ * Revocation is keyed on the token's "jti" and "iss" claims rather than
  * a hash of the raw token, since JWTs do not have a single canonical
  * byte representation (see warning above) and a raw-token or
  * digest-based denylist can be bypassed via ECDSA signature
  * malleability or lenient JWT parsing. Both claims are embedded in the
  * signed payload, so neither can be altered without invalidating the
- * signature.
+ * signature. The (jti, iss) pair is used because jti uniqueness is
+ * only guaranteed per issuer — a malicious or rogue issuer could mint
+ * a JWT with the same jti as a legitimate one, causing a collision.
+ * Note: this denylist is audience-maintained (operated by the relying
+ * party), not by the issuer itself.
  * Use a DB in order to allow multiple instances to check for revoked
  * tokens and allow cleanup at centralized DB level.
  */
@@ -263,7 +266,7 @@ public class TokenRevoker {
     private DataSource storeDS;
 
     /**
-     * Verify if a given token (identified by its "jti" + "aud" claims)
+     * Verify if a given token (identified by its "jti" + "iss" claims)
      * is present in the revocation table.
      *
      * @param decodedToken Verified, decoded token (signature already validated)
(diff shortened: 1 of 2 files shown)


cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
```diff
@@ -207,19 +207,36 @@ Since JWTs are stateless, There is no session maintained on the server(s) servin
...
+
+Because of this, the denylist must be keyed on a value that is **stable across these malleable encodings**, not on the token's raw bytes. The recommended approach is to use the `jti` (JWT ID) claim, which is a unique identifier assigned by the issuer at creation time and embedded inside the signed payload — making it immune to the malleability classes above, since any tampering with it invalidates the signature. Combining `jti` with the `aud` (audience) claim allows revocation to be scoped per relying party in multi-audience deployments.
+
+As a complementary hardening measure, JWT libraries should be configured (or chosen) to reject non-canonical encodings outright — strict base64url decoding, no padding tolerance — rather than relying on revocation logic to compensate for parser leniency.
```
randomstuff: I am wondering if we should include this sentence. I would find more satisfying to use an implementation which is strict however:

* nearly all the implementations I have tried are currently not strict;
* and this does not directly constitutes a vulnerability as long as we don't use dangerous patterns.

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
```diff
@@ -207,19 +207,36 @@ Since JWTs are stateless, There is no session maintained on the server(s) servin
...
 ##### Block List Storage
 
+The following example demonstrates a denylist keyed on `jti` and `aud` rather than the raw token, per the recommendation above.
+
 A database table with the following structure will be used as the central denylist storage.
 
 ``` sql
-create table if not exists revoked_token(jwt_token_digest varchar(255) primary key,
-revocation_date timestamp default now());
+create table if not exists revoked_token(
```
randomstuff, line 232: We should probably include an expiration date (typically the expiration of the token + delay?) in order to be able to purge the table?

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
```diff
@@ -207,19 +207,36 @@ Since JWTs are stateless, There is no session maintained on the server(s) servin
...
+- **Non-strict JWT parsing.** Many JWT libraries tolerate multiple, non-canonical encodings of the same logical token — for example, base64url values with extraneous padding, alternate-but-decodable character substitutions, or trailing bytes with differing unused bits that decode to identical content. These variants are byte-for-byte different from the original token and therefore also bypass a hash-based denylist, regardless of signing algorithm (HMAC, RSA, or ECDSA).
+
+Because of this, the denylist must be keyed on a value that is **stable across these malleable encodings**, not on the token's raw bytes. The recommended approach is to use the `jti` (JWT ID) claim, which is a unique identifier assigned by the issuer at creation time and embedded inside the signed payload — making it immune to the malleability classes above, since any tampering with it invalidates the signature. Combining `jti` with the `aud` (audience) claim allows revocation to be scoped per relying party in multi-audience deployments.
```
randomstuff: Actually I think it should be scoped by the issuer, not by the audience in most cases?

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
```diff
@@ -228,85 +245,95 @@ Code in charge of adding a token to the denylist and checking if a token is revo
...
+        // only considers the first audience for simplicity; a production
+        // implementation handling multi-audience tokens should check
+        // revocation against every value in the list.
+        String audience = decodedToken.getAudience().isEmpty()
+            ? null : decodedToken.getAudience().get(0);
+
+        if (jwtId == null || jwtId.trim().isEmpty() || audience == null) {
+            // A token with no "jti" or "aud" cannot be safely revoked by
+            // this mechanism; treat it as untrusted in that case.
+            throw new IllegalArgumentException("Token has no \"jti\" or \"aud\" claim");
```
randomstuff: (((Note: maybe such a token should not be accepted at all in this case.)))

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
randomstuff: Sorry, I notice that the error was mine. In the original bug report, I mentionned "aud" but I was intended to say "iss" 😢

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
shishir-cyber: I'll drop the "libraries should reject non-canonical encodings" line and just frame jti+iss as the fix, since it 
does not depend on parser strictness to begin with.

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
shishir-cyber: I'll add an expiration timestamp column (token's exp + buffer)

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
shishir-cyber: You're right, iss makes more sense here, the denylist is something the issuer maintains, and scoping by issuer is the more natural boundary than audience. I'll update the warning text, schema, and the 
Java code to use iss instead of aud throughout.

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
shishir-cyber: Would you prefer I add a note in the cheat sheet recommending jti be enforced as a required claim at token issuance/validation, or move the check into a separate validation example?

randomstuff: I don't know. After some thinking, I guess we don't need to add anything about it in order to avoid too much verbosity.

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
randomstuff: >the denylist is something the issuer maintains

Actually, I would say the deny list we are talking about is something the audience is maintaining. For issuer-maintained revocation list, the blessed solution is too use [Token Status List](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-status-list) (TSL) (draft).

That is maybe something we should talk about :

* mention that the goal here is audience-maintained revocation list;
* add a ref to [TSL](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-status-list) as a blessed and scalable issuer-side revocation list.

This can be addressed in a follow-up PR if you wish.

randomstuff: > scoping by issuer is the more natural boundary than audience

The reason for using (jti, iss) is that `jti` is issued by issuer and therefore we can only guarantee that it is unique for a given issuer. A malicious/rogue issuer could issue JWT with jti of another issuer in order to trigger jti "collisions".

jmanico:
Hey @randomstuff you are now one of the three leads of this series and have admin access to this repo. We cannot keep up with AI submissions so you are authorized to "squash and merge" anything as you see fit to help us maintain this project.

Welcome. I'll add you to the various lead pages soon as well. cc @mackowski

cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.md
shishir-cyber: I will add a note clarifying that this denylist is audience-maintained, not issuer-maintained. I'll leave the TSL
reference for a follow-up PR as you suggested.

shishir-cyber: > > scoping by issuer is the more natural boundary than audience
> 
> The reason for using (jti, iss) is that `jti` is issued by issuer and therefore we can only guarantee that it is unique for a given issuer. A malicious/rogue issuer could issue JWT with jti of another issuer in order to trigger jti "collisions".

I will update the doc comment and warning prose to reflect this reasoning instead of the previous explanation.

jmanico (approved):

randomstuff:
@shishir-cyber, thanks! :confetti_ball:

PR #2255, opened 2026-06-24
Fix: add missing imports and undefined references in AI Agent Security Cheat Sheet

Inamajet:
While reading the AI Agent Security Cheat Sheet, I noticed several code examples reference modules and functions that are never imported or defined.

Section 5 (Output Validation) uses json.dumps() and re.search() without importing json or re.

Section 7 (Multi-Agent Security) calls generate_uuid() and CircuitBreaker() without any definition or import.

A developer copying these examples would get NameError exceptions immediately.

This PR adds the missing imports to both sections and defines generate_uuid() using Python's built-in uuid module, with a note that CircuitBreaker requires the pybreaker library.

### DIFF
--- a/cheatsheets/AI_Agent_Security_Cheat_Sheet.md
+++ b/cheatsheets/AI_Agent_Security_Cheat_Sheet.md
@@ -186,6 +186,7 @@ class SecureAgentMemory:
 #### Action Classification and Approval Flow
 
 ```python
+import uuid
 from enum import Enum
 from dataclasses import dataclass
 
@@ -228,7 +229,7 @@ class HumanInTheLoopController:
         
         # Queue for human review
         action = PendingAction(
-            action_id=generate_uuid(),
+            action_id=str(uuid.uuid4()),
             tool_name=tool_name,
             parameters=self._sanitize_params_for_display(params),
             risk_level=risk_level,
@@ -524,7 +525,7 @@ class SecureAgentBus:
         
         # Check circuit breaker
         if self.circuit_breakers[sender_id].current_state == "open":
-            raise CircuitBreakerOpen(f"Agent {sender_id} is temporarily blocked")
+            raise RuntimeError(f"Agent {sender_id} is temporarily blocked")
         
         # Validate recipient authorization
         if recipient_id not in sender["allowed_recipients"]:
--- a/cheatsheets/AI_Agent_Security_Cheat_Sheet.md
+++ b/cheatsheets/AI_Agent_Security_Cheat_Sheet.md
@@ -277,6 +277,8 @@ For destructive, financial, administrative, or externally visible actions, add c
 #### Output Validation Pipeline
 
 ```python
+import json
+import re
 from pydantic import BaseModel, validator
 from typing import Optional, List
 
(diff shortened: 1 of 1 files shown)


cheatsheets/AI_Agent_Security_Cheat_Sheet.md
```diff
@@ -481,6 +483,12 @@ from typing import Optional
 import jwt
 from datetime import datetime, timedelta
 
+import uuid
+from pybreaker import CircuitBreaker  # pip install pybreaker
+
+def generate_uuid() -> str:
```
randomstuff: I would just use:

~~~python
action_id=str(uuid.uuid4())
~~~

and drop the extra function.

cheatsheets/AI_Agent_Security_Cheat_Sheet.md
```diff
@@ -481,6 +483,12 @@ from typing import Optional
 import jwt
 from datetime import datetime, timedelta
 
+import uuid
+from pybreaker import CircuitBreaker  # pip install pybreaker
```
randomstuff, line 488: Actually, I don't believe this class comes from this lib because none of these parameters appers to be supported:

~~~python
        self.circuit_breakers[agent_id] = CircuitBreaker(
            failure_threshold=5,
            recovery_timeout=60
        )
~~~

cf.

~~~python
class CircuitBreaker:
    """
    [...]
    """

    def __init__(
        self,
        fail_max: int = 5,
        reset_timeout: float = 60,
        success_threshold: int = 1,
        exclude: Iterable[type[ExceptionType] | Callable[[Any], bool]] | None = None,
        listeners: Sequence[CBListenerType] | None = None,
        state_storage: CircuitBreakerStorage | None = None,
        name: str | None = None,
        throw_new_error_on_trip: bool = True,
    ) -> None:
        ...
~~~

and `is_open()` does not exist either. I'm not sure this is referencing an actual existing API.

Inamajet (commented):
corrected CircuitBreaker to use fail_max and reset_timeout per pybreaker's actual API, and simplified uuid generation to inline str(uuid.uuid4()) usage. Thank you for catching that.

cheatsheets/AI_Agent_Security_Cheat_Sheet.md
Inamajet: updated CircuitBreaker to use fail_max and reset_timeout per pybreaker's actual API. Also removed the is_open() reference. Thank you.

Inamajet:
Confirmed i fixed CircuitBreaker instantiation to use fail_max and reset_timeout in the register_agent method, and replaced is_open with current_state == "open" in send_message. All instances now match pybreaker's actual API.

Inamajet:
@randomstuff what you think are any other changes necessary?

randomstuff:
Oh sorry. Yes, I think `CircuitBreakerOpen` does not exist either (which make we wonder where it comes from :thinking: ). Maybe this should be replaced this with some stdlib exception, then?

Inamajet:
@randomstuff  i replaced CircuitBreakerOpen with RuntimeError from Python's standard library, and updated generate_uuid() in HumanInTheLoopController to use str(uuid.uuid4()) inline. All API issues should now be resolved.

Issue #2264, opened 2026-06-30
Update: File Upload Sheet < extension bypass examples and config file abuse


Vishal-HaCkEr1910:
## What is missing or needs to be updated?
1. This cheat sheet already has a couple of extension bypass examples (double extensions, null bytes) ,but a few common ones aren't mentioned.

- case manipulation in ext.
- alternative extensions tied to the same interpreter.

2. The sheet doesn't currently mention that an attacker can upload a web server config file (like .htaccess or web.config) into the upload folder . The existing advice to store files outside the webroot already covers this, so it would just be a short note tying the two together near "File Storage Location".

## How should this be resolved?
1. extension bullet points : 
- Case manipulation, e.g. .pHp , .PHP5 , to bypass case-sensitive blocklist filters
- Alternative extensions tied to the same interpreter eg. .phtml , .phar , .pht for PHP , or .jsp/.jspx , .asp/.aspx for Java/.NET

2. And add one short sentence under "File Storage Location" noting that
server config files (.htaccess, web.config) can be uploaded to change
how a directory is handled , with the existing "store outside the
webroot" point .

Happy to send a pr if this is not already covered yet.

mackowski:
Thanks @Vishal-HaCkEr1910 — this is a good, in-scope addition. Both items are well-known, documented bypasses that fit the cheat sheet's threat model, and the scope is small. ACK — a PR is welcome.

janderik:
File upload extension bypass is a pervasive issue. Here's a comprehensive defense strategy:

**PHP code for robust validation using `finfo_file()`:**

```php
function validateUploadedFile(UploadedFile $file): array
{
    $errors = [];

    // 1. MIME type validation via fileinfo (not client-provided)
    $finfo = new \finfo(FILEINFO_MIME_TYPE);
    $realMime = $finfo->file($file->getPathname());

    $allowedMimes = [
        'image/jpeg',
        'image/png',
        'image/gif',
        'image/webp',
        'application/pdf',
    ];

    if (!in_array($realMime, $allowedMimes, true)) {
        $errors[] = "MIME type '$realMime' is not allowed.";
    }

    // 2. Extension validation against a strict allowlist
    $extension = strtolower($file->getClientOriginalExtension());

    $allowedExtensions = ['jpg', 'jpeg', 'png', 'gif', 'webp', 'pdf'];

    if (!in_array($extension, $allowedExtensions, true)) {
[...]

PR #2267, opened 2026-06-30
Add extension bypass examples and config file note to File Upload Cheat Sheet 

Vishal-HaCkEr1910:

- [] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](../templates/New_CheatSheet.md).
- [x] Any references to websites have been formatted as `http://portswigger.net/web-security/file-upload`

Two small additions to File_Upload_Cheat_Sheet.md :
1. Added case manipulation and alternative-extension examples.
2. Added a small para under "File Storage Location" section about web server config files.

- [x] Every technical claim, recommendation, or threat assertion added in this PR is supported by a primary source (RFC, NIST, OWASP standard, vendor documentation, peer-reviewed research) linked inline as `http://portswigger.net/web-security/file-upload`.

- [x] I have NOT used any AI tool to generate the contents of this PR.

This PR fixes issue #2264 .


jmanico (dismissed):

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -52,6 +52,8 @@ Ensure that the validation occurs after decoding the filename, and that a proper
 
 - Double extensions, _e.g._ `.jpg.php`, where it circumvents easily the regex `\.jpg`
 - Null bytes, _e.g._ `.php%00.jpg`, where `.jpg` gets truncated and `.php` becomes the new extension
+- Case manipulation, _e.g._ `.pHp`, `.PHP5`, to bypass case-sensitive blocklist filters
+- Alternative extensions tied to the same interpreter, _e.g._ `.phtml`, `.phar`, `.pht` for PHP, or `.jsp`/`.jspx`, `.asp`/`.aspx` for Java/.NET
```
randomstuff: Also on Windows (:fearful:)², alternate data streams (`foo.php::$DATA`)? (and maybe 8.3 file names?)

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -126,6 +128,8 @@ The location where the files should be stored must be chosen based on security a
 
 Storing files in a studied manner in databases is one additional technique. This is sometimes used for automatic backup processes, non file-system attacks, and permissions issues. In return, this opens up the door to performance issues (in some cases), storage considerations for the database and its backups, and this opens up the door to SQLi attack. This is advised only when a DBA is on the team and that this process shows to be an improvement on storing them on the file-system.
 
+Be aware that an attacker may attempt to upload a web server configuration file (e.g. `.htaccess`, `web.config`) into the upload directory. If the web server allows per-directory configuration overrides, such a file could be used to change how that directory handles file types, for example mapping an otherwise harmless extension to be executed by the server. Storing files outside the webroot, as recommended above, removes this risk entirely, since the uploaded directory is never directly served by the web server in the first place.
```
randomstuff: Maybe worth recommending disabling .htaccess and such in upload directories?

cheatsheets/File_Upload_Cheat_Sheet.md
Vishal-HaCkEr1910: Yeah , I updated the sheet with two bullets since the fix is different for both of these
NTFS-ADS : reject filenames with a colon .
8.3 FILE NAMING : disable short name generation , updation plus WAF filtering for the tilde enumeration side.

cheatsheets/File_Upload_Cheat_Sheet.md
Vishal-HaCkEr1910: Agreed, added it in the previously made para as a second layer of defence .

jmanico (dismissed):

cheatsheets/File_Upload_Cheat_Sheet.md
```diff
@@ -52,6 +52,10 @@ Ensure that the validation occurs after decoding the filename, and that a proper
...
+- Alternative extensions that may be mapped to the same server-side interpreter depending on configuration, _e.g._ `.phtml`, `.php5`, `.pht` for PHP, or `.jsp`/`.jspx`, `.asp`/`.aspx` for Java/.NET ([PortSwigger](https://portswigger.net/web-security/file-upload))
+- Windows NTFS Alternate Data Streams, _e.g._ `shell.asp:.jpg` or `shell.php::$DATA`, where the colon is interpreted as a stream separator and the actual file created on disk carries the extension before the colon - reject any filename containing a colon (`:`) ([OWASP - Unrestricted File Upload](https://owasp.org/www-community/vulnerabilities/Unrestricted_File_Upload))
+- Windows 8.3 short filenames, _e.g._ `.htaccess` resolving to `HTACCE~1`, which can still be functional even if a filter only matches the long filename — disable 8.3 short filename generation on the server (`fsutil behavior set disable8dot3 1` for new files; existing short names require separate remediation), or use latest IIS version along with WAF to filter requests with `~`, `*` ,`?` etc .([OWASP - Unrestricted File Upload](https://owasp.org/www-community/vulnerabilities/Unrestricted_File_Upload))
```
randomstuff: I think there is a small typographic error here:

> `?` etc .(

I think, it should be:

> `?`, etc. (

randomstuff: Otherwise LGTM  :smile: :+1:

Vishal-HaCkEr1910: I have fixed the typo.

jmanico (approved):

Vishal-HaCkEr1910:
@randomstuff, I have fixed the typo you flagged in the latest commit. Since you mentioned it was otherwise LGTM, could you take another look and approve if it looks good? Thanks!

mackowski (changes requested):
Thanks for this, and for iterating on the review feedback — the case-manipulation and alternative-extension bullets plus the config-file note feel like useful additions to this sheet, and I think they're in the right place.

I do want to raise one thing on the **Windows 8.3 short filename bullet** before this goes in, because from what I can tell the premise may be inverted. My understanding is that an 8.3 name is a generated *alias for a file that already exists* — so uploading a file named `HTACCE~1` would just create a file called `HTACCE~1`, which Apache wouldn't read as `.htaccess` (`AccessFileName` defaults to `.htaccess`). Reading the OWASP page you cite, it seems to describe something narrower: *replacing an existing file* via its short name. If that's right, it's more of an overwrite primitive than an extension-validation bypass, and might sit better next to the existing "Overwrite an existing file on the system" threat. My own leaning is that it could be dropped entirely, since the sheet already leads with generating filenames server-side — but I may be missing a case you had in mind, so keen to hear your take.

Relatedly, I *think* `use latest IIS version along with WAF to filter requests with ~, *, ?` is aimed at IIS tilde short-name enumeration, which as I understand it is a request-side information-disclosure issue rather than an upload bypass — filtering `~` in request paths wouldn't close an upload filter hole. If that's a fair read, it might be worth dropping that clause. The `fsutil` half looks correct to me and seems worth keeping, though I couldn't find `fsutil`, `disable8dot3`, or the WAF/IIS guidance anywhere on the OWASP page it's cited to. If it helps, [fsutil behavior](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/fsutil-behavior) and [fsutil 8dot3name](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/fsutil-8dot3name) both look like they'd cover those points directly.

Happy to be corrected on any of the above if I've misread the mechanics — and glad to re-review once we've settled on a direction.

jmanico (dismissed):

Vishal-HaCkEr1910:
You're right . The 8.3 bullet had the mechanism backwards — the short name is an alias generated for an existing long-named file, not something you get by uploading a file under the short name directly, so HTACCE~1 was never going to resolve as .htaccess. Every long filename gets an auto-generated 8.3 alias by NTFS itself, so the alias can only ever point back at a file that already exists — it's not something an attacker can manufacture by choosing an upload filename.
The real case is the overwrite one you described: if a collision/overwrite check only validates the long filename, an attacker who can predict an existing sensitive file's short alias can reference it directly and overwrite it, bypassing an exact-name or extension check. I have dropped it from the extension-bypass list and moved a corrected version under Filename Safety, tied to the random-filename recommendation already there — since generating the stored filename server-side removes the attacker's ability to target a specific existing file in the first place.

Agreed on the IIS/WAF ~ * ? line too — that's a request-parsing information-disclosure bug used to discover short names, not to bypass an upload filter. Filtering ~ stops that discovery but does nothing for the overwrite issue itself, so I have dropped it rather than force a reframe.

Kept the fsutil remediation , just re-cited it to Microsoft's own fsutil 8dot3name / fsutil behavior docs instead of the OWASP page, since you're right that it wasn't actually there. Worth noting for scope: 8.3 generation is off by default on non-system volumes since Windows Server 2012 , so this mainly bites deployments still on the system volume or older defaults.
Pushed the update, please let me know if it's fine.

jmanico (approved):

randomstuff (approved):

mackowski (approved):

Issue #2275, opened 2026-07-04
Update: [JSON_Web_Token_Cheat_Sheet] Token Infirmation DisclosureUpdate: [JSON_Web_Token_Cheat_Sheet]


randomstuff:
## What is missing or needs to be updated?

The Token Disclosure section of the JWT cheat sheet mentions encryption of the JWT with AEAD.

The example encrypts the raw token using AES-GCM.

Several things are somewhat weird / not clear:

* it is not clear is we are talking about encryption during transmission on the token or at rest encryption;
* it is not talking about using en encrypted JWT (JWE).

~~~java
Aead aead = AeadFactory.getPrimitive(keysetHandle);
byte[] cipheredToken = aead.encrypt(jwt.getBytes(), null);
return DatatypeConverter.printHexBinary(cipheredToken);
~~~

The implementation might be OK for at-rest encryption (but is not specific to JWTs).

The implementation is weird if we are talking about transferring the JWT. In this case, we expect to use a JWE.

## How should this be resolved?

I would probably remove this section nearly entirely and mention using encrypted JWE (JWE) in this case.

shishir-cyber:
Hi @randomstuff , I would like to work on this if it gets acknowledged. Given my prior work on the same cheat sheet (#2242, #2258), I'm familiar with the file and the conventions. My plan would be to remove the current Token Ciphering implementation section and replace it with a concise note explaining the at-rest vs in-transit distinction and pointing to JWE (RFC 7516) as the standard approach for JWT confidentiality during transmission.

randomstuff:
I @shishir-cyber, actually I have started working on a PR which deals with that and other things. I'll let you know if that does not land 😄

randomstuff:
@shishir-cyber This topic is discussed in #2283. The goal might be to:

1) add a small notes on the JWT cheat sheet on the topic (see discussion in #2283);
2) in the future add a dedicated cheat sheet about JWE.

If you are still interested, you can work on the PR to do (1) in parallel to #2283 (or you can do it afterward #2283 is merged if you want to avoid fixing conflicts).

shishir-cyber:
@randomstuff I will wait for #2283 to get merged and then add the JWE note as a follow up.

janderik:
Great catch! The current Token Disclosure section is indeed confusing. I agree that it needs clarification.

**Issues with the current implementation:**

1. **Ambiguous context**: It's unclear whether this is for:
[...]

PR #2283, opened 2026-07-09
Cleanup/streamline JWT cheat sheet

randomstuff:

- [X] This PR is focused: it modifies a single cheat sheet, or a small coordinated set, and the scope is described in the PR body.
- [X] Every technical claim, recommendation, or threat assertion added in this PR is supported by a primary source (RFC, NIST, OWASP standard, vendor documentation, peer-reviewed research) linked inline as `[text](URL)`.
- [X] I have read each source I cite and confirm it actually supports the claim. I have not relied on summaries, hearsay, or model-generated citations.

This PR fixes issue #2275 (and goes beyond).


cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -73,767 +63,100 @@ base64url(
...
-**Note:**
+If revocation of the JWTs by the issuer is needed, the [Token Status Lists](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-status-list) (TSL) can be used:
 
-Here ciphering is added mainly to hide internal information but it's very important to remember that the first protection against tampering of the JWT is the signature. So, the token signature and its verification must be always in place.
+- the JWT contains the URI of a TSL;
+- the TSL aggregates the revocation status of several tokens in compressed form;
+- the audience can fetch the TSL to obtain the revocation status of the JWT.
 
-#### Implementation Example
+This works similar to [Certificate Revocation Lists](https://datatracker.ietf.org/doc/html/rfc5280#section-5) (CRLs) for certificates.
```
kwwall: I'm not sure this a good way to phrase this, in part because it is misleading in the sense of the the amount of work that may be implied. In principle, the 2 are similar, because both are built on deny-lists, but I think that's where the similarity ends. The infrastructure that CRLs most commonly rely upon (that is, web browsers and web PKI, with CAs maintaining the CRL) is provided for the most part, out-of-the-box. (The biggest exception is when you are handling TLS connections directly in code and therefore have to code something on the client-end to see if the server-side certificate has been revoked.)  But you usually pretty much have to build this infrastructure from scratch here, or at least a significant part of it. 

I think maybe if may just be better to leave something like line 220 in place, or perhaps change this sentence to just note something like:
```text
This works on a deny-list of JWT hashes that your application or infrastructure must maintain.
```

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -73,767 +63,100 @@ base64url(
...
-1. Store the token using the browser *sessionStorage* container, or use JavaScript *closures* with *private* variables
-1. Add it as a *Bearer* HTTP `Authentication` header with JavaScript when calling services.
-1. Add [fingerprint](JSON_Web_Token_Cheat_Sheet.md#token-sidejacking) information to the token.
+## Replay protection
 
-By storing the token in browser *sessionStorage* container it exposes the token to being stolen through an XSS attack. However, fingerprints added to the token prevent reuse of the stolen token by the attacker on their machine. To close a maximum of exploitation surfaces for an attacker, add a browser [Content Security Policy](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html) to harden the execution context.
+### JWT denylist
 
-But, we know that *sessionStorage* is not always practical due to its per-tab scope, and the storage method for tokens should balance *security* and *usability*.
+In some cases, the audience of the token might want to maintain a local denylist. This might be for example used a simple form of JWT replay protection.
```
kwwall: Given the folks reading this for the first time may no absolutely zero about JWTs, I think it's potentially unclear of what is meant by the "audience of the token" here to newbies. To them, it could mean:

1. the "aud" claim of the JWT. (This seems most likely.)
2. or, it could mean all the people who are using the JWT. (This seems doubtful.)

Presuming your meaning is indeed, number 1, why not just refer to it as the

```text
"aud" claim of the JWT
```
Or, if by some oddball chance, that's _not_ what you meant, it definitely needs clarified.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
kwwall: I'm not sure this a good way to phrase this. In principle, they are similar, because both are built on deny-lists, but I think that's where the similarity ends. The infrastructure that CRLs most commonly rely upon (that is, web browsers and web PKI, with CAs maintaining the CRL) is provided for the most part, out-of-the-box. (The biggest exception is when you are handling TLS connections directly in code and therefore have to code something on the client-end to see if the server-side certificate has been revoked.)  But you usually pretty much have to build this infrastructure from scratch here, or at least a significant part of it, so at a bare minimum it undervalues the work effort that is needed to implement this. (Yeah, some JWT auth services handle this; e.g., I think Ping Identity's PingFederate is one of them, but I'm not sure how common this is.)

kwwall (changes requested):
In general, this looks okay, in part because I too was bothered by:

- The inconsistency of all the examples. (The seemingly arbitrary use of implementation programming languages in particular I thought was strange.)
- The length of some of the examples was disconcerting. We generally have decided to place such things into collapsable sections, but this was not.
- Obscure terminology such as the use of "_ciphered_ data" rather than "_encrypted_ data".
- Using a Java example of AEAD with Google Tink and AES/GSM/NoPadding, rather than JWE, which seems more in keeping with JWT.

However, while issue #2275 made some of these points (particularly the last one), all it did it remove the whole Google Tink example and ignored the reason for that being there in the first place, rather than replacing it with JWE, which is what I was expecting.

Therefore, I would like to see that changed as well as the other 2 comments that I made.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
randomstuff: The intent was to anchor the readers on something they probably already know. But maybe this is more misleading than actually useful.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
randomstuff: Yeah, maybe "consumer" of the token would be clearer?

randomstuff:
> However, while issue https://github.com/OWASP/CheatSheetSeries/issues/2275 made some of these points (particularly the last one), all it did it remove the whole Google Tink example and ignored the reason for that being there in the first place, rather than replacing it with JWE, which is what I was expecting.

Yes, maybe adding some short discussion about using JWE for claims encryption if that is required. I actually wanted to avoid going too deep in JWE for now (keep them for a later revision) in order to initially focus on JWS.

The reasons being that encryption adds a lot of complexities (and is easier to screw up badly) which might need time and care to discuss properly:

* key exhaustion
* harvest now decrypt later
* key rotation won't "save" your old tokens
* should probably switch to hybrid PQ/T encryption right now (whereas this is less critical for digital signatures)
* explaining what exactly does the AEAD of JWE bring in practice in term of security
* nested JWT

That being said, we can add a minimal section for encryption:

* warning against trying to rely on signed JWT for claim confidentiality (and PII)
* saying that JWE should be used in this case
* saying that usually we want to use nested JWT
* but that JWE is not covered yet

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
randomstuff: That is somewhat covered by the introduction:

> An application (**presenter**) presents the token to a consuming application (**audience**) which can verify the token's authenticity and validity and take decisions or actions based on these claims.

but I'm not sure if this is enough and if we should expect the user to actually remember the concepts tht beginning of the cheatsheet.

I wonder if a more explicit "definitions/concepts" parts (audience, issuer, subject, etc.) would help but that might appear too academic for a cheat sheet (?).

kwwall: I'm personally not fond of referring to this as "audience" because I confused it with the "aud" claim itself.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
kwwall: I figure that was intent, but I think most devs who aren't writing their own code for TLS sockets (which probably includes most of them) probably only know that CRL checking is handled between the web browser and CA and then don't need to do anything for that case. So I don't want them to think they won't need to do anything here because a lot of the time, that just isn't true. Hence my suggestion. It's not a showstopper, but many of those who are seeing our OWASP Cheat Sheets are completely new to the topic, so it's probably not a good thing to assume prior knowledge, especially to an unrelated topic like web PKI. Just my $.02. Not a showstopper though.

kwwall:
@randomstuff wrote:

> > However, while issue #2275 made some of these points (particularly the last one), all it did it remove the whole Google Tink example and ignored the reason for that being there in the first place, rather than replacing it with JWE, which is what I was expecting.
> 
> Yes, maybe adding some short discussion about using JWE for claims encryption if that is required. I actually wanted to avoid going too deep in JWE for now (keep them for a later revision) in order to initially focus on JWS.
> 
> The reasons being that encryption adds a lot of complexities (and is easier to screw up badly) which might need time and care to discuss properly:
> 
> * key exhaustion
> * harvest now decrypt later
> * key rotation won't "save" your old tokens
> * should probably switch to hybrid PQ/T encryption right now (whereas this is less critical for digital signatures)
> * explaining what exactly does the AEAD of JWE bring in practice in term of security
> * nested JWT

I think JWE is probably complex enough to have its own Cheat Sheet page for the complexity reasons you mention.

> That being said, we can add a minimal section for encryption:
> 
> * warning against trying to rely on signed JWT for claim confidentiality (and PII)
> * saying that JWE should be used in this case
> * saying that usually we want to use nested JWT
> * but that JWE is not covered yet

I'd prefer having a new JWE cheat sheet made and mention the minimal stuff there, and just reference it and we can flesh it out as we have time or volunteers step up. Unfortunately, I am not able to assist very much at the present time.

jmanico:
I agree that JWE deserves it's own cheatsheet. They are typically not even needed except in regulated environments. Also, one should not put PII or other sensitive data in a JWT. Most times I see folks using JWE it because they designed their claims incorrectly.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
randomstuff: > I'm personally not fond of referring to this as "audience" because I confused it with the "aud" claim itself.

Yeah, well actually the "aud" claim refers to the actual audience so … but understand your concern.

randomstuff: Replaced "audience"  → "consumer"

randomstuff:
The plan is to address confidentiality considerations (en mention JWE) in an upcoming PR (as disussed in #2275).

And later include a separate Cheat sheet about more in-depth JWE.

randomstuff:
I have slightly reorganized the "JWT denylist" section and expanded the "alternative solutions" discussion.

mackowski:
Maintainer nudge to move this forward 🙂

The substance here is a clear improvement and it's well-sourced, so I don't want it stalling. Reading the thread, there seems to be consensus already: defer JWE / claims-confidentiality to a follow-up PR, and give JWE its own cheat sheet later (per @randomstuff, @kwwall, @jmanico above).

@kwwall, if you're good with that, could you dismiss/convert your `CHANGES_REQUESTED` so this isn't blocked on the JWE replacement? And @randomstuff, could you open a tracking issue for the JWE mention + dedicated sheet so it's not lost?

One small thing worth folding into this PR before merge (quick, non-blocking): the "Threats" section now only covers `alg:none`, so it's worth one bullet on pinning the expected algorithm / algorithm-confusion (RS256 to HS256), since that's a primary threat in RFC 8725. Happy to leave that as a fast follow if you'd rather merge first.

Thanks all, let's get this one merged.

randomstuff:
@mackowski, OK, I'll finish that in the near future.

kevinwwall-owasp (approved):
@randomstuff & @mackowski  - This looks 👍 to me. I don't see my original 'Request changes' review to undo that part, so I hope this subsequent approval will unblock that.

randomstuff:
@mackowski, I've added some key type confusion material (either to be merged in this PR or in a follow-up PR).

Additional threats would have to be added in follow PRs.

randomstuff:
For example, one very easy error to make is to blindly trust/follow" the `jwk`, `jku`, `x5u`, `x5t`, `x5t#S256` header parameters.

Some other topics:

* issuer confusion
* audience confusion
* token type confusion

kwwall (dismissed):
All the changes that I requested originally have been addressed to my satisfaction.  Sorry for the mix-up approval yesterday. I didn't realize that I was still logged in as as my new alternate 'kevinwwall-esapi' associated with my OWASP email address, and since that's not in the CODEOWNERS file (and please don't add it there), it probably didn't do what was intended. I still couldn't find the original review, but I did mark some of the comments as Resolved.

randomstuff:
I have merge the base PR and added the key type confusion threat as a follow up PR #2320

Issue #2302, opened 2026-07-23
Update: [JWT] Additional threats


randomstuff:
## What is missing or needs to be updated?

After cleanup PR #2283, the content would have to be fleshed out.

## How should this be resolved?

Add something about the threats:

* key type confusion (→ added in #2283);
* issuer confusion;
* audience confusion;
* token type (cross-JWT) confusion;
* blindly trusting/following the jwk, jku, x5u, x5t, x5t#S256 header parameters.

Check:

* [RFC88725](https://www.rfc-editor.org/info/rfc8725/)
* [RFC88725 bis](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rfc8725bis-07) (draft)

ArockiaRajamanickam:
Hi @randomstuff, I'd like to work on this one. My plan is four new subsections under the threats section, following the structure you set up in #2283: issuer confusion, audience confusion, token type confusion (cross-JWT), and trusting the key selection header parameters (jwk, jku, x5u and friends). I'll base the content on RFC 8725 and the 8725bis draft, and I'll wait for #2283 to land before opening the PR so it builds on the new layout. Happy to adjust scope if you had something different in mind.

randomstuff:
Sounds good to me. There might some delays before the mere of #2283, though.

I'd suggest using a separate separate PR for each threat: easier to review, easier to merge. So maybe start with one and we'll see where it goes.

ArockiaRajamanickam:
Makes sense, one per threat it is. I've opened #2317 with the first one, the `jwk`/`jku`/`x5u` header parameters.

Since you mentioned #2283 may take a while, I wrote it against the current `## Issues` layout using the existing Symptom / How to Prevent headings, so it can go in independently. If #2283 lands first I'll happily reshape it into the new format, or rebase, whichever suits you.

I'll hold off on issuer confusion, audience confusion and token type confusion until you've had a look at that one, in case you want a different shape or level of detail.

randomstuff:
#2283 merged (without the part about key type vulnerability which has its own PR #2320)

mackowski:
**Progress update.** Three of the five threats listed here have landed:

- **Key type confusion** — #2320 ([`bf62cf0d`](https://github.com/OWASP/CheatSheetSeries/commit/bf62cf0d)), as a full section under `Threats on JWTs`.
[...]

Issue #2266, opened 2026-06-30
Enforce US English in CI via textlint (phased: corpus cleanup first, then enable the rule)


mackowski:
## Summary

We require **US English** but do not enforce it. British spellings regularly slip into merged cheat sheets. This proposes enforcing US English through the textlint check we already run in CI, in **two phases** so we don't break unrelated PRs:

1. **Phase 1 — cleanup PR(s):** fix the existing British spellings in the corpus.
2. **Phase 2 — enable enforcement:** add British→US mappings to the textlint terminology rule so CI flags them going forward.

Phase 2 must come **after** Phase 1, because the lint job lints whole files (not the PR diff), so the rule cannot go green until the existing debt is fixed.

## Where the rule already lives

- [`CONTRIBUTING.md` → "Language"](https://github.com/OWASP/CheatSheetSeries/blob/master/CONTRIBUTING.md#language): *"Use US English. Spell check before submitting a PR."*
- `AGENTS.md` (proposed in #2265) restates it: *"US English. Spell-check…"*

So this is **enforcing an existing rule**, not introducing a new one.

## Evidence / motivation

- The recently merged `AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.md` contains 7 British spellings (`programmes`, `authorised`, `organisation`, `behaviour`, `normalisation`, `minimisation`, `defence`) — none caught by CI or review.
- Current British-spelling debt across the 120 cheat sheets: **~66 occurrences in 26 files (~22% of the corpus)**. Examples:

| Term | Occurrences | Files |
|------|------------:|------:|
| behaviour | 16 | 6 |
| defence | 10 | 7 |
| sanitis(e/ed/ation) | 8 | 3 |
| organis(e/ation) | 6 | 5 |
| minimis(e/ation) | 4 | 4 |
| whilst | 4 | 4 |
| authoris(e/ation) | 3 | 3 |
| modelling | 3 | 2 |
| (+ colour, catalogue, favour, normalise, optimise, analyse, grey, amongst, initialise) | | |

- We already maintain British→US pairs in [`.textlintrc`](https://github.com/OWASP/CheatSheetSeries/blob/master/.textlintrc) (e.g. `analysed`→`analyzed`, `localised`→`localized`, `cancelling`→`canceling`), so the corpus only passes today because *those specific* forms were fixed. This is just extending an established pattern.

## Why CI can't simply be flipped on

[...]

PR #2317, opened 2026-07-31
Add JWT threat: trusting key material named in the token header

ArockiaRajamanickam:
First of the threats listed in #2302, split out on its own as @randomstuff suggested.

Adds one subsection to `## Issues`: **Trusting Key Material Named in the Token Header**. It covers the `jwk`, `jku`, `x5u` and `x5c` header parameters, which let a token nominate the key that verifies it, plus the `kid`, `x5t` and `x5t#S256` selection hints. Both consequences are described: acceptance of self-signed tokens, and SSRF through the URL-valued parameters, which links to the SSRF cheat sheet rather than restating it.

A note on the base. This is written against the current `## Issues` layout with the `#### Symptom` / `#### How to Prevent` headings the other threats use, rather than against #2283, since you mentioned that merge may be a while. If #2283 lands first I am happy to rewrite it into the new prose plus Mitigations plus References shape, or to rebase, whichever is less work for you.

Sourcing: [RFC 8725 "Do Not Trust Received Claims"](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim) for the `kid` injection and `jku`/`x5u` SSRF guidance, and [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114) as a concrete case of a library trusting a key embedded in the header. I read both rather than working from memory, and confirmed the RFC anchor resolves to the section named and that the CVE description matches the claim via the NVD API.

Checks: `npm run lint-markdown` and `npm run lint-terminology` are clean. The link check reports two 429s, on the spiffe and tink links — both are pre-existing on master and both are GitHub rate-limiting rather than genuinely dead; neither is a link this PR adds.

Remaining from #2302 once you have had a look at this one: issuer confusion, audience confusion, and token type (cross-JWT) confusion. I will send them as separate PRs in the same shape.

    the contents and I affirm the results. The LLM used is `Claude Fable 5 (via Claude Code)`
    and the prompt used is `Write the "blindly trusting the jwk, jku, x5u, x5t header parameters" threat from OWASP CheatSheetSeries issue #2302 as a single subsection of the JWT cheat sheet, matching the existing Symptom / How to Prevent structure. Ground it in RFC 8725, fetch and read every source before citing it, cover both the self-signed token and SSRF consequences, and link to the SSRF cheat sheet instead of duplicating it.` I have independently verified every citation and technical claim against the cited source.

This PR fixes part of issue #2302.


cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -795,6 +795,24 @@ The simplest way to prevent this attack is to ensure that the secret used to sig
 
 Alternatively, consider the use of tokens that are signed using a digital signature (public-key cryptography) rather than using an HMAC and secret key.
 
+### Trusting Key Material Named in the Token Header
+
+#### Symptom
+
+A JWS header can carry the verification key itself or a pointer to it: `jwk` (an embedded public key), `jku` (a URL to a JWK Set), `x5u` (a URL to an X.509 certificate), and `x5c` (an embedded certificate chain), alongside the key selection hints `kid`, `x5t` and `x5t#S256`. An application that resolves or selects its verification key from these header parameters can be steered into trusting a key the attacker controls, because the header is unauthenticated attacker input. The token ends up vouching for its own key.
```
randomstuff: > `jwk` (an embedded public key)

An attacker might attempt to use a secret key here as well. Maybe some implementation out there might be vulnerable to that.

randomstuff: > The token ends up vouching for its own key.

I am not certain this sentence adds anything compare to the previous ones. I'd consider dropping it.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -795,6 +795,24 @@ The simplest way to prevent this attack is to ensure that the secret used to sig
...
+
+An attacker strips the original signature, puts their own public key in `jwk`, or points `jku` or `x5u` at a JWK Set they host, then signs the modified token with the matching private key. A verifier that trusts the key it just read from the token accepts the forgery. [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114) is a real instance of this: the affected library trusted a public key embedded in the JWS header and so accepted tokens re-signed by anyone.
```
randomstuff: More generally I would say that the attacker "forges its own token, includes their own public key […]"

randomstuff: Reference to CVE-2018-0114 is good. I would add a reference to the CVE maybe at the end (questionable).

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -795,6 +795,24 @@ The simplest way to prevent this attack is to ensure that the secret used to sig
...
+
+Because `jku` and `x5u` are arbitrary URLs that the server fetches, they are also a [server-side request forgery](Server_Side_Request_Forgery_Prevention_Cheat_Sheet.md) vector, including through hostnames that resolve to internal addresses.
+
+#### How to Prevent
+
+Do not take the verification key from the token. Select it only from trust material that is bound to the issuer out of band, such as a pinned key or the `jwks_uri` published in the issuer's metadata.
+
+Ignore `jwk`, `jku`, `x5u` and `x5c` when making trust decisions, or disable them in your library. Treat `kid`, `x5t` and `x5t#S256` purely as a hint for choosing among keys you already trust, and validate or sanitise `kid` before using it in a lookup, since it also reaches databases and directories as an [injection](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim) vector.
```
randomstuff: `x5`  might be used in some contexts assuming it is anchored by an already trusted anchor. Ditto for `x5u`. More generally, I would have a more nuance wording than just "ignore". These parameters are here for some reasons and there are arguably legitimate use cases when we should not ignore them 😄

randomstuff: We shoule mention the correct and typical use case of using the `kid` parameter to identify a key from an already configured JWKS URL.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -795,6 +795,24 @@ The simplest way to prevent this attack is to ensure that the secret used to sig
...
+
+If your deployment genuinely has to fetch keys by URL, allowlist the exact hostnames, require HTTPS, send no cookies or credentials with the request, and reject URLs that resolve to loopback, link-local or private address ranges. See the [Server Side Request Forgery Prevention Cheat Sheet](Server_Side_Request_Forgery_Prevention_Cheat_Sheet.md) for the wider problem.
```
randomstuff: Yes OK. Maybe we can just point to the SSRF cheat sheet without addiing any details on this cheat sheet?

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -795,6 +795,24 @@ The simplest way to prevent this attack is to ensure that the secret used to sig
...
+
+An attacker strips the original signature, puts their own public key in `jwk`, or points `jku` or `x5u` at a JWK Set they host, then signs the modified token with the matching private key. A verifier that trusts the key it just read from the token accepts the forgery. [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114) is a real instance of this: the affected library trusted a public key embedded in the JWS header and so accepted tokens re-signed by anyone.
+
+Because `jku` and `x5u` are arbitrary URLs that the server fetches, they are also a [server-side request forgery](Server_Side_Request_Forgery_Prevention_Cheat_Sheet.md) vector, including through hostnames that resolve to internal addresses.
+
+#### How to Prevent
+
+Do not take the verification key from the token. Select it only from trust material that is bound to the issuer out of band, such as a pinned key or the `jwks_uri` published in the issuer's metadata.
```
randomstuff: > Do not take the verification key from the token.

Maybe something along these lines: "Do not take the verification key from the token unless this key can be trusted through some chain of trust to a root trust anchor associated with the issuer."

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
randomstuff: I would formulate it like this:"An attacker *can* …"

randomstuff:
I made some comments.

I wait for https://github.com/OWASP/CheatSheetSeries/pull/2283 before merging. As you noted, some rework might be needed to align with the form of #2283.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Good point. Changed to `jwk` (an embedded key) and added a sentence noting an attacker can try to smuggle a symmetric key through the same parameters hoping it gets used for MAC verification.

ArockiaRajamanickam: Dropped.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Reworded to the more general form: the attacker forges their own token and includes their own key, rather than stripping the signature off an existing one.

ArockiaRajamanickam: Kept it inline where the claim is made, since this cheat sheet's `## References` section is document-wide rather than per-threat. Happy to move or duplicate it there if you'd rather.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Agreed, that was too absolute. Reworked so the distinction is anchoring rather than avoidance: these parameters are fine where the chain validates up to an anchor already trusted for that issuer, and the thing to avoid is treating them as the source of trust instead of a pointer within it.

ArockiaRajamanickam: Added: `kid`, `x5t` and `x5t#S256` are described as the normal way to choose which key from an already configured JWKS should verify a token. The sanitisation advice now sits alongside that rather than replacing it.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Trimmed to a single pointer at the SSRF cheat sheet, with the allowlisting and address-range detail removed.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Done, switched to "An attacker can forge…" and the same for the symmetric-key sentence.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
ArockiaRajamanickam: Used your wording, near enough verbatim: the key may be taken from the token where it can be tied through a chain of trust to a root trust anchor associated with the issuer.

ArockiaRajamanickam:
Thanks, that's a much better section than what I sent.

The main thing you caught is that I'd written it as "don't use these parameters", which is wrong. `x5c` and `x5u` validated up to an anchor you already trust for that issuer are legitimate, and `kid` picking a key out of a configured JWKS is the normal case rather than an edge case. I've reworked the section so the distinction is anchoring, not avoidance: the problem is treating these parameters as the source of trust instead of as a pointer within it.

Also applied: the symmetric-key variant of the `jwk` abuse, the more general "an attacker can forge their own token" phrasing, your wording for the chain-of-trust caveat, dropped the redundant sentence, and cut the SSRF paragraph down to a plain pointer at that cheat sheet.

The one I didn't act on is the CVE placement. I left it inline where the claim is made, since `## References` here is document-wide rather than per-threat, but say the word and I'll move or duplicate it.

Two commits so it's easy to see what changed against your comments.

ArockiaRajamanickam:
#2283 landed, so I've rebased onto it and rewritten the section in its format: prose, then `Mitigations:`, then `References:`. The branch is now a single commit against current master, so the diff is just the new subsection.

Your review points are all carried over into the new wording:

- The parameters are no longer described as things to ignore. The section now says the distinction is anchoring rather than avoidance, and calls out `x5c`/`x5u` validated to an already trusted anchor, and `kid`/`x5t`/`x5t#S256` selecting a key from a configured JWKS, as the legitimate cases.
- Added the symmetric key variant of the `jwk` abuse.
- "An attacker can forge their own token…" rather than the strip-and-resign narration.
- Your chain-of-trust wording is the first mitigation bullet.
- The redundant sentence is gone, and the SSRF paragraph is now just a pointer at that cheat sheet.
- The CVE moved into a `References:` block, which is what you suggested and now has a natural home in the new format.

Also worth noting: the merged #2283 doesn't include the "Key type confusion" section that was in the branch earlier, so this section currently follows "Unsecured JWTs" directly. Say the word if you'd rather it sat somewhere else in the order.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -196,6 +196,26 @@ Mitigation:
...
+
+These parameters have legitimate uses, so the distinction is anchoring rather than avoidance. `x5c` and `x5u` are usable where the certificate chain validates up to an anchor already trusted for that issuer, and `kid`, `x5t` and `x5t#S256` are the normal way to choose which key from an already configured JWKS should verify a given token. What must not happen is treating any of them as the source of trust rather than as a pointer within it.
+
+Mitigations:
+
+- Do not take the verification key from the token unless that key can be tied, through a chain of trust, to a root trust anchor associated with the issuer.
+- Prefer trust material established out of band, such as a pinned key or the `jwks_uri` published in the issuer's metadata.
+- Validate or sanitise `kid` before using it in a lookup, since it also reaches databases and directories as an injection vector.
```
randomstuff: We are aligning to US English wording (see #2266). For consistency, that would be "sanitize".

randomstuff:
LGTM except for one cosmetic issue. :+1:

I'd wait for a second review before merging though. (@jmanico, @mackowski)

ArockiaRajamanickam:
Fixed, thanks — that's the only instance in the section, and I checked the rest of it for other British spellings while I was there.

No rush on the second review.

ArockiaRajamanickam:
Rebased onto master. #2320 had landed a "Key type confusion" section in the same place, so this had gone conflicting; my section now sits after yours, which reads better anyway since it moves from confusing one key for another to trusting a key the token names.

Net diff against master is still the same 20 lines in one file. textlint and markdownlint are clean against the current config, including the US English enforcement from #2338.

randomstuff (approved):

jmanico (approved):

Issue #2313, opened 2026-07-28
Proposal: Verifying Third-Party Agent Execution Evidence Cheat Sheet


astrogilda:
I would like to propose a short cheat sheet, and I would rather check the idea here before writing anything.

When a team adopts a third-party agent, MCP server or skill, they usually end up holding some record of what it did: a log, a telemetry export, a scan result. The reviewer's question is whether that record can be trusted, and often it cannot, for a reason that is invisible on the page. The record was produced by the same process the record is about. This is not hypothetical. Products in this space ship fail-open telemetry, execution logs the observed process can write to, and receipts sealed with a shared secret so anyone who can verify can also forge.

What I think would help is a short set of questions a reviewer can ask about a record they have been handed, with the failure each one catches. Six of them, about three pages, and I would rather cut than add. Things like: could the observed component have altered this, was it watched during execution or reconstructed afterwards, is the delivery path fail-open, and can it be verified by someone who cannot also produce it.

It sits next to the MCP and agentic material rather than overlapping it, since those cover what can go wrong and this covers how you would know. The closest fit in the series is Section 10 of the MCP Security Cheat Sheet, Monitoring, Logging and Auditing, whose five bullets ask for logging, SIEM ingestion, alerting, redaction and audits, and none of which asks whether the record can be trusted given who produced it.

I would rather co-route this through the GenAI project than write it solo, and I am happy for it to land under whatever ownership you prefer. If someone already working on the agentic entries wants to co-author, I would prefer that. For disclosure: I maintain in-toto/attestation#570 and its conformance suite, so I have an interest here. Happy to be told it is out of scope.

Edited 2026-08-20 to point at the right document. This named MCP08 in the MCP Top 10, which does require tamper-evident logging with cryptographic hashing and write-once storage, so the sentence was wrong about that page. The gap is in the series' own MCP Security Cheat Sheet, Section 10, which is where @mackowski checked it below.

Rul1an:
Worth a line under your last question, the one about whether a record can be verified by someone who cannot also produce it, because the same problem reappears one level up. It applies to the corpora people check records against, not only to the records.

Agreement between implementations is weak evidence when they share an author or a lineage. And an N-of-N score reports how many cases someone thought to write, not how much of the normative surface is pinned, so a clean sheet can mean the suite is thorough or that it is small.

Six questions and three pages is the right size, so this may well be one of the things you meant to cut. If it goes ahead I would be glad to help with drafting.

[...]

PR #2324, opened 2026-08-03
ci: add Index.md drift check

sedat4ras:
Closes #2273

The published website already regenerates `Index.md` at site-build time (`scripts/Generate_Site_mkDocs.sh` runs `Update_CheatSheets_Index.py` on every deploy), so this drift only affects the copy committed in the repo — which is what people browsing GitHub see. This PR keeps that committed copy honest:

- **Regenerates `Index.md`** to match the current contents of `cheatsheets/` (92 → 120 cheat sheets), so the check lands green.
- **Adds a standalone CI workflow** (`index-drift-check.yml`) that runs the existing `Update_CheatSheets_Index.py` on PRs touching `cheatsheets/**`, `Index.md`, or the script itself, and fails with an actionable error if the committed `Index.md` diverges from the regenerated output.

Deliberately does not build on #2247 — it runs the script as-is, per maintainer guidance in the issue.

Tested locally: passes on a freshly regenerated index; fails as expected when a cheat sheet is added without updating `Index.md`.

PR #2332, opened 2026-08-04
Add Verifying Third-Party Agent Execution Evidence Cheat Sheet

astrogilda:
Here is the new cheat sheet, Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md, proposed and approved in #2313. Thank you @mackowski for testing my claim against the MCP Security sheet yourself instead of taking my word for it, and @jmanico for the go ahead. The gap you found is exactly what the new file is for: MCP Security Section 10 and AI Agent Security Section 6 both say to log every tool invocation, and both of those instructions are met by a log the observed component wrote about itself. Nowhere does the series yet ask whether such a log deserves belief, given who produced it.

The sheet hands reviewers a checklist, not an implementation guide. Six questions to ask about a record a supplier gave you, each paired with the failure it catches, and three outcomes to grade the answers with: independently re-checkable, supplier assertion, or no information about that execution. The questions ask who could have written or altered the record, whether anybody captured it while the work ran or assembled it afterwards, and whether its subject is named precisely enough to re-check. They also ask whether an omission would be visible to you, what became of the evidence once recording or delivery failed, and whether a party able to check the artifact could equally well manufacture one. Those six behave identically against a plain signed JSON log, an in-toto Statement, a COSE receipt, or a proprietary vendor export, and no mechanism it names anywhere becomes a requirement.

Scope here is one file and nothing besides, since Index.md needs no entry, which I checked instead of assuming. Your site build runs Update_CheatSheets_Index.py, regenerating the index from whatever sits in the directory, and mkdocs.yml points docs_dir at cheatsheets with no nav key at all. I put the file straight into cheatsheets, and relocating it to cheatsheets_draft costs one line if you would rather it started there. The committed Index.md already lags the directory badly, so hand editing would only leave a conflict for whoever rebuilds it next.

Your own toolchain ran before I opened this, driven from the package.json in the repo. Markdown lint and terminology lint both exit 0, and the link checker under your config reports 21 of 21 good, a count that includes three relative links into sibling cheat sheets. I also walked the citation density workflow by hand and found an inline citation under all five H2 headings, while the scope workflow sees a single cheat sheet touched and 136 lines added.

Every quotation came from the primary document, never from somebody's summary. Local copies of all ten sources sit on my disk, and I matched each quoted string against them character for character. The habit earned its keep here: an earlier draft cited slsa.dev/spec/v1.0/verifying-artifacts, which answers 200 and yet carries a Retired banner across the top of the page, and a link checker cannot see a banner. That citation now points at v1.2, which is approved and carries the same wording.

@Rul1an raised something in the issue thread that deserves credit. Question 6 asks whether a record can be checked by a party who cannot also produce one, and he noticed the identical problem returning a level higher up, among the conformance corpora people check records against. Agreement between implementations is weak evidence when they share a lineage, and a spotless score reports how many cases somebody thought to write down. He is right, and I drafted a passage on corpora before cutting it, because the sheet was growing past what any reviewer will sit through and corpora are really their own subject. The passage can come back, and I would far prefer his co-authorship on it over paraphrasing his argument back at him.

What falls outside gets stated plainly, so Section 7 of MCP Security keeps ownership of the live channel between agent and server. Runs spanning several parties are the harder case, because the failure surfaces at the seams: one agent logs a delegation, its counterparty holds nothing showing receipt, and no question about a single artifact reaches across the gap. A second sheet covering delegation across parties is the natural follow up.

A note on the format rules, since it is the box most likely to draw an argument. Your style guide asks that quotations from other articles use blockquote syntax. Mine are short fragments sitting inside sentences, where a blockquote would snap the surrounding sentence in half, so I followed existing practice: Software Supply Chain Security quotes NIST SP 800-204D and SLSA identically at lines 5 and 136. Happy to lift mine into blockquotes if you would prefer reading them that way.

- [x] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](../templates/New_CheatSheet.md).
- [ ] All your assets are stored in the **assets** folder. Left unticked because the change adds no assets.
- [ ] All the images used are in the **PNG** format. Left unticked because the change adds no images.
- [ ] The CI build of your PR pass, see the build status [here](https://github.com/OWASP/CheatSheetSeries/actions). Left unticked because CI has not run yet at the moment I press the button. The local equivalents are green and I will confirm here once your build finishes.

- [x] I have used AI tools to generate the contents of this PR. I have verified the contents and I affirm the results. The LLM used is `Claude Opus 4.8` and the prompt used is `tighten this OWASP cheat sheet that I generated on verifying third-party agent execution evidence for this repo`. I have independently verified every citation and technical claim against the cited source.

This PR fixes issue #2313.


astrogilda:
Thanks @mackowski and @jmanico for the ack on #2313, and sorry to bump it.

The PR has been open since the 4th and every workflow on it is sitting behind the first-contribution approval gate. Five runs exist against head 2225dd7c, created within seconds of the pull request, and all five completed with conclusion `action_required`: Build website, PR Scope Check, Citation Density Check, Markdown Link Check and Markdown Lint Check. None of them is queued and none has executed, so there is nothing green or red to read yet. The `author_association` on this pull request reads NONE, which is what puts them there. One press of approve and run releases all five together, and I would like it green before anyone spends review time on it.

I ran the repo's own checks against the branch in the meantime. markdownlint and textlint came back clean. The citation check passes on all five of the new sections, and the scope check is fine at one file and 136 lines. The only thing link-check flagged was a 429 on the ASVS page, which answers 200 on a direct fetch, so I think that's rate limiting. I cite that page four times and can cut it to one if it keeps tripping.

Happy to fix whatever CI reports.

Edited 2026-08-20. This said no workflows had run and that there were zero runs. Five runs exist against head 2225dd7c and have done since the pull request opened; all five are held at action_required behind the first-contribution approval, which from the outside looks the same as an empty queue.

astrogilda:
All five workflow runs on this pull request sit at `action_required`. That is the first-contribution approval GitHub asks for, and author_association on this one reads NONE, so a single press of approve and run releases Build website, PR Scope Check, Citation Density Check, Markdown Link Check, and Markdown Lint Check together. None of them is queued. I raise it because from the outside the two states look identical.

Since it has been a couple of weeks I re-checked what rots while a branch waits. All eighteen external links in the sheet resolve today. The citation density workflow wants an inline `[text](https://...)` under every added H2, and each of the five sections has one. The scope workflow sees one cheat sheet touched and 136 net additions against MAX_CHEATSHEETS_TOUCHED: 3 and MAX_NET_ADDITIONS: 1500. The branch still merges onto master with no conflict.

Two choices are yours and neither costs me anything. If a new sheet should begin life in `cheatsheets_draft`, that is a one line move and I will push it whenever you say. If you would prefer the short quoted fragments lifted into blockquotes, I will change them. They sit inline today to match Software Supply Chain Security, which quotes NIST SP 800-204D at line 5 and SLSA 1.0 at line 136 the same way.

@mackowski, thank you for going and checking the claim against the MCP Security sheet yourself, and @jmanico for the go ahead on #2313.

jmanico (approved):

andrzejsydor (approved):

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
+# Verifying Third-Party Agent Execution Evidence Cheat Sheet
+
+## Introduction
+
+A team that adopts an agent, an MCP server, or a skill it did not build ends up holding a record of what that component did, and someone has to say what that record establishes. [MCP Security](https://cheatsheetseries.owasp.org/cheatsheets/MCP_Security_Cheat_Sheet.html#10-monitoring-logging-auditing) and [AI Agent Security](https://cheatsheetseries.owasp.org/cheatsheets/AI_Agent_Security_Cheat_Sheet.html#6-monitoring-observability) both tell you to log every tool invocation, and each of those instructions is satisfied by a record the observed component wrote about itself. This cheat sheet covers the other side of that exchange, where the record was produced by the same party it describes. It gives a reviewer six questions to ask of such a record, and a way to score the answers.
```
mackowski: **Expand MCP on first use, and unpack the opening.** Every other sheet in the series expands it (`MCP_Security_Cheat_Sheet.md:5`, `AI_Agent_Security_Cheat_Sheet.md:32`). "The other side of that exchange" is also doing a lot of unexplained work here, and the passive "is satisfied by" hides who acts.

```suggestion
A team that adopts an agent, a Model Context Protocol (MCP) server, or a skill it did not build ends up holding a record of what that component did, and someone has to say what that record establishes. [MCP Security](https://cheatsheetseries.owasp.org/cheatsheets/MCP_Security_Cheat_Sheet.html#10-monitoring-logging-auditing) and [AI Agent Security](https://cheatsheetseries.owasp.org/cheatsheets/AI_Agent_Security_Cheat_Sheet.html#6-monitoring-observability) both tell you to log every tool invocation, and you can satisfy both instructions with a record the observed component wrote about itself. This cheat sheet covers what happens next: how to judge a record whose producer is the same party it describes. It gives a reviewer six questions to ask of such a record, and a way to score the answers.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Three Review Outcomes
+
+Score each question below with one of three outcomes, and score the record no higher than its weakest answer. [MCP08:2025 Lack of Audit and Telemetry](https://owasp.org/www-project-mcp-top-10/2025/MCP08-2025%E2%80%93Lack-of-Audit-and-Telemetry) recommends "cryptographic hashing (HMAC, SHA-256) to log files for integrity" and "audit trail self-verification, where logs cross-reference session data for consistency". Both are sound for an operator who holds the logs and the key. A reviewer holds neither.
```
mackowski: **This is the one change I'd most like to see.** As written, "score the record no higher than its weakest answer" combined with line 113's "an unsigned export gives no information" grades a typical SaaS vendor JSON log at the bottom outcome, whose column reads "Nothing." That contradicts your own line 14, which grants supplier assertions operational value. A developer applying the rule literally concludes their existing telemetry is worthless and stops reading. Scoping the rule to *disputes with the supplier* keeps the argument intact without that side effect.

```suggestion
Score each question below with one of three outcomes. The weakest answer bounds what the record can settle in a dispute with the supplier; it does not bound what the record is worth operationally. [MCP08:2025 Lack of Audit and Telemetry](https://owasp.org/www-project-mcp-top-10/2025/MCP08-2025%E2%80%93Lack-of-Audit-and-Telemetry) recommends "cryptographic hashing (HMAC, SHA-256) to log files for integrity" and "audit trail self-verification, where logs cross-reference session data for consistency". Both are sound for an operator who holds the logs and the key. A reviewer holds neither.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+| Outcome | What it means | What it will carry |
+|---|---|---|
+| Independently re-checkable | A party who cannot produce the record can confirm it, from material the producer does not control when the check runs. | A disagreement with the supplier. |
+| Supplier assertion | A claim by an interested party. It may well be true, and nothing in the record establishes it. | Operational awareness, and anything the supplier has no incentive to misstate. |
+| No information about this execution | The record cannot distinguish a clean run from no run at all. | Nothing. |
```
mackowski: **Table header and first cell.** "What it will carry" is an unexplained metaphor, and the cell "A disagreement with the supplier." is not comprehensible standing alone — it reads as though the record *is* a disagreement. Also switches line 14's contrastive "and" to "but" (a tic that recurs at lines 42 and 102, and reads as an error to non-native readers).

```suggestion
| Outcome | What it means | What the record can support |
|---|---|---|
| Independently re-checkable | A party who cannot produce the record can confirm it, from material the producer does not control when the check runs. | Enough weight to settle a disagreement with the supplier. |
| Supplier assertion | A claim by an interested party. It may well be true, but nothing in the record establishes it. | Operational awareness, plus anything the supplier has no incentive to misstate. |
| No information about this execution | The record cannot distinguish a clean run from no run at all. | Nothing. |
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+Three activities get called verification, and this sheet keeps them apart: reading a record or re-running the producer's own tool over it is re-checking, watching an event as it happens is witnessing, and re-checking offline as a party who trusts neither the component nor its operator is what these questions test. The [Logging Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html#verification) uses the word in its own sense, of testing that logging works as specified, and both senses are in ordinary use across this series.
```
mackowski: **56-word sentence with three comma-joined definitions**, and the third clause breaks the parallel structure (term-last instead of term-first), so a reader has to backtrack. Naming each term in bold and splitting off the last clause fixes it without losing anything.

```suggestion
Three different activities are called verification, and this sheet keeps them apart: **re-checking** is reading a record or re-running the producer's own tool over it; **witnessing** is watching an event as it happens; **independent re-checking** is checking offline as a party who trusts neither the component nor its operator. The six questions below test the third. The [Logging Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html#verification) uses the word in its own sense, of testing that logging works as specified, and both senses are in ordinary use across this series.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Six Questions to Ask About an Execution Record
+
+These are questions about a record as an artifact, so they apply to a plain signed JSON log, a structured attestation such as an [in-toto Statement](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md), a [COSE receipt](https://datatracker.ietf.org/doc/html/rfc9942), or a supplier's proprietary export. Each closes with the outcome that common answers map to.
```
mackowski: **Expand COSE on first use** — same rule as MCP above.

```suggestion
These are questions about a record as an artifact, so they apply to a plain signed JSON log, a structured attestation such as an [in-toto Statement](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md), a [COSE (CBOR Object Signing and Encryption) receipt](https://datatracker.ietf.org/doc/html/rfc9942), or a supplier's proprietary export. Each closes with the outcome that common answers map to.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+- Count parties, not processes. A recorder the supplier operates is one party however many hosts it spans.
+- Look for an input the component did not choose. A recorder fed only by the component records the component's report.
+- Check whether the signing service will sign arbitrary content on the component's request.
+- If it will, that signature is worth what a readable key is worth.
+- Look for a copy landing where the component cannot change it. [ASVS 16.4.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md) asks that logs "cannot be modified".
+
+**Outcome.** Most agent telemetry shipping today is a supplier assertion, and it is worth saying so plainly: SDK-emitted spans, a framework's own traces, an MCP server's logs, and a hosted provider's export are all fed by the component and handled downstream by a single party. Independently re-checkable needs an input the component did not choose and a party with no stake in the answer. A supplier who will not describe the path scores no information. Where the answer is supplier assertion and you still have to proceed:
```
mackowski: **Two small things.** "and it is worth saying so plainly" is meta-commentary the list right after it already earns. And `**Outcome.**` should be `**Outcome:**` — colon-style bold lead-ins are the series convention (`**Example:**` appears 57 times across `cheatsheets/`; the period style appears nowhere). Same change applies at lines 53, 67, 84, 98, and 113.

```suggestion
**Outcome:** Most agent telemetry shipping today is a supplier assertion: SDK-emitted spans, a framework's own traces, an MCP server's logs, and a hosted provider's export are all fed by the component and handled downstream by a single party. Independently re-checkable needs an input the component did not choose and a party with no stake in the answer. A supplier who will not describe the path scores no information. Where the answer is supplier assertion and you still have to proceed:
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Corroborate against a record your own side produces, such as a gateway or egress log covering the same calls.
+- Contract for a copy delivered to a party the supplier cannot edit.
+- Size the decision so this record is not its only basis.
```
mackowski: **The leverage gap, and the one bullet that currently dead-ends.** Several bullets across the sheet assume procurement power a typical team does not have — "contract for a copy" here, "ask for a sandbox tenant" (line 82), "get the behavior in writing" (line 90), "ask for gap markers from a real outage" (line 95). A team pulling a community MCP server off GitHub or signing up for a hosted agent on a credit card gets none of it, and the sheet never says so.

Separately, "Size the decision so this record is not its only basis" is the exit path most readers take (by your own line 34, most telemetry lands in supplier assertion) and it terminates in a non-instruction. Naming the decision classes makes it actionable.

The gateway/egress point is the one move that works with zero supplier cooperation — it's currently mentioned obliquely twice (lines 36, 82) and deserves to be named as the fallback.

```suggestion
- Corroborate against a record your own side produces, such as a gateway or egress log covering the same calls. Where you have no leverage over the supplier at all, this is the move that always remains: run the component behind a gateway, proxy, or egress point you control, and record there.
- Contract for a copy delivered to a party the supplier cannot edit. This needs procurement leverage that a team adopting a community component will not have.
- Size the decision so this record is not its only basis. A supplier assertion is adequate for debugging, capacity planning, and detecting non-adversarial failure; it is not adequate as the sole basis for incident attribution, a customer-facing breach statement, a compliance attestation, or a contractual dispute.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+### 2. Was the record witnessed while the work ran, or assembled afterwards?
+
+A record assembled after execution, from state the component still controls, inherits every weakness of that state, and the artifact looks identical either way. Capture during the run narrows the window in which an account can be composed to fit a known outcome.
+
+- Treat a timestamp applied over a finished record as a bound on how late it was written.
+- A timestamp field inside the record is an assertion by the producer.
+- [RFC 3161](https://datatracker.ietf.org/doc/html/rfc3161) countersigning is stronger, but the authority "may be operated as a Trusted Third Party (TTP) service, though other operational models may be appropriate".
+- Apply question 1 to whoever operates that authority.
+- Look for a bound in the other direction, since contemporaneity is a claim about how early the record was fixed.
+- [RFC 9334 Section 10](https://datatracker.ietf.org/doc/html/rfc9334#section-10) gives three approaches. Send a nonce to be signed into the record, and on a match "the appraising entity knows that the Claims were signed after the nonce was generated".
```
mackowski: **The direction of your bound is right, but not its scope.** A nonce proves the record was *signed* after the nonce was generated. It does not establish that the entries were captured while the work ran — which is exactly the threat this question names ("an account composed to fit a known outcome"). RFC 9334 §10.2 says so directly, right after the sentence you quote: the epoch is "rough" because "[t]he epoch applies to the entire Claim set instead of a more granular association" and "[t]he time between the creation of Claims and the collection of Claims is indistinguishable."

Also swaps "contemporaneity" for plainer wording.

```suggestion
- Look for a bound in the other direction, since a claim that the record was made during the run is a claim about how early it was fixed.
- [RFC 9334 Section 10](https://datatracker.ietf.org/doc/html/rfc9334#section-10) gives three approaches. Send a nonce to be signed into the record, and on a match "the appraising entity knows that the Claims were signed after the nonce was generated".
- RFC 9334 calls this epoch "rough": a nonce dates the signature, not the entries, since "the time between the creation of Claims and the collection of Claims is indistinguishable".
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+- A challenge issued mid-run also reaches components that never produce a discrete finished document.
+- Read whether the record covers one execution or several, and whether per-step material survives.
+
+**Outcome.** A countersignature from an authority the supplier does not operate, or a nonce you supplied, makes the timing independently re-checkable. A producer-asserted timestamp, or a countersignature from the supplier's own authority, is a supplier assertion. No timing evidence at all, or entries aggregated across runs with nothing per-step surviving, gives no information about this execution.
```
mackowski: Follows from the previous comment: as written this grades "a nonce you supplied" as making *the timing* independently re-checkable, which claims more than RFC 9334 supports.

```suggestion
**Outcome:** A countersignature from an authority the supplier does not operate, or a nonce you supplied, makes the *signing* time independently re-checkable. That bounds when the record was fixed; it does not by itself establish that the entries were captured as the work ran. A producer-asserted timestamp, or a countersignature from the supplier's own authority, is a supplier assertion. No timing evidence at all, or entries aggregated across runs with nothing per-step surviving, gives no information about this execution.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+### 3. What is the record about, and is that thing named precisely enough to re-check?
+
+A correctly signed record whose subject is named loosely, as "the agent" or "the latest build", can be presented next to a different artifact than the one it describes. Nothing is forged and no signature fails; the reviewer draws a conclusion about the thing in front of them from evidence about something else.
+
+- Look for the subject identified by content, using a cryptographic digest of the artifact, image, configuration, or transcript described.
+- The [in-toto Statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) is one worked example: every subject element "MUST have `digest` set" and subjects "are matched purely by digest".
```
mackowski, line 62: **No strength floor on the digest.** In the adversarial-supplier model this sheet assumes, a collision lets one record cover two artifacts — and the framework you cite as the worked example explicitly permits `md5`, `sha1`, and `gitCommit` (SHA-1 form). The DigestSet spec already says the right thing, so this is just surfacing it.

```suggestion
- Look for the subject identified by content, using a cryptographic digest of the artifact, image, configuration, or transcript described.
- Check which algorithm that digest uses. The [in-toto DigestSet](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md) permits `md5`, `sha1`, and Git SHA-1 object IDs alongside stronger options, tells consumers they "MUST only accept algorithms that they consider secure", and recommends "at least `sha256`". Treat SHA-256 as the floor.
- The [in-toto Statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) is one worked example: every subject element "MUST have `digest` set" and subjects "are matched purely by digest".
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+### 4. If something was left out, would you be able to tell?
+
+Reviewers read absence as evidence: no denied call means nothing was denied. That reading holds only when the recorder was positioned to see the event and could not decline to write it. A component able to suppress one entry produces a clean record and a clean summary.
+
+- Check whether the record states its window. A window you inferred from the entries present is not a stated one.
+- Treat a transparency log as detection of removal after entry. It says nothing about an entry that was never made.
+- [RFC 9162](https://datatracker.ietf.org/doc/html/rfc9162) describes an append-only log built on Merkle Trees, which can prove that any particular instance "of the log is a superset of any particular previous instance".
+- [Sigstore's Rekor](https://docs.sigstore.dev/logging/overview/) is one deployed instance of such a log.
```
mackowski, line 79: **The strongest correctness point in my review.** The sheet's whole method is "count the parties who would have to cooperate" — and the log is one of them. RFC 9162 §1 says this in the paragraph immediately after the sentence you quote:

> The log auditing mechanisms described in this document can be circumvented by a misbehaving log that shows different, inconsistent views of itself to different clients. Therefore, it is necessary to treat each log as a trusted third party. While mechanisms are being developed to address these shortcomings [...] such mechanisms are outside the scope of this document.

Against a targeted client, a split-view log defeats even removal-after-entry detection. As written, line 84 grades an inclusion proof from an independent log as unconditionally "independently re-checkable," and line 108 says fetching one "carries no such defect" — both read as zero-trust when the guarantee is conditional on witnessing or gossip.

**This also fixes a misplaced quotation mark on line 75** — the opening quote currently sits before "of the log" instead of before "any", so the line renders as broken text.

```suggestion
- Treat a transparency log as detection of removal after entry. It says nothing about an entry that was never made.
- [RFC 9162](https://datatracker.ietf.org/doc/html/rfc9162) describes an append-only log built on Merkle Trees, which can prove that "any particular instance of the log is a superset of any particular previous instance".
- A log is itself a party. RFC 9162 notes that its auditing mechanisms "can be circumvented by a misbehaving log that shows different, inconsistent views of itself to different clients", so "it is necessary to treat each log as a trusted third party". Credit an inclusion proof only where the log's tree heads are witnessed or gossiped outside the log operator, and apply question 1 to that operator.
- [Sigstore's Rekor](https://docs.sigstore.dev/logging/overview/) is one deployed instance of such a log.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Get the behavior on the unavailable path in writing. Continue-and-drop, buffer-and-retry, and halt are three different behaviors under load.
+- Read the overflow behavior, not the happy path. A bounded buffer that overwrites oldest-first under pressure is continue-and-drop at the moment it matters.
+- Look for delivery failures recorded where the component has no reach. [ASVS 16.4.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md) asks that logs are "securely transmitted to a logically separate system", which is where a gap marker has to land.
+- [ASVS 16.5.3](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md) separately asks that an application "fails gracefully and securely, including when an exception occurs, preventing fail-open conditions". That is the component's own behavior.
+- Interrupt the recording path in your own environment and watch whether execution continues.
+- Where the component is hosted, ask for gap markers from a real outage.
+- An operator who has never recorded a gap is telling you the marker is not emitted.
```
mackowski: **Stated as a certainty when it isn't.** A young or low-volume deployment may simply not have had an outage yet, so a reviewer acting on this could mis-grade an honest supplier. The conditional says the useful part without the overreach. (It also fixes "is telling you", which presents your inference as the operator's speech.)

```suggestion
- Treat "we have never recorded a gap" as unverified until they can show you one, from a real outage or from a test they run for you.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+**Outcome.** A gap marker landing in a system the component does not control, or an interruption you performed yourself, is independently re-checkable. Documented fail-closed behavior you have not seen exercised is a supplier assertion. Treat "no findings" from a recorder that may have been fail-open as no information about that execution. Completeness is what this question tests; the accuracy of what was written is question 1.
```
mackowski: **State what fail-closed costs.** The question itself is neutral between continue-and-drop, buffer-and-retry, and halt, which is right — but a reader optimizing only for evidence quality will read "halt" as strictly better. AGENTS.md rule 2 asks that a control's limits be stated explicitly, and your own sibling sheet treats recorder unavailability as a denial-of-service lever (Logging Cheat Sheet, "Attacks on Logs → Availability"). ASVS 16.5.3, which you already cite, asks for failing "gracefully **and** securely."

```suggestion
**Outcome:** A gap marker landing in a system the component does not control, or an interruption you performed yourself, is independently re-checkable. Documented fail-closed behavior you have not seen exercised is a supplier assertion. Treat "no findings" from a recorder that may have been fail-open as no information about that execution. Fail-closed is not automatically the right choice: halting on an unreachable recorder turns a recording outage into an availability incident and hands an attacker a cheap denial-of-service lever, so the right point on that scale depends on how sensitive the action is. What matters for review is that the behavior is documented, exercised, and marks the gap. Completeness is what this question tests; the accuracy of what was written is question 1.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+### 6. Can the record be verified by someone who cannot also produce one?
+
+If the key material needed to check a record is the key material needed to create one, every party who can check can also forge, and the record cannot settle a disagreement between them. [RFC 2104](https://datatracker.ietf.org/doc/html/rfc2104) describes message authentication codes as "used between two parties that share a secret key in order to validate information transmitted between these parties". A shared-secret receipt is a reasonable control between a client and a server that already trust each other, and it carries nothing to the third party a reviewer is.
```
mackowski: "the third party a reviewer is" is a contorted noun phrase, and the contrast is joined with "and" where it needs "but". The argument itself is correct and, unusually, doesn't overreach — it scopes shared-secret receipts properly rather than dismissing them.

```suggestion
If the key material needed to check a record is the key material needed to create one, every party who can check can also forge, and the record cannot settle a disagreement between them. [RFC 2104](https://datatracker.ietf.org/doc/html/rfc2104) describes message authentication codes as "used between two parties that share a secret key in order to validate information transmitted between these parties". A shared-secret receipt is a reasonable control between a client and a server that already trust each other, but it proves nothing to a reviewer, who is a third party.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Check what a reviewer must hold to run the check.
+- A public key permits checking without conferring the ability to produce; a shared secret does not.
+- Count whose keys these are. Two keys held by one party are one party.
+- Check what the check depends on while it runs. Asking the producer's service whether a record is valid returns the claim under review.
+- Fetching public trust material, such as a root or an inclusion proof from an independent log, is a different dependency and carries no such defect.
+- Look for the format and signature suite stated in the record, so the check runs against a general-purpose implementation.
+- Pin your own roots of trust, and treat a change to them as a decision. A key delivered alongside the record it authenticates adds nothing.
+- [SLSA's verification guidance](https://slsa.dev/spec/v1.2/verifying-artifacts) shows the shape: configure "the verifier's roots of trust" as a map from recognized producer identities to what will be believed from each.
```
mackowski: Three changes in this block:

1. **Algorithm confusion (line 109).** This is the one bullet that tells a developer how to actually run the check, and as written it takes the suite from the artifact under review — the `alg:none` / HS256-signed-with-the-RSA-public-key pattern. RFC 8725 §3.1 is the standard countermeasure, and our own [JSON Web Token Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_Cheat_Sheet.html) covers the same failure.
2. **Key status at signing time (new bullet).** As written, a record signed by a since-rotated or compromised supplier key still grades "independently re-checkable". RFC 3161 §1 gives this as the core purpose of timestamping — evidence "that a digital signature was applied to a message before the corresponding certificate was revoked" — which ties this question back to question 2 nicely.
3. **"producer" → "builder" (line 111).** SLSA defines Producer and Builder as distinct roles, and the sentence you're paraphrasing says "recognized builder identities".

Line 108 also picks up the witnessing caveat from the transparency-log comment above.

```suggestion
- Fetching public trust material, such as a root or an inclusion proof from an independent log whose tree heads are witnessed outside its operator, is a different dependency and carries no such defect.
- Look for the format and signature suite stated in the record, so the check runs against a general-purpose implementation.
- Check the record against an algorithm set your verification policy fixes in advance. Never let the record's own header choose the algorithm or key type, and bind each pinned key to exactly one algorithm: [RFC 8725 Section 3.1](https://datatracker.ietf.org/doc/html/rfc8725#section-3.1) requires that libraries "MUST enable the caller to specify a supported set of algorithms and MUST NOT use any other algorithms when performing cryptographic operations".
- Pin your own roots of trust, and treat a change to them as a decision. A key delivered alongside the record it authenticates adds nothing.
- Check the signing key's status and validity window as of the time the record was signed, not as of today. Where a key may later be rotated or compromised, an independent timestamp (question 2) is what keeps an old signature meaningful.
- [SLSA's verification guidance](https://slsa.dev/spec/v1.2/verifying-artifacts) shows the shape: configure "the verifier's roots of trust" as a map from recognized builder identities to what will be believed from each.
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+
+## What These Questions Do Not Cover
+
+Three things sit outside the six questions, and a reviewer is better off knowing they are outside than assuming they were covered.
+
+- The live channel between agent and server. Message signing, nonces, tool-definition pinning, and mutual signing are in [Section 7 of the MCP Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/MCP_Security_Cheat_Sheet.html#7-message-level-integrity-and-replay-protection).
+- Records your own systems produce. Protecting logs at rest and in transit is in the [Logging Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html).
+- Runs spanning several parties. The interesting failure is at the seams: one agent's record shows a delegation and the counterparty has no record of receiving it, while both answer all six questions well.
```
mackowski: **Worth linking the closest neighbour we already have.** The AML/Sanctions sheet's Section 9 is the producer-side, payments-scoped counterpart to this framework — it asks that a receipt travel with the transaction "so every downstream party can independently verify [...] without trusting an upstream agent's word". Pointing at it keeps the boundary explicit and stops the two drifting apart. ("interesting failure" also reads a little off-tone for a control recommendation.)

```suggestion
- Runs spanning several parties. The failure to watch for is at the seams: one agent's record shows a delegation and the counterparty has no record of receiving it, while both answer all six questions well. For the payments-scoped version of this problem, see [Cross-Agent Payment Accountability](https://cheatsheetseries.owasp.org/cheatsheets/AML_Sanctions_AI_Agent_Payments_Cheat_Sheet.html#section-9-cross-agent-payment-accountability).
```

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,136 @@
...
+- [Logging Cheat Sheet](Logging_Cheat_Sheet.md)
+- [OWASP MCP Top 10](https://owasp.org/www-project-mcp-top-10/)
+- [OWASP ASVS 5.0, V16 Security Logging and Error Handling](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md)
+- [RFC 2104: HMAC: Keyed-Hashing for Message Authentication](https://datatracker.ietf.org/doc/html/rfc2104)
+- [RFC 3161: Internet X.509 Public Key Infrastructure Time-Stamp Protocol (TSP)](https://datatracker.ietf.org/doc/html/rfc3161)
+- [RFC 9162: Certificate Transparency Version 2.0](https://datatracker.ietf.org/doc/html/rfc9162)
+- [RFC 9334: Remote ATtestation procedureS (RATS) Architecture](https://datatracker.ietf.org/doc/html/rfc9334)
+- [RFC 9942: CBOR Object Signing and Encryption (COSE) Receipts](https://datatracker.ietf.org/doc/html/rfc9942)
+- [in-toto Attestation Framework: Statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md)
+- [SLSA v1.2: Verifying artifacts](https://slsa.dev/spec/v1.2/verifying-artifacts)
```
mackowski, line 156: Adds the two sources introduced by the suggestions above, keeping the RFCs in numeric order.

```suggestion
- [OWASP ASVS 5.0, V16 Security Logging and Error Handling](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md)
- [RFC 2104: HMAC: Keyed-Hashing for Message Authentication](https://datatracker.ietf.org/doc/html/rfc2104)
- [RFC 3161: Internet X.509 Public Key Infrastructure Time-Stamp Protocol (TSP)](https://datatracker.ietf.org/doc/html/rfc3161)
- [RFC 8725: JSON Web Token Best Current Practices](https://datatracker.ietf.org/doc/html/rfc8725)
- [RFC 9162: Certificate Transparency Version 2.0](https://datatracker.ietf.org/doc/html/rfc9162)
- [RFC 9334: Remote ATtestation procedureS (RATS) Architecture](https://datatracker.ietf.org/doc/html/rfc9334)
- [RFC 9942: CBOR Object Signing and Encryption (COSE) Receipts](https://datatracker.ietf.org/doc/html/rfc9942)
- [in-toto Attestation Framework: Statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md)
- [in-toto Attestation Framework: DigestSet](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md)
- [SLSA v1.2: Verifying artifacts](https://slsa.dev/spec/v1.2/verifying-artifacts)
```

mackowski (changes requested):
Thanks for this, @astrogilda — and thanks for the unusually careful sourcing. I checked every quoted string against its primary source and **all eight match character for character**. All three ASVS requirement numbers map to the right requirement text (16.4.2, 16.4.3, 16.5.3 — mis-numbered ASVS requirements are the classic fabrication tell, and it's absent here). RFC 9942 is indeed COSE Receipts. And your call on the retired SLSA page checks out: v1.0 carries a Retired banner, v1.2 carries `Status: Approved`, and the "verifier's roots of trust" sentence is byte-identical across both.

Your Index.md reasoning is also correct, and I verified it rather than taking your word for it: `Update_CheatSheets_Index.py` rewrites the file wholesale from `os.scandir`, `mkdocs.yml` has no `nav:` key, and the committed index already lags the directory by 24 entries. **Please don't hand-edit it.** Placement in `cheatsheets/` is right too — `cheatsheets_draft/` holds one specific WIP family, not an intake queue, so no need to move.

I also confirmed the gap you and I discussed in #2313 is real and still open: `MCP_Security_Cheat_Sheet.md:121` is "Log all MCP tool invocations with full parameters, user context, and timestamps" and `AI_Agent_Security_Cheat_Sheet.md:352` is "Log all agent decisions, tool calls, and outcomes". Neither asks who produced the record. Nothing in the series covers fail-open recording paths as an evidentiary defect. A standalone sheet is the right call.

I've left suggestions inline. **Five of them I'd consider blocking:**

1. **Line 74–76 — a transparency log is itself a party.** RFC 9162 §1, in the paragraph right after the sentence you quote, says its auditing mechanisms "can be circumvented by a misbehaving log that shows different, inconsistent views of itself to different clients" and that "it is necessary to treat each log as a trusted third party". Your own method is "count the parties who must cooperate", so the log operator needs counting. (This suggestion also fixes a misplaced quotation mark on line 75 that currently renders as broken text.)
2. **Lines 48–49, 53 — a nonce dates the signature, not the capture.** The direction of your bound is right, but RFC 9334 §10.2 calls the epoch "rough": "the time between the creation of Claims and the collection of Claims is indistinguishable". That's precisely the threat question 2 names.
3. **Line 109 — algorithm confusion.** As written, the reader takes the suite from the record under review, which is the `alg:none` / HS256-with-the-RSA-public-key pattern. RFC 8725 §3.1 is the fix.
4. **Line 9 — the scoring rule collapses to "Nothing".** "No higher than its weakest answer" plus line 113 grades an ordinary vendor JSON log at the bottom outcome, contradicting your own line 14. Scoping it to disputes with the supplier keeps the argument and loses the side effect.
5. **Lines 36–38 and passim — the leverage gap.** Five bullets assume procurement power a team adopting a community MCP server doesn't have, and the sheet never says so. The gateway/egress fallback is the move that always works; it deserves naming rather than two oblique mentions.

**Two things that would raise this a lot and aren't in the suggestions**, because they're additions rather than edits:

- **A worked example.** The intro promises the six questions apply identically to a signed JSON log, an in-toto Statement, a COSE receipt, and a vendor export — but COSE appears only there, and in-toto only once. A compact table scoring four record types (hosted-agent OpenTelemetry spans, self-hosted MCP server stdout, in-toto attestation, transparency-log-backed receipt) across Q1–Q6 would be about six lines of Markdown and would make the whole sheet immediately usable.
- **Back-links.** Nothing currently links *to* this sheet, so the hub-and-spoke only runs one way. One line each in MCP Security §10 and AI Agent Security §6 ("when the logs come from a component you did not build, see …") would do it — that's 2 lines across 2 files, still inside the 3-sheet scope limit, so please fold it into this PR.

Minor, not worth separate suggestions: `**Outcome.**` → `**Outcome:**` at lines 67, 84 and 113 as well (colon-style is the series convention — `**Example:**` appears 57 times, the period style nowhere); and the inline cross-links use absolute `.html#anchor` URLs while References uses relative `.md` paths. Both conventions exist in the repo, relative `.md` is dominant (271 vs 68) and is the one CI actually checks — worth considering, though I verified all five of your anchors resolve live, so this is style rather than correctness.

On the blockquote question you raised in the description: your call was right. Short in-sentence fragments are handled exactly that way elsewhere in the series, and lifting them into blockquotes would break the sentences.

And please do pass on the credit to @Rul1an — the point about conformance corpora inheriting the same defect one level up is a good one, and a follow-up sheet with them co-authoring sounds like the right home for it. Same for the multi-party delegation case you scoped out; that's a real second sheet.

Nice work. This is a well-argued piece and I'd like to merge it once the items above are addressed.

jmanico (dismissed):

astrogilda:
All seventeen suggestions are applied, in eight commits, one per theme. One item, the absolute versus relative link style, I answer below rather than apply. Three changes go beyond or around your suggestion text, and I flag each. Seventeen suggestions with five blocking is more work than most reviews put in. The transparency-log point found a real hole in the argument rather than a wording problem. Thank you.

On the transparency log you are right, and the correction runs deeper than the two lines you marked. RFC 9162 Section 1 puts my sentence and your sentence in one paragraph, and I read the first and stopped. The sheet counts the parties who must cooperate, so a log that can show inconsistent views to different clients is one. Question 4 now quotes both fragments in a bullet. It credits an inclusion proof only where the log's tree heads are witnessed or gossiped outside the log operator. Question 1 then applies to that operator.

I carried the same condition into two places you named in the body but did not mark. The question 4 outcome graded an independent log unconditionally. The question 6 bullet said an inclusion proof "carries no such defect". Both now say "an independent log whose tree heads are witnessed outside its operator". Without them the sheet would still read as zero trust exactly as you described. Your suggestion also repaired the misplaced opening quotation mark, which rendered as broken text.

On the nonce, applied, with one deviation. RFC 9334 Section 10.2 writes it as a bullet beginning "The time between the creation of Claims and the collection of Claims is indistinguishable", with a capital T. Your text lowercases it to sit mid-sentence, which reads better but breaks the character-for-character check you ran on the other eight. The bullet therefore uses the bracketed form from your review body: "[t]he time between the creation of Claims and the collection of Claims is indistinguishable". Everything else there is verbatim, including the plainer wording for contemporaneity. The question 2 outcome takes your text unchanged, so a nonce bounds the signing time and does not establish capture as the work ran.

On algorithm confusion, applied verbatim. That bullet was the one place telling a developer how to run the check, and it handed the algorithm choice to the artifact under review. It now fixes the algorithm set in the verification policy in advance and forbids the record's header from choosing. Each pinned key binds to exactly one algorithm, quoting RFC 8725 Section 3.1. RFC 8725 is in References in numeric order, per your last suggestion.

I took both other changes in that block. The key status bullet closes a genuine gap I had not seen: a record signed by a since-rotated supplier key was scoring independently re-checkable. Tying it back to question 2 through RFC 3161's own stated purpose beats the fix I would have written. The producer to builder correction is simply right. The SLSA v1.2 page says "the recognized builder identities", the word producer is absent from that sentence, and SLSA treats Producer and Builder as separate roles.

On the scoring rule, applied verbatim, and the most useful catch of the five for the sheet's readers. Weakest answer bounds everything, an unsigned export scores the bottom outcome, and that outcome's column reads Nothing. Together they tell a developer their existing telemetry is worthless. That is not what I meant, and it contradicts the row above. Scoping the bound to what the record settles in a dispute keeps the argument and loses the side effect. The table header and the two cells are yours as written, including the switch from and to but.

I fixed that switch in the two other places you flagged. On the reader with no bargaining power, applied verbatim, plus one more bullet you named in the body. The corroboration bullet now says outright: run the component behind a gateway, proxy, or egress point you control. That is the move that remains when you can demand nothing at all. The contract bullet admits what it costs. The sizing bullet names the decision classes rather than dead-ending.

Beyond your suggestion text I softened the unavailable-path bullet in question 5. It told the reader to get the behavior in writing and offered nothing to a reader who cannot ask. It now offers the documentation and the source as the route. The sandbox tenant bullet in question 4 already carried its own fallback, so I left it.

The rest are applied as written. MCP and COSE are expanded on first use, matching every other sheet, and I should have done it. The verification definitions are now three named terms, with the parallel restored and the last clause split off. Meta-commentary is out of the question 1 outcome. The six Outcome lead-ins are all colon style, including the three at 67, 84 and 113 you listed under minor. The HMAC sentence reads better as you wrote it.

The digest strength floor is a real hole in an adversarial-supplier model. The framework I hold up as the worked example does permit md5, sha1 and Git SHA-1 object IDs. Quoting the DigestSet consumer rule surfaces what the spec already says. The fail-closed paragraph is yours unchanged. The question was neutral between the three behaviors, but nothing stopped a reader treating halt as strictly better, and naming the denial-of-service lever fixes that. Treating "we have never recorded a gap" as unverified rather than as proof of a missing marker is more accurate and does not mis-grade a young deployment.

Cross-Agent Payment Accountability is the right neighbour for the multi-party exclusion, so that link is in. The worked table is in as a new section, scoring four sources across all six questions. Those are hosted-agent OpenTelemetry spans, self-hosted MCP server stdout, a supplier-signed in-toto Statement, and a receipt anchored in a witnessed transparency log. I scored it honestly rather than flatteringly, so the vendor export row takes the bottom outcome on four of six. The strongest row still scores no information on question 5.

The reading paragraph says it plainly: no row is independently re-checkable on all six, and none reaches that outcome on question 4 or 5. Neither of those asks what the bytes say. I would not have found that shape without your asking for the table.

The back-links are folded in, one line each in MCP Security Section 10 and AI Agent Security Section 6. Both use the relative .md form that dominates the repository and that the link checker validates. That takes the pull request to three files and 158 net additions, inside the three-sheet and 1500-line limits. Index.md is untouched, per your instruction and my own reasoning about the generator.

The item I answer rather than apply is the link style for five inline cross-links in the new sheet. Those five carry section anchors, and the two renderers slugify them differently. I rendered the MCP heading through the GitHub markdown API. It produces "10-monitoring-logging--auditing" with a double hyphen, because the ampersand drops out and leaves both spaces. The published site produces "10-monitoring-logging-auditing" with a single one, and I confirmed that id in the live page source. A relative .md link keeps whichever fragment I write, so no form is correct in both.

The absolute form is correct on the published site, where these anchors were verified and where a reader following a cross-link almost always stands. In References, where no anchor is needed, the sheet already uses the relative .md form you prefer. The convention is now explicit: relative for whole sheets, absolute for a deep link into a section. If you would rather have consistency than working anchors, I will convert all five and drop the fragments, in one commit.

Two small things I did not do, because your suggestion text did not cover them and I did not want to widen the diff. You mentioned the JSON Web Token Cheat Sheet in the algorithm confusion comment as covering the same failure. The suggestion cites only RFC 8725, so I cited only RFC 8725, and I will add the sheet link on a word from you. The AML and Sanctions sheet is linked inline from the exclusions section but not in References, because your References suggestion did not list it. I will add it if you would rather the two lists match.

Thank you for running the sourcing check rather than taking my word for it. Every string I added this round went through the same check against a local copy of the primary document. Every quotation in the sheet now verifies character for character against the sources it cites. Those are RFC 2104, 3161, 8725, 9162 and 9334, ASVS V16, MCP08:2025, the in-toto Statement and DigestSet specs, and the SLSA v1.2 page. A deliberate negative control in the same run proves the check was reading the sources rather than passing everything. All twenty-three external links answer 200 today, including the two new ones, and both new anchors resolve in the live page source.

Markdown lint and terminology lint are clean on the whole repository. I ran the citation density and scope checks locally against your workflow files too. Credit to @Rul1an on the conformance corpora point, which deserves its own sheet rather than a paragraph here. The multi-party delegation case is the other one, and I would be glad to co-author both.

astrogilda:
@mackowski I've applied all 17 suggestions and the branch is current with master, so would you mind taking another look?

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+### 1. Could the component the record describes have written or altered it?
+
+A component that writes its own execution log can omit the call that mattered, rewrite an argument, or emit a record for a run that did not happen, and the document still parses with every field complete. What settles the question is not where the recorder sits in a diagram, but how many parties would have to cooperate to produce this record for a run that did not happen.
+
+- List the parties between the event and the bytes you hold: the component, its host, the recorder's operator, the keyholder.
+- Count parties, not processes. A recorder the supplier operates is one party however many hosts it spans.
+- Look for an input the component did not choose. A recorder fed only by the component records the component's report.
+- Check whether the signing service will sign arbitrary content on the component's request.
+- If it will, that signature is worth what a readable key is worth.
+- Look for a copy landing where the component cannot change it. [ASVS 16.4.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md) asks that logs "cannot be modified".
```
mackowski, line 32: **A transparency log does not raise the party count.** Question 1's method is counting parties who must cooperate, but the bullet list never mentions anchoring — and the table at line 129 then credits a log-anchored receipt with independently re-checkable *origin*. A log appends what it is given rather than appraising it, so the supplier alone can log a receipt for a run that never happened. This is also what your own line 76 says ("It says nothing about an entry that was never made") and line 34 ("needs an input the component did not choose **and** a party with no stake in the answer" — a witnessed log supplies only the second). Adding the point here is what makes the cell change at line 129 read as consistent rather than arbitrary.

```suggestion
- Look for a copy landing where the component cannot change it. [ASVS 16.4.2](https://github.com/OWASP/ASVS/blob/master/5.0/en/0x25-V16-Security-Logging-and-Error-Handling.md) asks that logs "cannot be modified".
- Anchoring the record in a transparency log does not raise this count. A log appends what it is given rather than appraising it, and [RFC 6962](https://datatracker.ietf.org/doc/html/rfc6962#section-7.2) is explicit on the point: "The logs do not themselves detect misissued certificates". Anchoring buys tamper-evidence, append-only consistency, and public discoverability, not a second party who would have to cooperate.
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Outcome:** Most agent telemetry shipping today is a supplier assertion: SDK-emitted spans, a framework's own traces, an MCP server's logs, and a hosted provider's export are all fed by the component and handled downstream by a single party. Independently re-checkable needs an input the component did not choose and a party with no stake in the answer. A supplier who will not describe the path scores no information. Where the answer is supplier assertion and you still have to proceed:
+
+- Corroborate against a record your own side produces, such as a gateway or egress log covering the same calls. Where you have no leverage over the supplier at all, this is the move that always remains: run the component behind a gateway, proxy, or egress point you control, and record there.
```
mackowski: **The gateway fallback needs its does-not-cover clause.** This is the only bullet in the list without one — lines 37 and 38 both carry their limits — and it is the bullet aimed at readers with no leverage, which is the worst place to leave a limit unstated. A boundary recorder sees only boundary-crossing traffic: nothing for a local stdio MCP server (which the spec makes a first-class transport that clients should support whenever possible, launched as a local subprocess), and destinations rather than bodies through a CONNECT-mode proxy — so it cannot corroborate the "rewrite an argument" threat you name at line 25. Your own table row 2 is that deployment. I have also softened "always remains", since for a stdio server it does not.

```suggestion
- Corroborate against a record your own side produces, such as a gateway or egress log covering the same calls. Where you have no leverage over the supplier at all, this is usually the move that remains: run the component behind a gateway, proxy, or egress point you control, and record there. A boundary recorder covers only what crosses that boundary, so it sees neither a local stdio MCP server's file writes and in-process tool calls, nor request payloads where you do not terminate TLS.
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+
+- Check whether the record states its window. A window you inferred from the entries present is not a stated one.
+- Treat a transparency log as detection of removal after entry. It says nothing about an entry that was never made.
+- [RFC 9162](https://datatracker.ietf.org/doc/html/rfc9162) describes an append-only log built on Merkle Trees, which can prove that "any particular instance of the log is a superset of any particular previous instance".
+- A log is itself a party. RFC 9162 notes that its auditing mechanisms "can be circumvented by a misbehaving log that shows different, inconsistent views of itself to different clients", so "it is necessary to treat each log as a trusted third party". Credit an inclusion proof only where the log's tree heads are witnessed or gossiped outside the log operator, and apply question 1 to that operator.
+- [Sigstore's Rekor](https://docs.sigstore.dev/logging/overview/) is one deployed instance of such a log.
```
mackowski: **Rekor will be read as satisfying the condition set one line above.** The sentence is not false — "such a log" has a valid antecedent in line 77 — but placed immediately after the requirement that tree heads be "witnessed or gossiped outside the log operator", and with table row 4 reading "transparency log witnessed outside its operator", a reader will score Rekor-anchored receipts as row 4. As of today `sigstore/rekor-tiles` `CLIENTS.md` still reads "In the initial launch of Rekor v2, checkpoints will not be witnessed, while we wait for the launch of a public witness network", and the v2 GA post says witnessing "will be implemented soon". Phrasing it as a policy check rather than a status claim also keeps the sentence correct once the witness network does launch.

```suggestion
- [Sigstore's Rekor](https://docs.sigstore.dev/logging/overview/) is one deployed instance of such a log. Check the witness policy of the specific log you rely on before crediting an inclusion proof, since a log can be append-only and still publish checkpoints nobody outside its operator co-signs.
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+**Outcome:** A signature checkable against a root you pinned, with a stated suite and no call to the producer, is independently re-checkable, and what it establishes is who asserted the record and that the bytes are unchanged. A shared-secret receipt, or a check that queries the producer's service, is a supplier assertion. An unsigned export gives no information beyond what the producer chose to send.
+
+## Four Record Types Scored
+
+The six questions are indifferent to format, so running them over four common record types shows what each one actually buys. Scores are for the record as it typically arrives, before the reviewer does any of the work the questions name.
+
+| Record | Q1 origin | Q2 timing | Q3 subject | Q4 coverage | Q5 failure | Q6 check |
+|---|---|---|---|---|---|---|
+| Hosted agent's [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/traces/) spans, exported by the vendor | Assertion | Assertion | Assertion | None | None | None |
+| Self-hosted MCP server stdout you capture yourself | Assertion | Re-checkable | None | Assertion | None | None |
```
mackowski: **Row 2, Q2 timing: `Re-checkable` → `Assertion`.** Capturing your own stdout live is *witnessing*, which line 17 separates from the independent re-checking you say the six questions test. It also fails your own definition at line 13 — "a party who cannot produce the record can confirm it" — since you produced it. Neither route Q2's outcome allows at line 54 (a countersignature from an authority the supplier does not operate, or a nonce you supplied) is present. This is the only cell in the table crediting reviewer self-help as re-checkable, and it is inconsistent with how the same row is scored on Q4 and Q5, where it gets `Assertion`/`None` under the line 122 convention even though lines 87 and 101 say self-help reaches the top outcome there. I have added the footnote this needs to the paragraph at line 131 so the downgrade does not lose the real distinction between this row and row 1.

```suggestion
| Self-hosted MCP server stdout you capture yourself | Assertion | Assertion | None | Assertion | None | None |
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+| [in-toto Statement](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) signed by the supplier | Assertion | Assertion | Re-checkable | None | None | Re-checkable |
+| Receipt anchored in a transparency log witnessed outside its operator | Re-checkable | Re-checkable | Re-checkable | Assertion | None | Re-checkable |
```
mackowski: **Row 4, Q1 origin: `Re-checkable` → `Assertion`.** This is the transparency-log point from the last round surviving in the table. Anchoring constrains alteration and suppression *after* entry, not fabrication *at* entry, and question 1 asks whether the component could have "written **or** altered" the record — anchoring answers only "altered". The party count for fabrication stays at one, so the cell contradicts lines 29, 76 and 34. Q2 stays `Re-checkable` here, correctly: co-signed checkpoints are timestamped and can serve as an independent signed timestamp.

```suggestion
| Receipt anchored in a transparency log witnessed outside its operator | Assertion | Re-checkable | Re-checkable | Assertion | None | Re-checkable |
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

cheatsheets/Verifying_Third_Party_Agent_Execution_Evidence_Cheat_Sheet.md
```diff
@@ -0,0 +1,156 @@
...
+
+Re-checkable, assertion and none abbreviate the three outcomes above. Read the table across rather than down. No row is independently re-checkable on all six, and no row reaches that outcome on question 4 or question 5 at all, because neither asks what the bytes say; both ask what the recorder was positioned to see and whether it could decline to write. The strongest row still scores no information on question 5. The self-hosted row is the one a reader can move without anyone's permission, since controlling the host makes injecting an event and interrupting the recording path both available.
```
mackowski: **The because-clause teaches a false general rule.** "because neither asks what the bytes say" is contradicted by your own outcomes: line 87 says coverage "is independently re-checkable when something outside the producer saw the sequence" and line 101 says "a gap marker landing in a system the component does not control, or an interruption you performed yourself, is independently re-checkable". As written this tells the reader Q4 and Q5 can never reach the top outcome, which would talk them out of injecting an event and interrupting the recording path — the two moves this very paragraph calls available without anyone's permission. The cells are right, but for the line 122 reason, not the stated one. I have also folded in the row 2 footnote from the Q2 change, and replaced "across rather than down" with something actionable.

```suggestion
Re-checkable, assertion and none abbreviate the three outcomes above. Read each row as a whole: what matters is the mix of outcomes for one record type, not how the types compare on a single question. No row is independently re-checkable on all six, and no row reaches that outcome on question 4 or question 5. That is not because those two questions can never reach it, since both outcomes above name routes that do, but because every route runs through work the reviewer has not done at the point these scores apply. The strongest row still scores no information on question 5. Row 2's assertions are your own rather than the supplier's, which helps your own investigation and carries no weight in a dispute. The self-hosted row is the one a reader can move without anyone's permission, since controlling the host makes injecting an event and interrupting the recording path both available.
```

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

mackowski (changes requested):
Thanks @astrogilda — re-reviewed after the 17 applied suggestions.

**Sourcing is verified a second time and it holds up.** All 17 quoted fragments match their primaries verbatim, ASVS 16.4.2/16.4.3/16.5.3 each map to exactly the text you quote, and your SLSA catch checks out by contrast (v1.2 reads `Status: Approved`, the old v1.0 URL reads `Status: Retired`). Prior blockers 2, 3 and 4 are cleanly resolved — I verified RFC 9334 §10.2 and RFC 8725 §3.1 against the RFC text, and §3.1's "each key MUST be used with exactly one algorithm" grounds your pinning rule properly.

**A correction I owe you on `Index.md`.** Last round I asked you not to hand-edit it. #2324 has since added a required Index Drift Check, so the generated index now has to be committed — your current version is correct. I simulated the workflow on `master` merged with your head and it passes byte-for-byte. Scope (3 sheets, 158 net), citation density (all 6 new H2s), markdownlint and textlint all pass locally too. The real runs are still held at `action_required` behind the first-contribution gate; I will release them.

**Five items still block merge, and they share one root cause: the prose is right and the table overclaims.** Every one of them is the scoring table or the concluding paragraph disagreeing with the body of the sheet.

1. Row 4's Q1 `Re-checkable` — the transparency-log point from last round, surviving in the table. Anchoring constrains alteration after entry, not fabrication at entry.
2. Row 2's Q2 `Re-checkable` — self-captured stdout is witnessing, and fails your own line 13 definition.
3. Line 79 — Rekor reads as meeting the witnessing bar set one line above; its checkpoints are not witnessed today.
4. Line 36 — the gateway fallback is the one bullet with no does-not-cover clause, and it is the bullet for readers with no leverage.
5. Line 131 — the because-clause contradicts your Q4 and Q5 outcomes and teaches a false general rule.

Suggestions are inline and apply cleanly against your current head. **Comments on lines 127, 129 and 131 are mutually dependent** — please take all three together or none, since the two cell changes without the paragraph rewrite would leave the text describing a table that no longer exists.

Two things I explicitly checked and am *not* asking you to change: the Q5 fail-open passage is well balanced and right to refuse a blanket fail-closed recommendation, and 22 of the 24 table cells are sound, including row 3's Q6.

The gap is real — I re-searched the corpus and nothing treats a self-reported record as an evidentiary question. Standalone sheet is the right call, the bidirectional back-links are the right shape, and placement in `cheatsheets/` is correct. Get the table honest and this merges.

*This review was done using LLM (Opus 5), using `.claude/commands/review-cheatsheet-pr.md`. Check for correctness*

astrogilda:
All five are in, with the three dependent edits as one change. Two carry my wording: the anchoring bullet cites RFC 9162, the RFC the sheet already uses, which repeats the 6962 sentence. The gateway bullet keeps both of your limits and adds their cost, that the recorder cannot tell you whether an argument was rewritten. The Rekor line is yours as written; rekor-tiles CLIENTS.md still says checkpoints are not witnessed, and the index, markdownlint and textlint are clean locally.

KadirArslan:
I think the cheat sheet should include a short and actionable Do / Don't section

The current content explains the evidence model well, but for the readers it would be good to have it

astrogilda:
@KadirArslan Thanks for asking for it. The section is on the branch, as the last section before References. Seven lines each way, and each one is an action a reviewer takes on a record somebody else produced. Your spelling was `Do / Don't`; I used `Do's and Don'ts`, the name six other sheets already give that section.

bIackr0se:
Q4 line 83 overstates what an independent log’s tree size can show. RFC 9162 §4.9 defines `tree_size` as the number of entries in the entire Merkle tree; witnessing it establishes append-only history, not that this execution contributed one leaf per expected event. Unrelated leaves can increase the size while an event from this run is missing. Could you require an observer-controlled per-execution sequence or reconciliation against an expected-event list before calling omission detection independently re-checkable? The later force/inject bullets describe the needed binding.

Rul1an:
I had planned to wait until this lands, but there is one narrow follow-up to the Q4 point above: could the Outcome paragraph carry the same limit?

Finding an injected event in the returned record demonstrates that the tested event reached that record. It does not establish that all expected events in the execution did. Witnessed log checkpoints likewise do not, by themselves, establish that execution's coverage.

I would distinguish a successful spot-check from a completeness claim. For the latter, the reviewer needs a stated execution/window, a basis for expecting events that the producer cannot silently redefine, and a check of the returned record against that expectation. The claim then stays within what the observer was positioned to see.

A small wording change in the Outcome would be enough; the existing force/inject bullets can stay as practical checks.

Prepared with Codex assistance.

astrogilda:
@bIackr0se, thank you, you're right: RFC 9162 Section 4.9 makes `tree_size` "the number of entries currently in the log's Merkle Tree", so it counts the whole log. In c3b94a5, Q4 asks for an observer-advanced sequence per execution, or expected events from your own boundary records, checked against the record.

@Rul1an, thanks too: the Outcome now calls an injected event that turns up a spot-check, and asks for your three parts before any completeness claim. Our conformance-corpora sheet starts there.

Issue #2365, opened 2026-08-13
New CS proposal: Passkey Security Cheat Sheet


basriakkaya:
## What is the proposed Cheat Sheet about?

The proposed **Passkey Security Cheat Sheet** would explain what passkeys are and give developers practical guidance for implementing passkey registration, authentication, credential management, and recovery securely on the relying-party side.

This proposal follows the direction given by maintainers in #2150 and #2168. That earlier contribution focused narrowly on credential-to-account binding and was closed because the project needs a cheat sheet covering the full passkey implementation lifecycle. This proposal addresses that requested scope rather than resubmitting the narrow document.

The repository currently contains short passkey descriptions in the Authentication and Multifactor Authentication cheat sheets, but it does not have a dedicated implementation guide for WebAuthn relying parties.

## What security issues are commonly encountered related to this area?

Passkeys remove shared password secrets and provide phishing resistance through relying-party scoping, but relying-party implementation errors can still weaken or bypass authentication. The cheat sheet would cover:

- Reused, predictable, expired, or insufficiently bound registration and authentication challenges.
- Failure to verify the expected origin, RP ID hash, ceremony type, credential ID, signature, and user-presence or user-verification flags.
- Treating user presence and user verification as equivalent, or assuming every passkey authentication is automatically multi-factor.
- Incorrect association of a newly registered credential with the authenticated account.
- Allowing a stolen or weak session to enroll, replace, or remove passkeys without reauthentication.
- Account recovery paths that are weaker than passkey authentication and become the practical account-takeover route.
- Username enumeration and privacy leakage through user handles, credential identifiers, attestation, and error responses.
- Incorrect assumptions about signature counters, synced credentials, backup eligibility, and backup state.
- Failure to support credential revocation, user-visible credential inventory, security notifications, and recovery from authenticator loss.
- Overly broad RP ID choices and unsafe cross-origin or embedded WebAuthn use.
- Implementing the WebAuthn protocol or cryptographic verification manually instead of using a maintained server-side library.

## What is the objective of the Cheat Sheet?

The objective is an architecture-first implementation reference that helps developers build and review an end-to-end passkey lifecycle without reproducing the WebAuthn specification.

The proposed outline is:

1. Passkey and WebAuthn terminology, including relying parties, clients, authenticators, discoverable credentials, and synced versus device-bound credentials.
2. Security properties and limitations, including phishing resistance, verifier-name binding, replay resistance, and the distinction between user presence and user verification.
[...]

PR #2369, opened 2026-08-14
Add Passkey Security Cheat Sheet

basriakkaya:
## Description

This PR adds a vendor-neutral Passkey Security Cheat Sheet for developers implementing the complete passkey relying-party lifecycle.

It follows the scope approved in #2365 and the latest maintainer direction for a cheat sheet that covers the entire passkey topic rather than only the WebAuthn JavaScript API.

## What this covers

- Passkey and WebAuthn security properties and limitations
- RP ID, origin, user presence, and user verification decisions
- Registration and authentication ceremony requirements
- Credential enrollment, revocation, and audit controls
- Account recovery and bootstrap protections
- Synced and device-bound credential considerations
- A deployment and review checklist

The guidance is architecture-first, language-independent, and uses no implementation-specific code. Technical claims are linked inline to W3C WebAuthn, NIST SP 800-63B-4, and related OWASP guidance.

## Validation

- [x] New Cheat Sheet template and writing guide followed
- [x] Change is limited to one new cheat sheet
- [x] Full Markdown lint passed
- [x] Full terminology lint passed
- [x] All internal and external links passed `markdown-link-check`
- [x] References use descriptive Markdown links
- [x] Registration and assertion requirements were checked against the W3C relying-party verification procedures

## Scope and sourcing

- [x] This PR is focused: it adds only `Passkey_Security_Cheat_Sheet.md`.
- [x] Security-critical recommendations are supported by primary W3C, NIST, or OWASP sources linked inline.
- [x] I reviewed the cited sources and verified that they support the associated claims.

## AI Tool Usage Disclosure

- [x] I have used AI tools to generate the contents of this PR. I have verified the contents and I affirm the results. The LLM used is OpenAI Codex (GPT-5). The prompt asked Codex to analyze the OWASP Cheat Sheet Series, identify genuine contribution candidates, and, after approval, prepare a maintainer-aligned Passkey Security Cheat Sheet covering the complete relying-party lifecycle. I independently reviewed every citation and technical claim against the cited W3C, NIST, and OWASP sources.

This PR fixes issue #2365.

### DIFF
--- a/cheatsheets/Passkey_Security_Cheat_Sheet.md
+++ b/cheatsheets/Passkey_Security_Cheat_Sheet.md
@@ -194,6 +194,8 @@ Choose policy according to the application's threat model:
 
 ## Deployment and Review Checklist
 
+Use this checklist together with the W3C [registration](https://www.w3.org/TR/webauthn-3/#sctn-registering-a-new-credential) and [authentication](https://www.w3.org/TR/webauthn-3/#sctn-verifying-assertion) verification procedures.
+
 ### Server and Protocol
 
 - Use a maintained WebAuthn server library and keep it updated.
--- a/cheatsheets/Passkey_Security_Cheat_Sheet.md
+++ b/cheatsheets/Passkey_Security_Cheat_Sheet.md
@@ -1,4 +1,4 @@
-# WebAuthn Security Cheat Sheet
+# Passkey Security Cheat Sheet
 
 ## Introduction
 
--- a/cheatsheets/Passkey_Security_Cheat_Sheet.md
+++ b/cheatsheets/Passkey_Security_Cheat_Sheet.md
@@ -1,0 +1,233 @@
+# Passkey Security Cheat Sheet
+
+## Introduction
+
+Passkeys are [WebAuthn public key credentials](https://www.w3.org/TR/webauthn-3/#sctn-use-cases) that let a user authenticate without sending a shared secret to the application. The authenticator keeps the private key and the relying party stores a public key. A passkey is scoped to a relying party identifier (RP ID), which provides phishing resistance when the ceremony and server-side verification are implemented correctly.
+
+This cheat sheet is for developers who build the relying-party side of passkey registration, authentication, credential management, and account recovery. The main recommendations are:
+
+- Use a maintained WebAuthn server library and validate every required ceremony field on the server.
+- Bind registration to the intended account and require recent authentication before changing passkeys.
+- Treat user presence (UP) and user verification (UV) as different security properties.
+- Design credential management and recovery with the same care as authentication.
+- Support more than one authenticator so that the loss of one device does not force a weak recovery path.
+
+## Understand the Security Model
+
+### Components and Terms
+
+- The **relying party (RP)** is the application that registers and authenticates users.
+- The **client** is the browser or other software that calls the WebAuthn API.
+- The **authenticator** creates credentials and signs authentication assertions. It can be built into a device or be a separate roaming authenticator.
+- A **passkey** is a discoverable WebAuthn credential. It may be device-bound or may be synced between devices by a credential provider.
+- The **credential ID** identifies a credential to the RP. The corresponding private key remains under authenticator control.
+- The **user handle** is an opaque RP-generated identifier for an account. It is not a username or email address.
+- **User presence (UP)** shows that a person interacted with the authenticator.
+- **User verification (UV)** shows that the authenticator locally verified the user, for example with a PIN or biometric.
+
+Do not treat UP and UV as equivalent. Require UV when the passkey is expected to satisfy a multi-factor or other high-assurance authentication policy. A passkey is not automatically multi-factor merely because an authenticator supports UV; the RP must [request UV and verify the returned UV flag](https://www.w3.org/TR/webauthn-3/#user-verification).
+
+### Security Properties and Limits
+
+[WebAuthn credentials are scoped to an RP](https://www.w3.org/TR/webauthn-3/#sctn-rp-benefits), and the signed response includes a hash of that RP ID. The client data also binds the response to the web origin and ceremony challenge. Correct verification therefore gives the RP phishing resistance, verifier-name binding, and replay resistance.
+
+These properties do not protect every part of the account lifecycle. Passkeys do not by themselves prevent:
+
+- A compromised authenticated session from adding an attacker-controlled passkey.
+- An insecure recovery process from bypassing passkey authentication.
+- Server-side credential records from being associated with the wrong account.
+- Application code, authorization, or session-management vulnerabilities.
+- Compromise of a device, authenticator, or account used to synchronize passkeys.
+
+Apply the controls in the [Authentication Cheat Sheet](Authentication_Cheat_Sheet.md), [Session Management Cheat Sheet](Session_Management_Cheat_Sheet.md), and [Transaction Authorization Cheat Sheet](Transaction_Authorization_Cheat_Sheet.md) to the surrounding application.
+
+### Choose the RP ID Carefully
+
+An RP ID determines where a credential can be used. Use the narrowest stable domain that covers the intended application. Do not use a registrable parent domain merely to share credentials with unrelated subdomains. The WebAuthn specification discusses the [security implications of the RP ID and origin relationship](https://www.w3.org/TR/webauthn-3/#sctn-validating-origin); compromise of an origin within the RP ID scope may affect the security of that RP deployment.
+
+The server must use the same expected RP ID during registration and authentication. Maintain an explicit allowlist of permitted origins, including scheme, host, and port where applicable. Do not construct an expected origin from an untrusted request header.
+
+[WebAuthn is restricted to secure contexts](https://www.w3.org/TR/webauthn-3/#sctn-api). Serve registration, authentication, and credential-management pages over HTTPS and protect them against script injection. Avoid cross-origin WebAuthn in embedded frames unless it is an intentional, reviewed design using the [required Permissions Policy and origin validation](https://www.w3.org/TR/webauthn-3/#sctn-iframe-guidance).
+
+### Use a Maintained Library
+
+Do not implement CBOR parsing, COSE key handling, attestation validation, or assertion verification yourself. Use a maintained server-side WebAuthn library that implements the W3C relying-party verification steps. Keep it updated and include successful and failing registration and authentication ceremonies in automated tests.
+
+The library does not make account workflows safe automatically. The application remains responsible for issuing and consuming challenges, choosing the expected RP ID and origins, binding ceremonies to the correct session and account, enforcing UV policy, and managing credential lifecycle events.
+
+## Secure Passkey Registration
+
+### Establish an Authorized Account Context
+
+Registration attaches a new authenticator to an account and must be treated as a sensitive account change.
+
+- For an existing account, require a recently authenticated session and reauthenticate the user before adding a passkey. Do not rely only on possession of a long-lived session cookie.
+- For a new account, bind the ceremony to the verified account-creation transaction. Do not allow a client-supplied account identifier to select a different account.
+- Protect the registration initiation and completion endpoints against cross-site request forgery as described in the [Cross-Site Request Forgery Prevention Cheat Sheet](Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.md).
+- Apply rate limits and log failed and successful enrollment attempts.
+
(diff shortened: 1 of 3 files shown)


jmanico (dismissed):
Great work, There are many things we can add or clarify, but this is a great starting point and I want to get it live.

cheatsheets/Passkey_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,235 @@
+# Passkey Security Cheat Sheet
+
+## Introduction
+
+Passkeys are [WebAuthn public key credentials](https://www.w3.org/TR/webauthn-3/#sctn-use-cases) that let a user authenticate without sending a shared secret to the application. The authenticator keeps the private key and the relying party stores a public key. A passkey is scoped to a relying party identifier (RP ID), which provides phishing resistance when the ceremony and server-side verification are implemented correctly.
+
+This cheat sheet is for developers who build the relying-party side of passkey registration, authentication, credential management, and account recovery. The main recommendations are:
```
randomstuff: Why is this a *passkey* Cheat Sheet and not a WebAuthn cheet sheet?

cheatsheets/Passkey_Security_Cheat_Sheet.md
```diff
@@ -0,0 +1,235 @@
...
+- The assertion signature is valid under the stored public key.
+- The UP flag is set and the UV flag is set when required by policy.
+- The returned user handle, when present, matches the credential's account.
+- Extension outputs and cross-origin indicators comply with RP policy.
+
+Create an authenticated application session only after every required check succeeds. Rotate the session identifier at authentication and apply the controls in the [Session Management Cheat Sheet](Session_Management_Cheat_Sheet.md).
+
+### Handle Signature Counters Conservatively
+
+An increasing signature counter can provide a signal that some authenticators have been cloned. The WebAuthn [signature counter considerations](https://www.w3.org/TR/webauthn-3/#sctn-sign-counter) explain why it is not a universal clone-detection mechanism. Some authenticators do not implement a counter, and a synced credential may be used from multiple authenticator instances whose counter behavior is not strictly monotonic.
```
randomstuff: > An *increasing* signature counter can provide a signal

Shouldn't that be "decreasing" (or more correctly "non-increasing")?

cheatsheets/Passkey_Security_Cheat_Sheet.md
kwwall: @randomstuff - Excellent point. All newly purchased passkeys today support WebAuthn, but they can support other things like FIDO2 (which is broader than WebAuthn). In fact, I have a really old Yubico Yubi Key that supports FIDO (now, FIDO1) but predated WebAuthn and all I use if for now is to with my password manager for some of the password vaults where I just using it as a 2FA mechanism. If this were truly about passkeys, this would have to be broader than what it is and cover use cases beyond just WebAuthn. But renaming the cheat sheet as WebAuthn can solve this.

sujalavnelavai: Hi @randomstuff , I am not sure whether can I reply to your comment in this PR , because this PR is not related to me. But I totally and definitely agree to your question, because most of the security controls here are fundamentally related to WebAuthn implementation controls. Especially passkey describes user facing category of WebAuthn credientials.

kwwall (changes requested):
Agree with @randomstuff that this should be renamed WebAutn_Security_Cheat_Sheet.md. And I meant to get to submit this yesterday, but my wife and I went out for dinner and when I got back, I forgot to finish it.

jmanico:
Kevin, I think we may be conflating a passkey with an authenticator. A YubiKey is an authenticator that can store and use a passkey; it is not itself a passkey. Users do not necessarily need to purchase a dedicated hardware authenticator. Passkeys can also be managed by platform authenticators built into phones and laptops, and they may be device-bound or even synchronized across several devices.

For that reason, I do not think we should rename this as a WebAuthn-only cheat sheet. The document covers substantially more than invoking the WebAuthn API. 

It addresses the full lifecycle required to deploy passkeys securely, including server-side verification and credential storage, registration and authentication, credential enrollment and revocation, account recovery, and the security differences between synchronized and device-bound passkeys.

Passkey Security Cheat Sheet therefore describes the intended scope more accurately.

kwwall:
@jmanico - FIDO2 is the more general authentication spec, an if I'm not
mistaken, Passkeys can also be used with general FIDO2, not just WebAuthn,
but I don't think this proposed Cheat Sheet even mentions FIDO2.  That's
probably why @randomstuff, @sujalavnelavai, and I are questioning it.

Maybe name it Passkeys for WebAuthn, then.
--
Blog: https://off-the-wall-security.blogspot.com/    | GitHub: @kwwall
<https://github.com/kwwall>| OWASP ESAPI Project
<https://owasp.org/www-project-enterprise-security-api/>co-lead | OWASP and
ACM lifetime member
NSA: All your crypto bit are belong to us.

cheatsheets/Passkey_Security_Cheat_Sheet.md
randomstuff: OK, I'll clarify the intent of my question. :bow:

My understanding was that *passkey* was defined as "WebAuthn credentials stored in software" (eg. stored in the browser) in contrasted to backed by a FIDO2 hardware. [Apparently](https://en.wikipedia.org/wiki/WebAuthn#Passkey_branding), some people out there might use the term *passkey* for both "software" and "FIDO2-backed" WebAuth credentials. So my question was whether this Cheat Sheet was intending to covered only software implementations and whether it should cover both software and FIDO2 WerbAuthn credentials.

In any case, maybe it might be useful to clarify that this covers both forms of WebAuthn credentials (or not).

basriakkaya: Thanks for clarifying. I kept the Passkey Security scope approved in #2365, but made the protocol boundary explicit. The introduction now states that this covers the WebAuthn relying-party side of passkey deployment across platform and roaming authenticators, including security keys that support discoverable credentials, and both synced and device-bound passkeys. It also distinguishes legacy FIDO U2F/CTAP1 second-factor credentials as out of scope and explains the WebAuthn/FIDO2/CTAP relationship.

cheatsheets/Passkey_Security_Cheat_Sheet.md
basriakkaya: Good catch. I changed this to “non-increasing” and aligned the explanation with the WebAuthn verification guidance: it is a signal, not proof, and may also indicate an authenticator malfunction or out-of-order assertion processing.

basriakkaya:
Thanks everyone. I pushed a focused scope clarification rather than renaming the file while the naming discussion remains split. The sheet now explicitly covers WebAuthn RP deployment of passkeys across platform and roaming authenticators, synced and device-bound credentials, and distinguishes legacy U2F/CTAP1 second-factor credentials. I also corrected the signature-counter condition. Markdown, terminology, and link checks all pass. @kwwall, would you please re-review whether this clarification addresses your change request?

jmanico:
Kevin, I think we are conflating the credential with the authenticator.

More precisely, a passkey is a discoverable FIDO public-key credential. A YubiKey is a FIDO authenticator/security key that can create and store device-bound passkeys. The YubiKey itself is not a passkey.

It is also important to distinguish passkeys from FIDO2 credentials generally. FIDO2 supports both discoverable and non-discoverable credentials; **the discoverable credentials are what the current standards terminology calls passkeys.**

**FIDO2 is the umbrella consisting of WebAuthn and CTAP.** WebAuthn defines the registration and authentication ceremonies used by the relying party and client, including the server-side verification model. CTAP is used when the client platform communicates with an external authenticator such as a YubiKey.

Therefore, I do not think “Passkeys for WebAuthn” makes the title more accurate. **Passkeys are already WebAuthn discoverable credentials.** The Cheat Sheet should explain the complete passkey implementation model: WebAuthn, relevant CTAP concepts, relying-party validation, credential-record storage, account recovery, credential lifecycle, user verification, attestation, and support for multiple passkeys.

basriakkaya:
Addressed the requested naming change:

- Renamed the file to `WebAuthn_Security_Cheat_Sheet.md`
- Renamed the document and PR to **WebAuthn Security Cheat Sheet**
- Preserved the explicit coverage of platform and roaming authenticators, synced and device-bound passkeys, and the WebAuthn/FIDO2/CTAP boundary
- Kept the corrected non-increasing signature-counter guidance

The two outdated inline threads are resolved. The current lint failure is from the upstream `Authentication_Cheat_Sheet.md` terminology check, outside this PR's one-file diff; build, link, and scope checks pass. Ready for another review.

jmanico:
Please move the name back to passkey, thats the right name for this cheatsheet.

WebAuthN is only the JavaScript API's. We need a more comprehensive cheatsheet than that.

basriakkaya:
Addressed the latest maintainer direction from @jmanico:

- Restored the file name to `Passkey_Security_Cheat_Sheet.md`
- Restored the document and PR title to **Passkey Security Cheat Sheet**
- Kept the comprehensive WebAuthn/FIDO2/CTAP, authenticator, lifecycle, recovery, synced, and device-bound credential coverage

Ready for another review.

mackowski:
Thanks for the work on this.

Separately, @kwwall shared [this passkeys/WebAuthn explainer](https://authlayer.dev/blog/passkeys_webauthn_educational/) with us. Reading it next to this draft surfaced two gaps worth closing, plus one presentation idea. All three are suggestions and questions, not blockers.

## 1. Native app passkeys / non-web origins

`Choose the RP ID Carefully` says to allowlist origins "including scheme, host, and port" — that is web-only. A native app's origin is not an HTTPS origin. On Android the server sees `android:apk-key-hash:<hash>` in `clientDataJSON`, and the app-to-domain binding depends on a Digital Asset Links file at `https://<RP ID>/.well-known/assetlinks.json`; on Apple platforms it depends on an `apple-app-site-association` file at `https://<RP ID>/.well-known/apple-app-site-association` ([RP ID deep dive](https://web.dev/articles/webauthn-rp-id)).

This sits squarely inside this sheet's threat model: if the association file is over-broad, or the server accepts app origins it never explicitly allowlisted, an unrelated app can obtain assertions for your RP ID — the same phishing-resistance failure the sheet warns about for web origins, through a channel the current text does not mention.

Would you be open to a few sentences in `Choose the RP ID Carefully` (or a short subsection) covering:

- app origins use a different format and need their own explicit allowlist entries, enumerated rather than pattern-matched;
- the association file (`assetlinks.json` / AASA) is part of the RP's attack surface and needs the same change control as the origin allowlist?

## 2. Fallback at authentication time, not only at recovery

`Account Recovery and Bootstrap` covers email/SMS *recovery* bypassing a stronger policy, and `Use Safe Failure Behavior` covers generic errors and rate limiting — but nothing covers what the application offers when the passkey ceremony itself fails. That is the main practical route around passkeys today: adversary-in-the-middle proxies do not break WebAuthn, they steer the user to whatever weaker method is still enabled. NIST SP 800-63B-4 §3.2.5 is explicit that authenticators involving manual entry of an authenticator output (OTP, out-of-band) SHALL NOT be considered phishing-resistant, so a silently available fallback means the account's real assurance is the fallback's.

We already cover the general case as [MFA Downgrade Attacks](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Multifactor_Authentication_Cheat_Sheet.md#mfa-downgrade-attacks) in the MFA sheet, so per our no-duplication rule I do not think this wants a new section here. One or two passkey-specific sentences in `Use Safe Failure Behavior` plus a cross-link (`[MFA Downgrade Attacks](Multifactor_Authentication_Cheat_Sheet.md#mfa-downgrade-attacks)` in-sheet) would be enough — something like: a failed passkey ceremony should not silently downgrade the user to a non-phishing-resistant method, and where a fallback exists it should be a deliberate, logged, policy-documented decision rather than the default error path. Does that shape work for you?

## 3. Sequence diagrams

The strongest feature of that article is its UML sequence diagrams for the registration and authentication ceremonies, and that is specifically what @kwwall liked about it.

The registration and authentication sections here are precise, but they are also dense: to follow which party checks what, a reader has to hold the RP, the client, the authenticator and the server in their head at once. A diagram makes that ordering visible at a glance, and it makes the individual checks in `Verify the Registration Response` and `Verify the Authentication Response` much easier to place in the flow.

Would you be interested in adding one diagram for the registration ceremony and one for authentication? I think it would noticeably improve readability for the developer audience this sheet targets. Happy to help with the mechanics if you want to give it a go.

Issue #2323, opened 2026-08-02
[New Cheat Sheet Proposal] AI-Powered Advertising Systems Security


elang2:
## Context

AI-powered advertising systems (programmatic bidding, real-time targeting, creative generation, audience segmentation) represent a massive and growing attack surface not covered by existing OWASP cheat sheets. Current AI-related sheets (AI Agent Security, RAG Security, MCP Security, LLM Prompt Injection Prevention) focus on LLM application security but do not address the unique risks of advertising infrastructure where AI makes financial decisions at scale.

This proposal was suggested by @bkimminich from the OWASP Project Committee as a more appropriate starting point after I submitted a standalone Top 10 proposal for this domain.

## Why This Is Needed

- Programmatic advertising processes over $ multi-Billion+ globally, with AI making targeting and bidding decisions in under 100ms
- Ad fraud costs the industry $ Multi Billion + annually, increasingly driven by adversarial AI
- Privacy regulations (GDPR, DMA, CPRA, COPPA) impose specific requirements on ad targeting that differ from general data processing
- The intersection of real-time financial transactions, user behavioral profiling, and AI inference creates attack vectors not found in other application domains
- No existing OWASP resource covers this space

### 1. Privacy & Consent Enforcement
- Multi-signal privacy checks (do-not-target, do-not-sell, do-not-share, opt-out)
- Consent validation before ad auction (IAB TCF string parsing)
- Age-of-majority checks for COPPA compliance
- Real-time opt-out enforcement with cache-backed lookups
- Privacy filtering before bid requests reach third parties

### 2. AI Model Security in Advertising
- Training data poisoning in bidding and targeting models
- Model inference attacks exposing user segment membership
- Adversarial inputs manipulating ad recommendations and bid prices
- Two-tier content safety: fast rule-based filters + LLM semantic review
- Confidence thresholding to reduce false positives
- Rate limiting and deployment locks for AI generation endpoints

### 3. Ad Fraud & Invalid Traffic (IVT) Prevention
- Multi-tier detection: pre-bid filtering + post-impression validation
- Bot detection via behavioral analysis and user-agent fingerprinting
- Dynamic blocklist generation (daily refresh from fraud pattern analysis)
- Outlier detection thresholds (impression caps per device/day)
- MRC-compliant filtering pipelines for billing integrity

### 4. Real-Time Bidding (RTB) Protocol Security
- OpenRTB request validation and sanitization
- Strict timeout enforcement on DSP bid responses
[...]

PR #2380, opened 2026-08-30
Add AI-Advertising Security cheat sheet series (#2323)

elang2:
Addresses #2323 with one concise AI-Advertising Security cheat sheet. Longer companion guides at OWASP/www-community#1319, per @jmanico's proposed split.


jmanico:
This is two cheatsheets being added. I dont think this topic required both. And this is not a place for comprehensive guides, this is for brief cheatsheets. Can we combine these into one concise cheatsheet?

elang2:
> This is two cheatsheets being added. I dont think this topic required both. And this is not a place for comprehensive guides, this is for brief cheatsheets. Can we combine these into one concise cheatsheet?

on it.  The area seem very vast but will merge these together into one.

jmanico:
If you want something more expansive, you can also post these different guides on the main owasp wiki - and this "cheat" can point to them if you want.

randomstuff:
Isn't this quite specific/niche anyway? (though I'm very not in this space …)

jmanico:
All topics we cover are vast. The cheatsheet series is a place for concise summaries or "cheats" on these topics.

So I suggest:

1) Add these two comprehensive guides to the main OWASP wiki
2) Build a cheatsheet summary, just one, here at the cheatsheet series.

Then you get all of what you want.

Is that fair?

elang2:
> All topics we cover are vast. The cheatsheet series is a place for concise summaries or "cheats" on these topics.
> 
> So I suggest:
> 
> 1. Add these two comprehensive guides to the main OWASP wiki
> 2. Build a cheatsheet summary, just one, here at the cheatsheet series.
> 
> Then you get all of what you want.
> 
> Is that fair?

Makes sense. Thanks for the guidance !  Super excited to be contributing here and looking forward to the iterations and feedback. I have updated the current PR to make the cheatsheet even more concise and then added deep dive guides in the PR below.   

https://github.com/OWASP/www-community/pull/1319

jmanico (commented):
Thanks for reworking this into a single concise sheet — that's exactly the split I had in mind, and the ad-tech material (OpenRTB, TCF/GPP, SIVT/MRC, ads.txt, VPAID→SIMID, C2PA, callback signing) has no other home in the series. I checked all 87 links: none is fabricated or dead, and the hard citations (DSA 26(3), AI Act 50(2)/50(4), Annex III 4(a), RFC 9421 §3.2.1, NIST §2.3.3, all five arXiv IDs) check out against the primary texts. This should land.

A few things to fix first:

1. regs.coppa == 1 can't be the child-directed gate. OpenRTB 2.6 §3.2.3 defines Regs as what the sender deems applicable, so it's an unauthenticated hint that fails open when omitted. Gate on a classification you control; treat unknown as child-directed.

2. Four detectors are presented as boundaries, and their own citations say otherwise — the two-VLM ensemble (Qi et al. show adversarial examples transfer across models, so both agree), trigger scanning (NIST §2.3.3 says these approaches don't reliably work), the pinned-anchor check (the category call is itself made from attacker-controlled text), and MMR (arXiv:2402.07179 never mentions MMR). Keep them as defense-in-depth with the limitation stated, and let the non-LLM rules baseline be the actual boundary.

3. Annex III mapping: 4(a) is employment only; housing isn't in Annex III; credit is 5(b) and covers creditworthiness, not ad targeting.

4. Three link swaps: C2PA URL points at the version index rather than the hard-binding section (and 2.1 is superseded by 2.4); the ATLAS link goes to the homepage — the technique IDs are correct, just link AML.T0051 directly; the OpenSSF page doesn't mention DSSE/in-toto/dataset manifests, so cite sigstore/model-transparency for that parenthetical.

5. One editorial ask: our audience is developers, not security architects, and this reads expert-to-expert. Biggest single win is deleting the Defense column from the tables — it restates the bullets above it nearly verbatim. Also worth filling in the numbers behind your best ideas (SIVT quarantine window, dedup TTL, k and n).

Minor: title should be "AI-Powered", with the matching Index.md change in the same commit or index-drift-check fails. markdownlint, textlint, and the citation gate are clean locally.

elang2:
@jmanico  Addressed your review comments  https://github.com/OWASP/CheatSheetSeries/pull/2380/commits/537ce440d16103f2e80a1d990f64f8aee65e8e1d

jmanico:
Much better. If @mackowski approves we can merge this.

elang2:
> Much better. If @mackowski approves we can merge this.

Thank you ! Also this please https://github.com/OWASP/www-community/pull/1319  . The guides i have per your advise. Hope these are the right place to put them

mackowski:
Thanks @elang2. I ran this through our AI-assisted review workflow (`.claude/commands/review-cheatsheet-pr.md`), which fans out reviewers across security correctness, developer practicality, link/source quality, duplication, and language. Worth saying up front: on sourcing it came back clean — it independently checked the citations and found **no fabricated or dead links**, and the hardest claims (EU AI Act Annex III 4(a) on targeted job advertisements, NIST AI 100-2e2025 §2.3.3) verify almost verbatim against the primary texts. That is unusually good for a sheet this citation-dense.

It did flag a set of potential issues that I think are worth resolving before merge:

1. **Line 11, consent gating.** "Gate on the model call, not the identifier write" moves the gate off an independently mandatory control — ePrivacy Art. 5(3) attaches consent to the storage/access itself, regardless of whether a model is later invoked. Suggest gating **both**, keeping your (valid and useful) point that the model-call gate is the one usually missing.
2. **Line 11.** "LLM contextual review" isn't profiling under GDPR Art. 4(4), which requires processing of personal data. Contextual-only review is the standard consent-free fallback, so as written this appears to gate the no-consent path.
3. **Line 57, Qi et al. Table 3.** That table reports Perspective-API toxicity percentages over a 34.8/34.2/9.2% no-attack baseline, not attack success rate — so the attack-attributable transfer effect is roughly +9 to +23 percentage points. Also worth noting the two most architecturally similar models transferred best, which suggests ensemble members should differ in vision encoder, not just weights.
4. **Line 14.** The OpenRTB anchor `#object_regs` doesn't exist — it's `#objectregs`, so the link lands at the top of the spec. The `regs.coppa` fail-open claim itself is correct and well supported.
5. **Line 59, Chrome Prompt API.** The API is gated per frame by Permissions Policy — a cross-origin ad iframe gets no access unless delegated via `allow="language-model"`. Suggest saying explicitly not to delegate it to ad slots. Your CSP-doesn't-inherit point is correct and should stay.
6. **Line 79.** Suggest dropping `CN` — RFC 9525 (which obsoletes 6125) says the Common Name RDN MUST NOT be used to identify a service. SPKI hash or a typed SAN entry instead.
7. **Lines 74-75.** The signed field set doesn't cover the request body, so `currency`, `quantity`, and label fields stay mutable while the signature still verifies. Suggest requiring `Content-Digest` as a covered component (RFC 9421 §7.2.8), plus a rule that uncovered fields must not influence money or training labels.
8. **Lines 52, 37, 101.** Three spots where your own "not the boundary" discipline could apply to itself: Tramèr explicitly shows label-only extraction still works; signed provenance protects against artifact substitution rather than a backdoor from an authorized supplier; and AI Act provider/deployer status is set by Art. 3(3)/3(4) rather than Art. 50.

@jmanico @elang2 — what do you both think of these? Some are clear-cut wording fixes, but a few (especially 1, 2 and 8) are judgement calls about how firmly to state a legal or threat-model position, and I'd rather agree on those than have me impose a reading. Elan, you have far more production ad-tech context than I do, so push back where the review has it wrong.

mackowski (approved):

jmanico (approved):

Issue #2384, opened 2026-08-31
New CS proposal: Web Cache Security


KadirArslan:
## What is the proposed Cheat Sheet about?

Web Cache Security Cheat Sheet would provide developers and operators with practical, vendor-neutral guidance on using HTTP and application-level caches safely. It would focus on preventing data exposure and response integrity issues. The cheat sheet would cover private browser caches, shared reverse proxies and CDNs, as well as in-memory and distributed application caches.

## What security issues are commonly encountered related to this area?

* Sensitive, authenticated, or personalized responses being stored in a shared cache and then served to another user.
* Assuming that authentication cookies or a `Set-Cookie` response header automatically make a response private.
* Confusing `Cache-Control: no-cache`, which allows caching but requires revalidation, with `no-store`, which prevents storage.
* Web cache poisoning, where an attacker-controlled header, query parameter, host value, path component, or other input that is not part of the cache key changes a cached response.
* Web cache deception, where a dynamic or authenticated endpoint accepts a static-looking path or file extension and the personalized response gets cached.
* Cache-key collisions caused by inconsistent URL normalization, query-string handling, forwarded headers, or routing between the cache and the origin.
* Cross-user or cross-tenant data leakage when application cache keys do not include tenant identity or other authorization-related context.
* Cache hits bypassing object-level or tenant-level authorization checks.
* Cached content remaining accessible even after permissions, account status, or the underlying resource has changed.
* Secrets or session identifiers being exposed through cache keys, logs, purge URLs, statistics, or administrative interfaces.
* CDN, reverse-proxy, or framework rules overriding safer origin cache policies based on paths, file extensions, or response status.
* Incident response that only purges a poisoned cache entry without fixing the cache key or origin behavior that caused the issue.

## What is the objective of the Cheat Sheet?
The objective is to provide a concise, architecture-first reference for implementation and security reviews. It would help developers decide what can be safely cached and which request inputs must be considered when selecting a cache entry.

## What other resources exist in this area?
- [RFC 9111: HTTP Caching](https://www.rfc-editor.org/rfc/rfc9111.html) defines HTTP cache storage, reuse, cache keys, validation, authenticated-response behavior, and security considerations.
[...]

PR #2385, opened 2026-08-31
Create Web_Cache_Security_Cheat_Sheet.md

KadirArslan:
This PR contains new CS related to the cache security
Cache's are using maybe every distributed system. That's why I thought that It would be great to have single, actionable doc.

- [X] This PR is focused: it modifies a single cheat sheet, or a small coordinated set, and the scope is described in the PR body.
- [X] Every technical claim, recommendation, or threat assertion added in this PR is supported by a primary source (RFC, NIST, OWASP standard, vendor documentation, peer-reviewed research) linked inline as `[text](URL)`.
- [X] I have read each source I cite and confirm it actually supports the claim. I have not relied on summaries, hearsay, or model-generated citations.

This PR fixes issue #2384.


KadirArslan:
Resolved conflicts

mackowski:
Thanks for new PR, good work. Small ideas for improvements

### 1. Maybe add `must-revalidate` to the `Authorization` bullet (line 32)

> Do not mark a response influenced by `Authorization` or session cookies as `public`, or give it an `s-maxage`, unless sharing the response has been deliberately designed and reviewed.

This bullet is right as far as it goes — the only thing I'd add is the third directive. [RFC 9111 §3.5](https://www.rfc-editor.org/rfc/rfc9111.html#name-storing-responses-to-authen) names three that lift the shared-cache prohibition for requests carrying `Authorization`:

> A shared cache MUST NOT use a cached response to a request with an Authorization header field [...] to satisfy any subsequent request unless the response contains a Cache-Control field with a response directive [...] that allows it to be stored by a shared cache, and the cache conforms to the requirements of that directive for that response.
>
> In this specification, the following response directives have such an effect: must-revalidate (Section 5.2.2.2), public (Section 5.2.2.9), and s-maxage (Section 5.2.2.10).

The one not listed is arguably the one developers are most likely to type: `Cache-Control: max-age=0, must-revalidate` often gets reached for in the belief that it means "don't cache." On an authenticated endpoint it doesn't quite — under §3 it makes the response storable by a shared cache, and under §3.5 reusable.

In fairness to the directive, it does constrain that reuse. [§5.2.2.2](https://www.rfc-editor.org/rfc/rfc9111.html#name-must-revalidate) says that once the response has become stale, a cache "MUST NOT reuse that response to satisfy another request until it has been successfully validated by the origin," and with `max-age=0` it's stale immediately, so a conformant shared cache revalidates each time. So this is nowhere near as bad as `public, max-age=600`. Two things still make it worth a mention, I think: the response is now stored on shared infrastructure at all, and revalidation only acts as an access control if the origin actually re-authorizes — an origin answering the conditional request with a bare `304` off an `ETag` or `Last-Modified` would hand user A's stored body to user B. And where the cache revalidates independently rather than inline, [§4.3.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-sending-a-validation-reques) has it synthesize the request from the stored response, "copying the method, target URI, and request header fields identified by the Vary header field" — so absent `Vary: Authorization` the origin wouldn't see a credential to check at all.

Might be worth naming the §3.5 rule and all three directives explicitly rather than leaving it implicit in a "do not" list.

### 2. Perhaps say why omitting a directive is risky (section 1, lines 24-45)

The advice itself is already here — line 26 and the Do's at line 127 both ask for an explicit policy on every route. The only thing I'd suggest adding is the mechanism behind it, since that's usually what gets a developer to act.

[RFC 9111 §4.2.2](https://www.rfc-editor.org/rfc/rfc9111.html#name-calculating-heuristic-fresh) allows a cache to assign heuristic freshness to a response carrying no explicit freshness information and no directive barring storage, as long as its status code is heuristically cacheable — and per [RFC 9110 §15.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.1) a plain 200 is. The RFC also notes that the older "don't apply heuristics to URIs with a query string" rule "has not been widely implemented," so a query string isn't much protection either.

The compounding bit is that §3.5 doesn't help here: it's scoped to the `Authorization` header field only, so a cookie- or session-authenticated 200 gets no implicit protection, which covers most browser-facing applications. At the moment "heuristic" appears once in the sheet, in passing at line 90.

The MDN guide you already cite makes this point in both of its relevant sections, so it may be an easy addition: its **Heuristic caching** section closes with "basically all responses should explicitly specify a `Cache-Control` header," and its **Default settings** section says "To avoid that heuristic caching, it's preferable to explicitly give all responses a default `Cache-Control` header."

### 3. `Vary` is a spec guarantee that not every CDN honors (section 2, lines 49-53)

Line 52 already scopes `Vary` correctly to representation-selecting headers, and line 53 rules out `Vary: Cookie` as an authorization boundary — so this is a small addition rather than a correction.

[RFC 9111 §4.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-calculating-cache-keys-with) does make `Vary` binding: a cache "MUST NOT use that stored response without revalidation unless all the presented request header fields nominated by that Vary field value match those fields in the original request." The gap is in deployment rather than the spec — some major CDNs don't implement that for arbitrary headers without configuration. [Cloudflare's Vary documentation](https://developers.cloudflare.com/cache/concepts/vary/) notes that "by default, Cloudflare's CDN constructs cache keys from a request's URL and a handful of specific headers," asks for per-header actions in Cache Rules, and warns that normalization "can be lossy: it may reorder values, drop quality values, lowercase values, or remove entries that are not in a configured allowlist." [Fastly, by contrast, supports it per spec](https://www.fastly.com/documentation/reference/http/http-headers/Vary/). Since the behavior varies by vendor, it might be worth not presenting `Vary` as self-executing.

Given the section's own rule that every response-changing input must be keyed, I could imagine a reader reaching for `Vary: Authorization` or `Vary: X-Tenant-ID` and assuming users are separated while the CDN ignores the header. Two bullets would probably cover it:

- Verify that your CDN or proxy actually keys on the header you named — on its own, `Vary` isn't a tenant or user boundary.
- `Vary: *` is the deliberate opt-out: per §4.1 "a stored response with a Vary header field value containing a member `*` always fails to match," and Cloudflare confirms it in practice — "A response containing `Vary: *` always bypasses cache, regardless of your Vary configuration."

### 4. Request smuggling / HTTP desync might be worth adding to the poisoning section (lines 66-76)

Section 3 frames unkeyed inputs as "the central design rule," which is fair — the one class I'd suggest adding is parsing discrepancies, since it's the only poisoning vector [RFC 9111 §7.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-cache-poisoning) itself names:

> One common attack vector for cache poisoning is to exploit differences in message parsing on proxies and in user agents; see Section 6.3 of [HTTP/1.1] for the relevant requirements regarding HTTP/1.1.

Cache-poisoned DoS (poisoning an error or oversized-header response into a shared entry) isn't covered either. As it stands, a developer who follows all seven bullets would still be exposed to a front-end/back-end framing disagreement.

There's no request-smuggling cheat sheet in the repo to cross-link to, so it would need covering inline — perhaps a bullet naming smuggling/desync and response splitting as poisoning vectors, with a few concrete controls:

- Reject messages carrying both `Content-Length` and `Transfer-Encoding`, and close the connection. [RFC 9112 §6.3](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.3) says such a message "might indicate an attempt to perform request smuggling [...] or response splitting [...] and ought to be handled as an error," and [§6.1](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.1) adds that whether the server rejects the request or processes it per `Transfer-Encoding`, it "MUST close the connection after responding to such a request to avoid the potential attacks." The connection-close tends to be the part implementations skip.
- Prefer HTTP/2 end-to-end and disable HTTP downgrading where that's possible. H2 at the front end with an HTTP/1.1 back end reintroduces desync via the downgrade — the rewrite regenerates `Content-Length`/chunked framing from headers the attacker influences — so "we're on HTTP/2" isn't by itself a fix. Where the back end only speaks HTTP/1.1, [PortSwigger's mitigation guidance](https://portswigger.net/web-security/request-smuggling) is to validate the rewritten request against the HTTP/1.1 specification and "reject requests that contain newlines in the headers, colons in header names, and spaces in the request method."
- Avoid reusing front-end-to-back-end connections, though that doesn't cover request tunneling.

PR #2387, opened 2026-08-31
docs: add Cross-JWT and token type confusion threat section to JWT Cheat Sheet

jankesec:
### What is missing or needs to be updated?

Following up on #2302, this PR adds a focused section on Cross-JWT and token type confusion to `cheatsheets/JSON_Web_Token_Cheat_Sheet.md`.

### Proposed Changes

- Explain that confusion occurs when validation rules fail to distinguish token kinds; overlapping claims and shared keys are enabling conditions rather than the root cause.
- Recommend explicit `typ` values for token profiles that support them, with examples from RFC 9068 and OpenID Connect Back-Channel Logout 1.0.
- Scope rejection of a missing `typ` to profiles that require or reliably provide explicit typing.
- Recommend mutually exclusive validation rules using distinct keys, required claims, or strict issuer and audience isolation when explicit typing is not interoperable.
- Include a compact PyJWT verifier example that validates the signature and claims before checking `typ`.

### Verification

- `npm test`
- `npm run link-check`

Both completed successfully.

### AI Tool Usage

- Tool: OpenAI Codex
- Prompt: "Review PR #2387, implement the requested review changes, verify every source, run the required checks, and update the PR."
- Use: Assisted with revising the threat framing and mitigation wording. The cited RFC and OpenID Connect sections were checked against their authoritative sources, and the final changes were reviewed and tested before submission.

Ref #2302

### DIFF
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
@@ -333,8 +333,9 @@ For example, an attacker could obtain an ID token or an email verification token
 Mitigations:
 
 - **Use explicit typing (`typ` header parameter):** The issuer should set the `typ` header parameter to a specific media type (or explicit type) distinguishing the token's purpose, such as `"typ": "at+jwt"` for OAuth 2.0 Access Tokens ([RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068)) or `"typ": "logout+jwt"` for Back-Channel Logout Tokens ([RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693)).
+- **Define custom media types for application-specific tokens:** For custom or internal token types (such as password-reset, email verification, or step-up authentication tokens), define and enforce a dedicated media type (e.g. `example-reset+jwt` or `application/vnd.example.reset+jwt`) to prevent them from being substituted for general access tokens.
 - **Validate `typ` at the verifier:** The verifier must explicitly check that the `typ` header matches the expected token type for that endpoint and reject tokens with mismatched or missing types when explicit typing is enforced.
-- **Use distinct signing keys or separate audiences:** If explicit typing cannot be used (e.g. legacy systems), use distinct cryptographic keys for different token purposes, or strictly validate distinct `aud` (audience) and `iss` values per token type.
+- **Caveat for legacy and standard tokens lacking explicit media types:** Some standard specifications (notably OpenID Connect Core 1.0 ID Tokens) do not define an explicit media type and often omit `typ` or set it to generic `JWT`. In cross-system environments where explicit typing cannot be interoperably enforced, rely on **distinct cryptographic signing keys** or strict **audience (`aud`) and issuer (`iss`) isolation** to prevent cross-token replay.
 
 Example of token with explicit typing:
 
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
@@ -363,8 +363,10 @@ payload = jwt.decode(
 )
 
 # Enforce explicit token type to prevent cross-JWT confusion
+# RFC 7515 §4.1.9 & RFC 9068: typ is case-insensitive, and "application/" prefix may be omitted
 header = jwt.get_unverified_header(token)
-if header.get("typ") != "at+jwt":
+token_type = str(header.get("typ", "")).lower()
+if token_type not in ["at+jwt", "application/at+jwt"]:
     raise jwt.InvalidTokenError("Invalid token type: expected at+jwt")
 ```
 
--- a/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
+++ b/cheatsheets/JSON_Web_Token_Cheat_Sheet.md
@@ -324,6 +324,55 @@ References:
 - [RFC 8725, Do Not Trust Received Claims](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim);
 - [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114), a key embedded in the JWS header trusted for verification.
 
+### Cross-JWT and token type confusion
+
+When an application or authorization server issues different kinds of JWTs for different purposes (such as access tokens, ID tokens, logout tokens, refresh tokens, or password-reset tokens) that are signed with the same key, a token issued for one purpose might be accepted by an endpoint expecting a different token type if the verifier does not check the token's intended context.
+
+For example, an attacker could obtain an ID token or an email verification token, and present it to an API endpoint expecting an access token. If both tokens share common claims (such as `sub`) and are signed by the same trusted key, a verifier that does not distinguish the token type might treat the ID token as an authorized access token. This vulnerability is known as "cross-JWT confusion" or "token type confusion" (see [RFC 8725 §2.11](https://datatracker.ietf.org/doc/html/rfc8725#name-use-explicit-typing)).
+
+Mitigations:
+
+- **Use explicit typing (`typ` header parameter):** The issuer should set the `typ` header parameter to a specific media type (or explicit type) distinguishing the token's purpose, such as `"typ": "at+jwt"` for OAuth 2.0 Access Tokens ([RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068)) or `"typ": "logout+jwt"` for Back-Channel Logout Tokens ([RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693)).
+- **Validate `typ` at the verifier:** The verifier must explicitly check that the `typ` header matches the expected token type for that endpoint and reject tokens with mismatched or missing types when explicit typing is enforced.
+- **Use distinct signing keys or separate audiences:** If explicit typing cannot be used (e.g. legacy systems), use distinct cryptographic keys for different token purposes, or strictly validate distinct `aud` (audience) and `iss` values per token type.
+
+Example of token with explicit typing:
+
+```json
+{
+  "typ": "at+jwt",
+  "alg": "ES256",
+  "kid": "auth-key-2026"
+}
+```
+
+Example of validation enforcing explicit token type:
+
+```python
+import jwt
+
+# Decode and verify signature, issuer, and audience
+payload = jwt.decode(
+    token,
+    public_key,
+    algorithms=["ES256"],
+    audience="https://api.example.com",
+    issuer="https://auth.example.com",
+    options={"require": ["exp", "iss", "aud"]},
+)
+
+# Enforce explicit token type to prevent cross-JWT confusion
+header = jwt.get_unverified_header(token)
+if header.get("typ") != "at+jwt":
+    raise jwt.InvalidTokenError("Invalid token type: expected at+jwt")
+```
+
+References:
+
+- [RFC 8725 §2.11, Use Explicit Typing](https://datatracker.ietf.org/doc/html/rfc8725#name-use-explicit-typing);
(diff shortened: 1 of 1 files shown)


cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,55 @@ References:
...
+
+For example, an attacker could obtain an ID token or an email verification token, and present it to an API endpoint expecting an access token. If both tokens share common claims (such as `sub`) and are signed by the same trusted key, a verifier that does not distinguish the token type might treat the ID token as an authorized access token. This vulnerability is known as "cross-JWT confusion" or "token type confusion" (see [RFC 8725 §2.11](https://datatracker.ietf.org/doc/html/rfc8725#name-use-explicit-typing)).
+
+Mitigations:
+
+- **Use explicit typing (`typ` header parameter):** The issuer should set the `typ` header parameter to a specific media type (or explicit type) distinguishing the token's purpose, such as `"typ": "at+jwt"` for OAuth 2.0 Access Tokens ([RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068)) or `"typ": "logout+jwt"` for Back-Channel Logout Tokens ([RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693)).
```
randomstuff: Some standard token type don't have a define media type (and value for `typ`). One major offender is OpenID ID Token. In these cases, it makes it difficult to apply this mitigation:

* we can't safely (in an interoperable way) include anything "JWT";
* we can't safely (in an interoperable way) validate the field.

I think this caveat should noted somewhere here.

I addition, we should probably recommend defining a custom media type for custom and/or internal  token types.

jankesec:
Great points regarding OIDC ID Tokens and custom media types. I have updated the section with the caveat for standard tokens lacking explicit media types (emphasizing distinct signing keys and audience isolation as the mitigation), and added the recommendation to define custom media types (e.g. `example-reset+jwt`) for application-specific/internal tokens. Thanks for the review!

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,55 @@ References:
...
+    public_key,
+    algorithms=["ES256"],
+    audience="https://api.example.com",
+    issuer="https://auth.example.com",
+    options={"require": ["exp", "iss", "aud"]},
+)
+
+# Enforce explicit token type to prevent cross-JWT confusion
+header = jwt.get_unverified_header(token)
+if header.get("typ") != "at+jwt":
```
randomstuff: Nitpick:

* this field is supposed to be case insensitive (and validated as such);
* we should accept the full type name "application/at+jwt"

jankesec:
Updated the validation snippet to handle case-insensitivity (`.lower()`) and accept both `at+jwt` and the full media type name `application/at+jwt` per RFC 7515 §4.1.9 / RFC 9068. Thanks!

jankesec:
Pushed a couple of citation fixes here as well, both the same kind of slip @mackowski caught on #2391.

The threat was cited as RFC 8725 §2.11, which doesn't exist, section 2 stops at 2.9. Cross-JWT Confusion is §2.8, and the anchor I'd used was the one for Use Explicit Typing, which is §3.11 and a mitigation rather than the threat the sentence describes. Both are corrected.

I'd also attributed `"typ": "logout+jwt"` to RFC 8693, which is OAuth 2.0 Token Exchange and doesn't mention logout at all. `application/logout+jwt` is registered to the OpenID Foundation, from Back-Channel Logout 1.0, so the reference now points there.

§3.12 is cited on the legacy caveat, since mutually exclusive validation rules are what the RFC prescribes when `typ` can't be enforced interoperably.

The example trim and the cross-link to the issuer and audience section will come with the rebase, once #2391 lands.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
 - [RFC 8725, Do Not Trust Received Claims](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim);
 - [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114), a key embedded in the JWS header trusted for verification.
 
+### Cross-JWT and token type confusion
+
+When an application or authorization server issues different kinds of JWTs for different purposes (such as access tokens, ID tokens, logout tokens, refresh tokens, or password-reset tokens) that are signed with the same key, a token issued for one purpose might be accepted by an endpoint expecting a different token type if the verifier does not check the token's intended context.
```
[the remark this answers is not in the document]
jankesec: Good point, corrected. RFC 6749 leaves the refresh token format to the authorization server and they are usually opaque, so listing them next to access tokens and ID tokens implied more than it should.

The main list now stops at password-reset tokens, and refresh tokens get their own sentence: they belong here where a deployment issues them as JWTs, though they are more often opaque strings whose format is internal to the authorization server. The case is worth keeping, since an access token and a refresh token from the same issuer are exactly the pair you would not want confused.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
...
+- **Validate `typ` at the verifier:** The verifier must explicitly check that the `typ` header matches the expected token type for that endpoint and reject tokens with mismatched or missing types when explicit typing is enforced.
+- **Caveat for legacy and standard tokens lacking explicit media types:** Some standard specifications (notably OpenID Connect Core 1.0 ID Tokens) do not define an explicit media type and often omit `typ` or set it to generic `JWT`. In cross-system environments where explicit typing cannot be interoperably enforced, rely on **distinct cryptographic signing keys** or strict **audience (`aud`) and issuer (`iss`) isolation** to prevent cross-token replay. RFC 8725 §3.12 sets out these fallbacks, requiring the validation rules for different kinds of JWTs to be mutually exclusive.
+
+Example of token with explicit typing:
```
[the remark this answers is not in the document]
jankesec: Correct, that snippet is only the JOSE header. Relabelled to "Example of a JOSE header with explicit typing".

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Answering in-thread as well, since this was only acknowledged in a top-level comment at the time and the thread reads as open.

Both are in the section now. The fourth mitigation bullet carries the caveat, naming OpenID Connect Core 1.0 ID Tokens as the case that omits `typ` or sets it to a generic `JWT`, and points at distinct signing keys or strict `aud` and `iss` isolation where explicit typing cannot be enforced interoperably. It now also cites RFC 8725 §3.12, which is where the RFC sets out those fallbacks.

The second bullet covers your other suggestion: defining a dedicated media type for custom or internal token types, with `example-reset+jwt` and `application/vnd.example.reset+jwt` as the illustrations.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Same here, noting it in-thread for the record.

The validation example handles both: it lowercases the header value before comparing, and accepts the full media type name as well as the short form, matching against `["at+jwt", "application/at+jwt"]`. There is a comment on the line citing RFC 7515 §4.1.9 and RFC 9068 so the reason is visible to a reader copying the snippet.

jmanico (commented):
Can you make this addition more concise? This cheatsheet is getting too big.

jankesec:
Thanks for the feedback, @jmanico.

I have streamlined the section to keep the cheat sheet concise:
- Condensed the introduction into a single concise paragraph focusing directly on the threat and RFC 8725 §2.8.
- Consolidated the mitigations into three punchy bullet points (explicit `typ`, verifier enforcement, and RFC 8725 §3.12 fallback isolation for tokens lacking explicit typing).
- Removed the separate JSON header snippet and redundant `jwt.decode` claim-validation boilerplate, leaving just a 4-line focused check on the `typ` header.
- Reduced the addition from 54 lines to 25 lines (>50% reduction) while preserving all RFC references.

`npm test` is clean.

jmanico (dismissed):

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,31 @@ References:
...
+- **Validate `typ` at the verifier:** Reject tokens with missing or unexpected `typ` values for that endpoint. Note that `typ` is case-insensitive and the `application/` prefix may be omitted ([RFC 7515 §4.1.9](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.9)).
+- **Use distinct keys or mutually exclusive validation:** Where explicit typing cannot be enforced interoperably (e.g., standard OIDC ID tokens omitting `typ`), use **separate signing keys** or strict **`iss` and `aud` isolation** so that validation rules for different token kinds remain mutually exclusive ([RFC 8725 §3.12](https://datatracker.ietf.org/doc/html/rfc8725#name-use-mutually-exclusive-vali)).
+
+Example of validation enforcing explicit token type:
+
+```python
+# Enforce explicit token type to prevent cross-JWT confusion
+header = jwt.get_unverified_header(token)
+if str(header.get("typ", "")).lower() not in ["at+jwt", "application/at+jwt"]:
```
mackowski, line 342: This is bad, you are doing validation on the attacker-controlled data. First you need to validate the token e.g.  signature verification. So do the normal validation of the token and later add also enforcement of the explicit token type

mackowski (changes requested):

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Good catch, thanks @mackowski. You're completely right. In an effort to keep the snippet as concise as possible following Jim's feedback, stripping out the token decoding and signature verification left only the unverified header check, which is bad practice to present as standalone guidance.

I have updated the snippet to perform signature and claim verification first using `jwt.decode_complete(...)`, and then enforce explicit `typ` checking on the verified token's header. This guarantees cryptographic verification of the token before trusting its contents while keeping the snippet focused and clean.

mackowski (dismissed):

jmanico (commented):
This is technically sound, overall its solid work and I would merge it.

Some nit picking:

1) The threat framing is causally wrong in a way that matters. The opening says confusion happens because the tokens share claims and are signed by the same key. Shared keys are a common precondition, not the cause. RFC 8725 §3.12 requires mutually exclusive validation rules regardless of key separation. As written, a reader can conclude that using separate keys makes the rest unnecessary, which also puts it at odds with the third mitigation bullet. One sentence fixes it: the cause is validation rules that fail to distinguish token kinds, with shared keys and overlapping claims as the usual enabling conditions.

2) logout+jwt has no citation. The bullet attaches the RFC 9068 link to at+jwt only, so the logout value now sits uncited after the RFC 8693 correction. It should point at OpenID Connect Back-Channel Logout 1.0, which is where application/logout+jwt is actually registered. Small, but the file's convention is that every protocol-specific value carries its source.

3) Bullet 2 and bullet 3 read as slightly contradictory. "Reject missing typ" as a flat rule is not interoperable, and bullet 3 then carves out the OIDC ID token case. Scoping bullet 2 to token kinds that carry an interoperable type would remove the tension without adding length.

jankesec:
Thanks, @jmanico. I addressed all three points in `30373a7`: the opening now identifies non-distinct validation rules as the cause, `logout+jwt` has the OpenID Connect Back-Channel Logout citation, and missing `typ` is rejected only for profiles that require or reliably provide explicit typing.

I also aligned the PR description with the current branch and added the required AI usage disclosure. `npm test` and `npm run link-check` both pass. Ready for another look.

PR #2391, opened 2026-09-01
docs: add Issuer and Audience confusion threat section to JWT Cheat Sheet

jankesec:
### What is missing or needs to be updated?

Following up on #2302 (and the maintainer recommendation in #2302 to handle threats in focused, standalone PRs), this PR adds the threat section on **Issuer and audience confusion** to `cheatsheets/JSON_Web_Token_Cheat_Sheet.md`.

*Description updated to match the current state of the branch after review. See the discussion below for what changed and why.*

### Proposed Changes:

- Added `### Issuer and audience confusion` under `## Threats on JWTs`.
- Separated the two variants, since different checks defeat them:
    - **Audience confusion**, where a token issued for one service is replayed against another service that trusts the same issuer.
    - **Issuer confusion**, where key resolution is not bound to `iss`. A verifier that looks up keys by `kid` across a union of several trusted issuers' JWK Sets will accept a token whose `iss` names one issuer and whose `kid` names a key belonging to another, with no stolen key and no multi-tenancy involved.
- Outlined mitigations based on **RFC 8725 §3.8 (Validate Issuer and Subject)** and **RFC 8725 §3.9 (Use and Validate Audience)**:
    - Exact, case-sensitive matching of the expected `iss` string.
    - Selecting the verification key from the set bound to the validated `iss`, never from a union of keys across issuers. Multi-tenant deployments are a sub-case rather than the precondition.
    - Verifying the presence of the recipient's identifier in the `aud` claim, whether `aud` is a string or an array.
    - Rejecting tokens where `iss` or `aud` is missing, presented as a deployment decision since RFC 7519 makes both claims optional.
- Added a code example in Python (PyJWT) that resolves the verification key from the issuer's JWK Set through `PyJWKClient`, so it demonstrates the key binding rather than assuming it. It uses `ES256`, matching the signature table earlier in this sheet.
- Cross-linked the existing `Trusting key material named in the token header` and `MAC` sections instead of restating them.
- Added references to RFC 8725 §2.7 (the threat itself), §3.8, §3.9, and RFC 7519 §4.1.1 and §4.1.3.

### Verification:

- `npm test` (`markdownlint` + `textlint`) executed locally, 0 errors, 0 warnings.

Ref #2302


jmanico (dismissed):

mackowski (commented):
Thanks for this, @jankesec. I went through the section against RFC 8725 and RFC 7519, and it seems to me it needs a bit more work before it lands. The direction is right, but a few things below would want fixing first.

### 1. The two RFC 8725 anchors point at the wrong sections

The labels are right, the numbers are off by a section:

```
2.7.  Substitution Attacks
2.8.  Cross-JWT Confusion
3.8.  Validate Issuer and Subject
3.9.  Use and Validate Audience
```

§2.8 is cross-JWT confusion, which is your other PR. Use `#name-validate-issuer-and-subject` and `#name-use-and-validate-audience`. §2.7 is worth citing too, as the *threat* rather than the mitigation.

One caveat while we are on citations: your third bullet requires both claims to be present, and I do not think any RFC mandates that for `iss` — RFC 7519 §4.1.1 calls it OPTIONAL, and §3.9's MUST covers `aud` and only where one issuer serves several relying parties. The advice is right for a cheat sheet; it is a deployment decision rather than an RFC requirement, so please do not cite it as one.

### 2. The key-binding bullet is scoped too narrowly, and the example does not implement it

Bullet 4 gates key-to-issuer binding on *"In multi-tenant environments"*. §3.8 applies whenever the token carries an `iss` claim — which your own bullet 3 requires — not only in multi-tenant deployments: *"the application MUST validate that the cryptographic keys used for the cryptographic operations in the JWT belong to the issuer."*

The cheapest illustration needs no stolen key at all: a verifier that looks up the verification key by `kid` alone, across a union of two trusted issuers' JWKS. The attacker sets `iss` to the issuer they want to impersonate and `kid` to a key from the other one. Bullets 1-3 are string comparisons and all pass; bullet 4 does not apply because the deployment is single-tenant.

Suggested replacement: *"select the verification key from the set bound to the validated `iss`, for example that issuer's `jwks_uri`; never verify against a union of keys from multiple issuers"*, with the tenant case as a sub-case.

The example needs the same change — it passes a bare `public_key`, so it demonstrates bullets 1-3 and silently skips 4, which is the one that matters most. `PyJWKClient` resolved from the validated issuer would show it. Worth linking to [Trusting key material named in the token header](#trusting-key-material-named-in-the-token-header) and the `iss`/`aud` key-reuse bullets under [MAC](#mac) rather than restating them.

Relatedly, *"if both use compatible key structures"* is the wrong precondition, and "both" has no antecedent. Key structure is irrelevant — verification needs the exact key. What enables this is key resolution not bound to `iss`.

### Smaller, non-blocking

- `algorithms=["RS256"]` contradicts the signature table earlier in this sheet, which marks RSASSA-PKCS1-v1_5 "Not recommended". #2387 already uses `ES256`.
- While you are in the snippet: `verify_iss` does nothing unless `issuer=` is passed — PyJWT's `_validate_iss` returns immediately when `issuer is None`. The named arguments do the work; `options` is redundant here and a reader who trims it loses the check silently.
- RFC 9207 does not seem to belong. It defines the `iss` *authorization response parameter* for OAuth mix-up attacks, not `iss` claim validation at a resource server. Drop it, or tell me the connection?
- The last reference is labeled "§4.1.1 and §4.1.3" but links to `#section-4.1`.

### Coordination — nothing for you to do yet

This and #2387 insert at the same anchor. Both are mergeable alone, so GitHub will not flag it, but `git merge-tree pr2391 pr2387` gives a content conflict, and both add a near-identical `jwt.decode(...)` block that would land twice, twenty lines apart.

@randomstuff — the call I would like your view on: **one section or two?** RFC 8725 treats these as one threat (§2.7, §2.8) with a four-part mitigation set (§3.8, §3.9, §3.11, §3.12), and this sheet already has the precedent — `Trusting key material named in the token header` covers seven header parameters in one section rather than seven. A single "Claim confusion" section covering `iss`, `aud` and `typ` would be one anchor, one example, no conflict. The alternative is to land this first and rebase #2387 on top.

Worth noting §3.12 either way — "Use Mutually Exclusive Validation Rules for Different Kinds of JWTs" is the RFC's own answer to the caveat in #2387 about `typ` not being interoperably enforceable, and neither PR cites it.

@jmanico — one yes/no: does §3.8's key-binding MUST apply unconditionally, or is the multi-tenant scoping defensible?

jmanico:
Agreed on removing the multi-tenant restriction. §3.8’s MUST applies whenever a JWT contains an iss claim, regardless of tenancy.

jankesec:
Thanks for the detailed review, @mackowski. You're right on all of it, and thanks @jmanico for settling the §3.8 question.

The section numbers were my mistake. I had the labels right but pulled the numbers from the threats section instead of the best practices one. Switching to `#name-substitution-attacks` for the threat itself, and `#name-validate-issuer-and-subject` and `#name-use-and-validate-audience` for the mitigations. I'll add §3.12 as well, since you're right that it's the RFC's own framing and neither PR mentions it.

On requiring both claims, fair point, I overstated that. RFC 7519 §4.1.1 does make `iss` optional, and §3.9's MUST is conditional on one issuer serving several relying parties. I'd still like to keep the advice, since a verifier that accepts a token with no `iss` has nothing to bind the key to, but I'll present it as a deployment choice rather than hanging an RFC citation on it.

The key-binding point is the one I'm most glad you caught. Your `kid`-across-a-union example is a lot better than what I had: it needs no stolen key and no multi-tenancy, and bullets 1-3 sail straight through it, which makes the gap visible in a way my version didn't. I've taken your wording for bullet 4 with the tenant case as a sub-case, and dropped "if both use compatible key structures". You're right that key structure isn't what enables it, and "both" didn't refer to anything.

The example had the same hole, so it now resolves the key through `PyJWKClient` against the expected issuer's JWK Set rather than taking a bare `public_key`. And I'll link across to [Trusting key material named in the token header](#trusting-key-material-named-in-the-token-header) and the `iss`/`aud` bullets under [MAC](#mac) instead of repeating them.

On the smaller ones:

- `RS256` to `ES256`, agreed. I should have caught that one myself, it contradicts a table in the same sheet.
- You're right about `options`. I went and looked, and `_validate_iss` returns straight away when `issuer` is None, so `verify_iss` genuinely does nothing on its own. Dropping the flags, keeping `require`, and adding a short note so the named arguments don't read as optional.
- RFC 9207 doesn't belong, no. I conflated the authorization response parameter with the claim, which are different things at different points in the flow. Removing it.
- Splitting the RFC 7519 link into §4.1.1 and §4.1.3, which covers Copilot's second comment too, and spelling out "identity provider" on first use for the other one.

On one section or two, I'd lean towards one, for the reasons you give. RFC 8725 treats them together, `Trusting key material named in the token header` is the precedent in this sheet, and a single anchor with one example spares the reader a near-duplicate `jwt.decode` a few lines further down. It also gives §3.12 an obvious home, which as you say answers the `typ` caveat in #2387.

That's your call and @randomstuff's, though. If you'd like them merged I'll fold #2387 in here as a single "Claim confusion" section covering `iss`, `aud` and `typ` and close the other PR. If you'd rather keep them apart I'll land this one and rebase #2387 on top. Either works for me.

The fixes above are needed in both shapes, so I'll push those now and leave the structure until you've decided.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,46 @@ References:
 - [RFC 8725, Do Not Trust Received Claims](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim);
 - [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114), a key embedded in the JWS header trusted for verification.
 
+### Issuer and audience confusion
+
+A JWT contains the `iss` (issuer) and `aud` (audience) claims to identify who created the token and for whom the token was issued. If a recipient verifies only the cryptographic signature and token expiration without strictly validating `iss` and `aud`, an attacker can reuse a valid token across different contexts, environments, or services.
```
randomstuff: nitick "A JWT *typically* contains" … " an attacker *may be able to* reuse"?

It rely depends on the context, not having `iss` and `aud` might be OK in some context (eg. you are generating a JWT that you consume yourself using a secret key which is only used for this specific purpose).

Another example is SD-JWT when the SD-JWT does not contain an `aud` claim which will instead be present in the key binding JWT.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,46 @@ References:
...
+
+For example, an attacker can obtain a valid token issued for a low-privilege service (or an external tenant in a multi-tenant system) and replay it against a high-privilege service or internal API. If the recipient does not enforce that its own identifier is present in `aud`, it accepts the token. Similarly, in multi-issuer environments or identity federation architectures, missing `iss` validation allows tokens issued by an untrusted or separate identity provider (e.g., a test/staging IdP) to be accepted if both use compatible key structures.
```
randomstuff: >  missing `iss` validation allows tokens issued by an untrusted or separate identity provider (e.g., a test/staging IdP) to be accepted if both use compatible key structures.

Maybe reword this? Test/staging environments should have different signing keys so in this specific case, it should not be a real problem.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,46 @@ References:
...
+
+Mitigations:
+
+- Always validate that the `iss` claim matches the expected issuer string exactly (case-sensitive comparison, including exact protocol and path).
```
randomstuff: In some cases, the `iss` will be used instead to auto-provision the IdP in the consumer through some auto discovery protocol.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,46 @@ References:
...
+# Validates signature, expiration, exact issuer, and recipient audience
+decoded_payload = jwt.decode(
+    token,
+    public_key,
+    algorithms=["RS256"],
+    audience="https://api.example.com/v1/payments",
+    issuer="https://auth.example.com/",
+    options={
+        "require": ["exp", "iss", "aud"],
+        "verify_aud": True,
```
randomstuff: `verify_aud` and `verify_iss` are both true by default if `verify_signature` is true so I'm not sure I would include them explicitly.

randomstuff (changes requested):

randomstuff:
>  A single "Claim confusion" section covering iss, aud and typ would be one anchor, one example, no conflict. 

I would keep this separate, token type on the on one section and this on another.

jankesec:
Thanks @randomstuff, that settles it. Two sections then.

@mackowski, I'll go with your alternative: land this one first and rebase #2387 on top.

The duplication you flagged does survive the split, though, so there are a couple of things I'd like to do when I rebase #2387, if you're happy with them. Its example currently repeats the same `jwt.decode(...)` with `issuer=` and `audience=` before it gets to the `typ` check, and that would land twenty lines below the one here. I'd cut it down to just the header check and point back to this section for the claim validation, so neither section restates the other. It also passes a bare `public_key`, which would read oddly sitting next to the `PyJWKClient` version here.

I think §3.12 fits better in #2387 than in this one. The last bullet there is the legacy and OIDC caveat, where `typ` can't be interoperably enforced and you fall back to distinct keys or `aud`/`iss` isolation, which is exactly what §3.12 prescribes. So I'll cite it there and cross-link the two sections rather than repeating it in both.

While re-reading #2387 I also found two citation errors of my own, the same kind as the ones you caught here:

- It cites "RFC 8725 §2.11" for cross-JWT confusion. There is no §2.11, section 2 stops at 2.9. The threat is §2.8 and Use Explicit Typing is §3.11. The anchor happened to be right, the number wasn't, and it was pointing at the mitigation while the sentence describes the threat.
- It attributes `"typ": "logout+jwt"` to RFC 8693, which is OAuth 2.0 Token Exchange and doesn't mention logout anywhere. `application/logout+jwt` is registered to the OpenID Foundation, from Back-Channel Logout 1.0. I'll correct that reference.

Pushing the fixes here now.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Good catch, taken. The opening now says "typically carries" and "may then be able to replay", and I've added your two cases as examples of deliberate omission: a token an application issues and consumes itself under a key used for nothing else, and an SD-JWT whose audience is established by the key binding JWT rather than by an `aud` claim.

The required-claims bullet is scoped the same way now, to deployments that actually rely on the claims, since a profile that establishes either by other means doesn't need the claim itself.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Agreed, that was a poor example. A staging environment normally signs with its own keys, so such a token fails at the signature and never reaches the `iss` check.

Dropped it. The issuer now has to be one the verifier also trusts, a partner tenant or a self-service account at a public identity provider, which is what makes the `kid`-across-a-union case work in the first place.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: True, and the bullet as written assumed a deployment with exactly one expected issuer.

Reworded: an equality check where a single issuer is expected, and membership of an explicit allowlist of issuer identifiers where issuers are provisioned dynamically, through a discovery protocol for example. The part worth keeping either way is that the allowlist stays explicit, rather than accepting whatever `iss` the token presents.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Agreed, removed.

@mackowski raised a related point in his review: `verify_iss` isn't only redundant but inert without `issuer=`, since `_validate_iss` returns immediately when `issuer` is None. The example now passes `require` alone, with a short note that the `issuer` and `audience` arguments are what perform the validation.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
 - [RFC 8725, Do Not Trust Received Claims](https://datatracker.ietf.org/doc/html/rfc8725#name-do-not-trust-received-claim);
 - [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114), a key embedded in the JWS header trusted for verification.
 
+### Issuer and audience confusion
+
+A JWT typically carries the `iss` (issuer) claim to say who created it and the `aud` (audience) claim to say who it was created for. Neither claim is mandatory, and some designs omit one deliberately: a token an application issues and consumes itself, under a key used for nothing else, or an SD-JWT whose audience is established by the accompanying key binding JWT rather than by an `aud` claim. Where a key does speak for more than one issuer or more than one recipient, however, a verifier that checks only the signature and the expiration accepts any token that key has signed, whoever issued it and whoever it was issued for. An attacker holding such a token, whether legitimately issued to them or obtained from a service they control, may then be able to replay it against a different recipient. RFC 8725 calls this a substitution attack.
```
randomstuff: Sorry to add some additional 🙇 comment but I'm wondering whether "who it was created for" isn't slightly ambiguous. This description might designate both the token holder (subject `sub`, authorized party `azp`, `client_id`) or the audience (`aud`). This nuance is quite important.

I'd suggest a wording such as "who is consuming the token" or "who is expected to validate the token". (or the "final recipient"?).

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
...
+
+Two variants are worth separating, because different checks defeat them.
+
+**Audience confusion.** An attacker presents a token issued for one service to a second service that trusts the same issuer. If the second service does not require its own identifier in `aud`, the token is accepted: a token minted for a low-privilege service is replayed against an internal API, and the attacker gains access that was never granted.
```
randomstuff: I would say "For example, a token minted for a low-privilege service is replayed against an internal API, and the attacker gains access that was never granted."

I would add another example on the "low-trust/third-party audience" axis. Something such as:

> On the other hand, a third-party application could try to replay a token against another (first-party or first-party) application, and gains accesses it it not supposed to access.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
...
+- Select the verification key from the set bound to the validated `iss`, for example that issuer's `jwks_uri`; never verify against a union of keys from several issuers. In multi-tenant deployments this means resolving the key set from an allowlist keyed by issuer rather than searching every tenant's keys.
+- Validate that the recipient's own identifier is present in `aud`, whether `aud` is a single string or an array of strings.
+- Where the deployment relies on these claims, reject tokens in which `iss` or `aud` is missing. RFC 7519 makes both optional, so this is a deployment decision rather than a specification requirement: a verifier with no `iss` has nothing to bind the key to, and one with no `aud` cannot tell whether the token was meant for it, but a profile that establishes either by other means does not need the claim itself.
+
+The related case, where the key material itself is taken from the token rather than merely selected by it, is covered in [Trusting key material named in the token header](#trusting-key-material-named-in-the-token-header). Where verification uses a MAC, see also the secret reuse bullets under [MAC](#mac).
```
randomstuff: Nitpick: "bullet" → "point"?

randomstuff (changes requested):
I noted a few points additional points 🙇. The main one is "who it was created for" which, I think, is ambiguous.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: No need to apologise, this is exactly the kind of thing worth catching. You're right that "who it was created for" reads as either the audience or the party the token is about, and those are different questions.

Reworded so `aud` says which party is expected to consume and validate the token, with a sentence separating that from who the token is about or who it was issued to, which `sub`, `azp` and `client_id` answer. Since you say the nuance matters, naming those three seemed better than leaving the contrast implicit.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Both taken. The low-privilege case now opens with "For example", and your third-party axis follows it: an application that legitimately receives tokens for its own use replaying one against another application, first-party or third-party, and reaching data or operations it was never granted.

I read the "(first-party or first-party)" in your suggestion as a slip for "first-party or third-party", so that's how it's written. Happy to narrow it if you meant something more specific.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Done, "points" it is.

jankesec:
All three are in, pushed as `f107c1a`.

@randomstuff, please don't feel you need to apologise for adding comments. The audience wording in particular was worth another round: the section is about `aud` confusion, so being loose about what `aud` actually designates was a bad place to be imprecise. Keep them coming if anything else looks off.

For anyone catching up, the section has moved a fair way since the first review:

- Key selection is bound to the validated `iss`, and the example resolves it through the issuer's JWK Set rather than taking a key that was passed in.
- The RFC 8725 citations point at §2.7 for the threat and §3.8 and §3.9 for the mitigations.
- `iss` validation is an equality check where one issuer is expected and an explicit allowlist where issuers are provisioned dynamically.
- The claims are treated as usually present rather than always, with the self-issued and SD-JWT cases called out.

`npm test` is clean.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Following that through, the same construction had survived one sentence later: "whoever issued it and whoever it was issued for". Named the two claims instead, so it now reads "whatever `iss` and `aud` it carries". Pushed as `1c13857`.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
```diff
@@ -324,6 +324,59 @@ References:
...
 - [CVE-2018-0114](https://nvd.nist.gov/vuln/detail/CVE-2018-0114), a key embedded in the JWS header trusted for verification.
 
+### Issuer and audience confusion
+
+A JWT typically carries the `iss` (issuer) claim to say who created it and the `aud` (audience) claim to say which party is expected to consume and validate it. That is a different question from who the token is about, or who it was issued to, which `sub`, `azp` and `client_id` answer. Neither claim is mandatory, and some designs omit one deliberately: a token an application issues and consumes itself, under a key used for nothing else, or an SD-JWT whose audience is established by the accompanying key binding JWT rather than by an `aud` claim. Where a key does speak for more than one issuer or more than one recipient, however, a verifier that checks only the signature and the expiration accepts any token that key has signed, whatever `iss` and `aud` it carries. An attacker holding such a token, whether legitimately issued to them or obtained from a service they control, may then be able to replay it against a different recipient. RFC 8725 calls this a substitution attack.
```
randomstuff: > That is a different question from who the token is about, or who it was issued to, which `sub`, `azp` and `client_id` answer.

I think, this sentence between the two other sentences confuse the reader about which "Neither claim" we are talking about. I think, we should drop the sentence about sub`, `azp` and `client_id`. The clarification of the first sentence is enough.

randomstuff (changes requested):
Minor change. I think it's good to go after that.

cheatsheets/JSON_Web_Token_Cheat_Sheet.md
jankesec: Fair, that sentence was sitting right where it could be misread. Dropped it, pushed as `6bc0581`. The first sentence carries the distinction on its own now:

`iss` says who created it, `aud` says which party is expected to consume and validate it, neither is mandatory.

randomstuff (approved):

Issue #1755, opened 2025-08-19
Update the page title of AJAX Security Cheat Sheet


ajayojha:
As per discussion under this [pull requestion](https://github.com/OWASP/CheatSheetSeries/pull/1729#issuecomment-3065668895) the [page title](https://cheatsheetseries.owasp.org/cheatsheets/AJAX_Security_Cheat_Sheet.html) should be "**Asynchronous Web Communication Security Cheat Sheet**". 

I have some suggestions based on the existing content on the AJAX Security Cheat Sheet page as the content is not aligned with the title and the existing content is not updated with latest best practices. The page content is not aligned with the newly decided page title as well. I would recommend creating three pages:

* Client-Side Best Practices
* Server-Side Best Practices
* Asynchronous Web Communication Security Cheat Sheet

Each should have the latest information. I will share the main points and subpoints for each page, if the suggestion aligns with the Cheat Sheet team.

jmanico:
The  terms "Client-Side Best Practices and Server-Side Best Practices" are really general, can we make this somehow more specific?

ajayojha:
How about these:

* Frontend Security Cheat Sheet
* Server-side Security Cheat Sheet

jmanico:
The main title I would suggest for most of this content https://cheatsheetseries.owasp.org/cheatsheets/AJAX_Security_Cheat_Sheet.html is "Client Side JavaScript Security"

I think the server side content can just be removed, it's covered elsewhere.

That way we can just edit this one cheatsheet instead of having the split it into three. It's fairly brief.

ajayojha:
> The main title I would suggest for most of this content https://cheatsheetseries.owasp.org/cheatsheets/AJAX_Security_Cheat_Sheet.html is "Client Side JavaScript Security"

The content scope is not limited to what is currently written for client side; it's good to add more content based on modern framework-agnostic practices. I would suggest "Client-Side Security" or "Frontend Security" instead of "Client Side JavaScript Security", as there is a good chance that TypeScript, JSX, Vue, and Svelte code examples will be added. Indirectly they all are transpiled into the JavaScript but still.

jmanico:
Go for it, I'm with you.

ajayojha:
Ok, thanks @jmanico. What should I do next? Should I add the content in the existing file, and later the file name will be changed, or should I create a new file in the `cheatsheets_draft` folder and start drafting the content? Before drafting, I will share the main titles.

[...]

PR #2394, opened 2026-09-02
Update AJAX Security Cheat Sheet title to Web Frontend Security Cheat Sheet

emihesi99:
This PR updates the page title as proposed in #1755. It replaces 'AJAX Security Cheat Sheet' with 'Asynchronous Web Communication Security Cheat Sheet' in the main cheat sheet and index files. Addresses #1755


jmanico (commented):
I think we agreed to rename this "Web Frontend Security Cheat Sheet" instead, is that ok?

Issue #2401, opened 2026-09-03
SSRF Prevention Cheat Sheet: URL parser differentials are not covered


tamerkalla:
`http://example.com\@evil.com` is `example.com` to JavaScript and `evil.com` to Python.

```
$ node -e "console.log(new URL('http://example.com\\\\@evil.com').hostname)"
example.com

$ python3 -c "from urllib.parse import urlsplit; print(urlsplit('http://example.com\\\\@evil.com').hostname)"
evil.com
```

Both are correct. WHATWG URL treats a backslash as a path separator in a
special scheme, so the authority ends at the backslash. RFC 3986 treats it as
an ordinary character, so the authority runs to the last `@` and the userinfo
is `example.com\`. Neither standard is going to adopt the other's rule.

## Why this matters for the SSRF Prevention Cheat Sheet

The cheat sheet's application-layer guidance is, in effect: parse the URL,
extract the host, check it against an allowlist, then make the request. That
is sound when one parser does both the checking and the fetching. It is not
sound when the URL crosses a service boundary **as a string**, validated in one
runtime and fetched by another. That is the common shape in a gateway, a webhook
dispatcher, a redirect follower, or any polyglot service mesh.

In that shape the allowlist check and the request can disagree about which
host the URL names, and neither component is buggy. The current text does not
mention this case, so a reader following it closely can still build a
bypassable allowlist.

## How wide the gap is

Over a deterministic enumeration of 312 URLs (2 schemes x 26 authority
constructions x 6 path suffixes, no sampling) the WHATWG reading and the
RFC 3986 reading name a **different host on 84 of them**. Seven distinct
authority constructions account for those 84, and each one works differently:

| authority | WHATWG reads | RFC 3986 reads | mechanism |
|---|---|---|---|
| `example.com\@evil.com` | `example.com` | `evil.com` | backslash is a slash in a special scheme |
| `0x7f.1` | `127.0.0.1` | `0x7f.1` | hexadecimal IPv4 canonicalisation |
| `017700000001` | `127.0.0.1` | `017700000001` | octal IPv4 canonicalisation |
| `2130706433` | `127.0.0.1` | `2130706433` | decimal IPv4 canonicalisation |
| `①.com` | `1.com` | `①.com` | UTS-46 mapping |
| `exаmple.com` (Cyrillic а) | `xn--exmple-4nf.com` | `exаmple.com` | IDNA to ASCII |
[...]

PR #2412, opened 2026-09-08
SSRF: validate with an allowlist, not with a parser

tamerkalla:
This adds two short paragraphs to the SSRF Prevention Cheat Sheet, at the end of the **URL** subsection under the Case 1 application layer defences. It is the change discussed and approved in issue #2401, revised after @randomstuff pointed out that the original single headline read as though parsing were the alternative to an allowlist.

One file, no table, no figures and no code sample, per the request in the issue to keep it brief.

## What it adds

> **Match the host against an allowlist, and build the request yourself.** The input here is an IP address or a domain name rather than a whole URL, so compare that value against an explicit [allowlist](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#allowlist-vs-denylist) of permitted destinations and then build the request from the allowlisted value together with a scheme, port and path the application fixes itself. Carrying any part of the user's original string through, its path and query included, hands the next component something it has to parse again.
>
> **Treat parser disagreement as a rejection.** Where a URL does cross a service boundary as a string and is parsed again at the other end, two implementations can read different hosts from the same bytes. `http://example.com\@evil.com` is such a string: a parser following the [WHATWG URL Standard](https://url.spec.whatwg.org/#url-parsing) treats the backslash under a special scheme as a path separator and reads the host as `example.com`, while the `userinfo` grammar in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.1) admits no backslash at all, so the string is not a URI that RFC defines. An implementation that does not enforce the grammar still returns a host: CPython's [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) reads it as `evil.com`. Reject a URL whose host is not read identically by every parser in play, rather than reconciling the readings.

## Why

The first rule is the control. The allowlist decides, and the value that was matched is the value that gets used, so whatever performs the fetch never parses the original string again.

The second is the failure mode that survives the first. The existing application layer guidance holds when one parser both validates the URL and fetches it. It does not hold when the URL crosses a service boundary as a string, validated in one runtime and fetched by another, which is the shape of a gateway, a webhook dispatcher or a redirect follower. There the two components can name different hosts from the same bytes. `http://example.com\@evil.com` is `example.com` to a parser following the WHATWG URL Standard, which treats the backslash under a special scheme as a path separator. RFC 3986 does not admit a backslash in `userinfo` at all, so a parser holding to its grammar rejects the string outright while a lenient one returns `evil.com`. Rejection and a differing host are both disagreement, and both are reasons not to proceed, which is why the disagreement is itself the signal to reject.

The addition does not teach that difference and deliberately carries no measurements. It says what to do about it.

- [ ] In case of a new Cheat Sheet, you have used the [Cheat Sheet template](../templates/New_CheatSheet.md). Not applicable, this edits an existing cheat sheet.
- [x] All your assets are stored in the **assets** folder. No assets added.
- [x] All the images used are in the **PNG** format. No images added.

The citations are the normative text plus the named implementation: [URL parsing](https://url.spec.whatwg.org/#url-parsing) in the WHATWG URL Standard, the [`userinfo` grammar in section 3.2.1](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.1) of RFC 3986, CPython's [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) for the lenient reading, and [Allowlist vs Denylist](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#allowlist-vs-denylist) for the allowlist itself. Every anchor was opened and checked before it was written. The divergence they describe was verified by running the same 312 URLs through both readings, Node 22.22.2 and CPython 3.11.15, with 84 producing a different host. Those figures are in issue #2401 and none of them are proposed for the cheat sheet.

This PR fixes issue #2401.


cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.md
```diff
@@ -245,6 +245,8 @@ Do not accept complete URLs from the user because URL are difficult to validate
 
 If network related information is really needed then only accept a valid IP address or domain name.
 
+**Validate with an allowlist, not with a parser.** Compare the target against an explicit allowlist of permitted hosts, and pass the allowlisted value downstream rather than the user's original string. Where a URL must be parsed at all, treat disagreement between two parsers as a rejection rather than something to reconcile: the [WHATWG URL Standard](https://url.spec.whatwg.org/#host-parsing) and [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2) do not always read the same host from the same string, so a host that is not read identically by every convention in play should never reach a fetch.
```
randomstuff: I find it hard to understand :/ For example "**Validate with an allowlist, not with a parser.**" does not seem clear to me.

tamerkalla:
@randomstuff You are right, and the fault is in the headline rather than in the detail. "Validate with an allowlist, not with a parser" reads as though parsing were the alternative to an allowlist, when you have to parse the URL to get a host to check in the first place. It compressed two separate rules into one slogan and lost both of them.

I have pushed a revision that names them separately:

> **Match the host against an allowlist, and forward what you matched.** Parse the target, compare the resulting host against an explicit allowlist of permitted destinations, and pass that allowlisted host downstream rather than the user's original string, so the component that performs the fetch never re-parses attacker-controlled text.
>
> **Treat parser disagreement as a rejection.** Where a URL crosses a service boundary as a string and is parsed again at the other end, the two parsers can read different hosts from it and neither is wrong: the WHATWG URL Standard and RFC 3986 do not always agree. Reject a URL whose host is not read identically by every convention in play, rather than reconciling the readings.

The first rule is the control: the allowlist decides, and the value that was matched is the value that gets used, so whatever fetches never parses the original string again.

The second is the failure mode that survives the first. When a URL is validated in one runtime and fetched by another, the two can read different hosts from the same bytes. `http://example.com\@evil.com` is `example.com` to a WHATWG parser, which treats the backslash as a path separator under a special scheme, and `evil.com` under RFC 3986, where it is an ordinary character and so the authority runs to the last `@`. Neither implementation is buggy, which is why there is nothing to reconcile and the disagreement is itself the signal to reject.

If two short paragraphs is more than the section wants, I am happy to cut it to the first rule alone and drop the second.

jmanico (dismissed):

tamerkalla:
@mackowski You said on #2401 that you were waiting for the pull request. It is open, and I think it is ready for your read.

jmanico has approved it. A reviewer then found the original headline unclear and was right: "validate with an allowlist, not with a parser" reads as though parsing were the alternative to an allowlist, when a host has to be parsed out of the URL before it can be checked against one. It is now two named rules instead of one sentence: match the host against the allowlist and forward what you matched, and treat parser disagreement as a rejection. Both citations are unchanged and the diff is still one file.

It needs a second code owner approval to merge. No urgency from my side if you would rather it waited; I am only flagging it because you asked to know when it existed.

mackowski (changes requested):
Thanks for this — the gap is real, and I appreciate you keeping it to four lines as asked. I reproduced both differentials locally, so the phenomenon itself isn't in question. Two things need fixing before this can go in.

**1. The safety guarantee in paragraph 1 doesn't hold.**

> pass that allowlisted host downstream ... so the component that performs the fetch never re-parses attacker-controlled text

The host usually isn't the only user-controlled part of the request. If the path comes from the user too, the fetcher re-parses it and the authority can move:

```
"http://" + "example.com" + "\@evil.com/p"
```

`curl` resolves and connects to `evil.com` for that string. So the sentence promises something the control doesn't deliver, and a developer following it literally rebuilds the bypass.

Could you rephrase to say the outbound request should be rebuilt from separately validated components using the HTTP client's structured API (host / port / path / query), percent-encoding each user-supplied segment, rather than by concatenation? And note that the allowlist covers the **host only**, so the redirect-following and DNS-pinning controls already documented above still apply.

**2. The citations don't support the claims they're attached to.**

The WHATWG `#host-parsing` section operates on an already-extracted host string — it contains no mention of `U+005C`, backslashes, or special schemes. The backslash rule lives in the basic URL parser's [authority state](https://url.spec.whatwg.org/#authority-state).

RFC 3986 §3.2 also doesn't make `\` an "ordinary character". Both `userinfo = *( unreserved / pct-encoded / sub-delims / ":" )` and `reg-name = *( unreserved / pct-encoded / sub-delims )` exclude it, so `example.com\@evil.com` isn't a valid RFC 3986 authority at all. The `evil.com` reading comes from permissive implementations (curl, Python's `urllib`), not from the RFC — Go's `net/url` and Ruby's `URI` both reject the string outright. That also means **"neither is wrong" overstates it**: both specs treat the input as malformed and merely recover differently. (Copilot's latest review flags this same point independently.)

The cleanest fix may be to switch the example to IPv4 canonicalization instead: `http://0x7f.1/` is `127.0.0.1` to WHATWG and `0x7f.1` to an RFC 3986 `reg-name`. That one genuinely *is* valid-under-both, and it lives under the `#host-parsing` anchor you already cite.

**Two smaller things:**

- The paragraph lands directly after "Do not accept complete URLs from the user", so please open with something like "If a complete URL cannot be avoided, ..." so it doesn't read as contradicting the sentence above it.
- "Reject a URL whose host is not read identically by every convention in play" needs two parsers to act on, and most services ship exactly one. A deterministic pre-parse reject gives the same protection with one: refuse any URL containing `\`, a raw `@`, control characters, or non-ASCII in the authority before parsing it.

Finally, please restore the **AI Tool Usage Disclosure** section from the PR template — it's required on every PR, and either checkbox is fine.

Happy to look again once updated.

tamerkalla:
Both Copilot findings are right, and the second one is a factual error rather than a wording problem. Fixed in the commit above.

**On the citation for the first rule.** The claim that forwarding the matched value stops a downstream re-parse was stated as though the linked specifications supported it, and they do not: they describe how parsers read a URL, not how an application should move a value between components. The sentence now says only what is true by construction, that what travels onward is a value the application selected from its own allowlist, so nothing further along parses the original input again. No new citation, because the narrowed sentence no longer needs one.

**On RFC 3986 and the backslash.** You are correct and the old text was wrong. The `userinfo` grammar in section 3.2.1 is `*( unreserved / pct-encoded / sub-delims / ":" )`, and a backslash is in none of those sets, so `http://example.com\@evil.com` is not a URI that RFC 3986 defines a host for. Describing `evil.com` as the RFC's reading conflated the grammar with what a permissive implementation happens to return.

The paragraph now separates the two:

> **Treat parser disagreement as a rejection.** Where a URL crosses a service boundary as a string and is parsed again at the other end, two implementations can read different hosts from the same bytes. `http://example.com\@evil.com` is such a string: a parser following the [WHATWG URL Standard](https://url.spec.whatwg.org/#url-parsing) treats the backslash under a special scheme as a path separator and reads the host as `example.com`, while the `userinfo` grammar in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.1) does not admit a backslash at all, so a strict parser rejects the string outright and a lenient one reads the host as `evil.com`. Reject a URL whose host is not read identically by every parser in play, rather than reconciling the readings.

That also makes the advice stronger rather than weaker: rejection and a differing host are both disagreement, and both are reasons not to proceed. The citation moved from section 3.2 to section 3.2.1, which is the grammar actually being relied on.

mackowski (dismissed):

tamerkalla:
All three of the remaining findings are fair. Pushed.

**Forwarding the host alone is not the same as forwarding the target.** This is the substantive one. A fetch needs a scheme, a port and a path as well, and if those are carried over from the user's string then the differential comes straight back in through the part that was not matched. The rule is now scoped to what this section actually accepts, an IP address or a domain name rather than a whole URL, and says to build the request from the allowlisted value plus a scheme, port and path the application fixes itself.

**On the citation for that rule.** It now cites the [Allowlist vs Denylist](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#allowlist-vs-denylist) section of the Input Validation Cheat Sheet. Worth flagging while I am here: the link to that section from this cheat sheet's own Case 1 text points at `Input_Validation_Cheat_Sheet.md#allow-list-vs-block-list`, and that anchor does not exist. The heading is `Allowlist vs Denylist`. It predates this pull request and I have left it alone rather than widen the diff, but it is a dead anchor if anyone wants it fixed.

**On the lenient reading.** Named and cited rather than left as "a lenient parser": CPython's [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit). Measured on CPython 3.11.15, it returns `evil.com` for that string.

The paragraphs now read:

> **Match the host against an allowlist, and build the request yourself.** The input here is an IP address or a domain name rather than a whole URL, so compare that value against an explicit [allowlist](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html#allowlist-vs-denylist) of permitted destinations and then build the request from the allowlisted value together with a scheme, port and path the application fixes itself. Carrying any part of the user's original string through, its path and query included, hands the next component something it has to parse again.
>
> **Treat parser disagreement as a rejection.** Where a URL does cross a service boundary as a string and is parsed again at the other end, two implementations can read different hosts from the same bytes. `http://example.com\@evil.com` is such a string: a parser following the [WHATWG URL Standard](https://url.spec.whatwg.org/#url-parsing) treats the backslash under a special scheme as a path separator and reads the host as `example.com`, while the `userinfo` grammar in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.1) admits no backslash at all, so the string is not a URI that RFC defines. An implementation that does not enforce the grammar still returns a host: CPython's [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) reads it as `evil.com`. Reject a URL whose host is not read identically by every parser in play, rather than reconciling the readings.

Every anchor above was opened and checked before it was written, which is how the dead one in the existing text turned up.

jmanico (dismissed):

tamerkalla:
@mackowski @randomstuff This is one approval short of the two required, and I think everything raised has now been answered.

@mackowski, your approval was dismissed when the commit after it landed, so it no longer counts toward the total. GitHub does not surface that anywhere, which is the only reason I am pointing at it rather than waiting.

What changed since each of you last read it:

The guidance is scoped to the case this section actually covers, an IP address or a domain name rather than a whole URL, and it now says to build the request from the allowlisted value with a scheme, port and path the application fixes itself, rather than to forward any part of the user's string.

The RFC 3986 citation moved to section 3.2.1, the `userinfo` grammar, because that is what the backslash claim rests on. The example now separates what the grammar admits from what a lenient implementation returns: RFC 3986 admits no backslash there at all, while CPython's `urllib.parse.urlsplit` still returns a host for it, and the two hosts differ.

An approval from either of you is the second one. If any of it still reads wrong, say so and I will fix it rather than let it sit.

jmanico (commented):
Actually I withdraw my approval. One more thing:

I have two small items before I can approve:

1. The CPython citation supports the general point that `urllib.parse` is not a validating RFC parser, but it does not document the specific claim that this input produces `evil.com`. Could you cite the canonical CPython implementation or another primary source demonstrating that exact behavior, and preferably phrase it precisely as `urlsplit(...).hostname` returning `evil.com`?

2. One non-blocking wording suggestion: since the preceding sentence scopes the input to an IP address or domain name, “its path and query included” sounds as though the original input still contained a URL. I would clarify that this refers to components copied from an original user-supplied URL.

With those addressed, I’m comfortable approving this.

tamerkalla:
@jmanico Both fair. Pushed a commit for each.

**The `urlsplit` claim.** The behaviour comes from how the authority is split rather than from anything specific to that one function: `urllib.parse` takes everything before the final `@` to be userinfo, at [`netloc.rpartition('@')`](https://github.com/python/cpython/blob/v3.11.15/Lib/urllib/parse.py#L208) in `Lib/urllib/parse.py`. The paragraph now cites that line rather than asserting a return value, since the library reference documents the function without giving this case.

Reproduced on CPython 3.11.15:

```
>>> from urllib.parse import urlsplit
>>> urlsplit("http://example.com\\@evil.com").hostname
'evil.com'
>>> urlsplit("http://example.com\\@evil.com").username
'example.com\\'
```

For the contrast the paragraph draws, Node 22.22.2 reads the same string as `example.com` with the backslash starting the path.

**Path and query.** You are right that it did not belong. The input is a host, so it has no path or query to carry and the sentence was describing a different threat. It now says what was meant: send on the allowlist entry that matched, not the string the user supplied.

tamerkalla:
@jmanico Both done properly this time, and apologies: my last round answered the substance of your two points but not what you actually asked for.

**1.** The sentence now states the result in the form you asked for, alongside the implementation citation:

> CPython's `urllib.parse` derives the host with `netloc.rpartition('@')`, so `urlsplit("http://example.com\@evil.com").hostname` returns `evil.com`.

**2.** You suggested clarifying that the copied components come from an original user-supplied URL, and I deleted the sentence instead, which lost the point you were making. It is back, scoped:

> Where that host was extracted from a URL the user supplied, do not copy the other components of that URL across either: carrying its path or query through, rather than rebuilding them, hands the next component something it has to parse again.

mackowski (approved):

jmanico (approved):

